PR #18, opened 2014-11-28
Prepending scope identifiers for global parameters to limit namespace collisions

jchodera:
Since [OpenMM #717](https://github.com/pandegroup/openmm/issues/717) is under debate, this PR attempts to prepend scope-limiting identifiers to global context parameters to limit the potential for namespace collisions in test systems and integrators.

So far, I have just fixed up the test systems.  I am still working on the integrators.

This solution is to be considered temporary at best.  I have used the convention `testsystems_TestSystemName_parametername` for global context parameters previously named `parametername`.  I have tried to avoid adding complexity to the energy expressions by introducing intermediate temporary variables in energy expression that should be compiled out by Lepton.  I have also eliminated some context parameters that didn't need to be exposed as context parameters.  Some others may still be eliminated.

Still a work in progress.


jchodera:
I think this is ready to merge.

PR #26, opened 2014-12-31
Fixing travis

kyleabeauchamp:


kyleabeauchamp:
Not sure.  I presume the simplest is to download from `main` and upload to `dev`.  We can add more fine-grained channels (e.g. travis, dev, jenkins) if we later deem it necesary.

jchodera:
Merging this.

PR #28, opened 2015-01-13
Added Sobol subrandom sampling approach to initialize particle positions

jchodera:
This PR changes the default subrandom sampling approach to initialize particle positions from a Halton sequence to a Sobol' sequence using [this Python library](http://people.sc.fsu.edu/~jburkardt/py_src/sobol/sobol.html).  It is distributed under the MIT license, so that license file is included as well.

I've also added a `LennardJonesGrid` that is close in functionality to the old `LennardJonesFluid` that created particles on a grid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ env:
 
 after_success:
   - echo "after_success"
-  - source devtools/ci/after_sucess.sh
+  - # source devtools/ci/after_sucess.sh # DISABLED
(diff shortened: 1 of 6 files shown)


jchodera:
I've also changed the `devtools/ci/after_success.sh` script to upload to `openmmtools-dev` and not overwrite `openmmtools` on binstar.

jchodera:
The Python 2.7 tests are passing, but we will need to address 3.x compatibility in a separate PR.

PR #33, opened 2015-01-14
[WIP] Segfault workaround

jchodera:
Attempting to work around `openmm` segfaults.


rmcgibbo:
Is there an issue/fix related to this upstream in OpenMM?

jchodera:
Yes. We're trying to create an isolated test case.

Here's an example of the behavior we're seeing in `yank`:
https://travis-ci.org/choderalab/yank#L1268

```
Testing reference/alchemical overlap for Lennard-Jones fluid without dispersion correction... ... /home/travis/miniconda/conda-bld/test-tmp_dir/run_test.sh: line 3:  3650 Segmentation fault      nosetests yank --nocapture --verbosity=2 --with-doctest
```

rmcgibbo:
> Yes. We're trying to create an isolated test case.

Got it.

jchodera:
Created OpenMM issue https://github.com/pandegroup/openmm/issues/781 with minimal test case.

The fix I've put in place reduces the default density for `LennardJonesFluid` and `LennardJonesGrid` to be appropriate for a gas or dilute fluid.  This should avoid the segfaults in yank's travis.

Should be good to merge once tests pass.

jchodera:
I've removed all the `openmmtools` packages from the binstar and restarted the Jenkins build, but we should probably just increment the minor version number in the future.  Sorry for forgetting to do that!

Issue #34, opened 2015-01-26
Use classes for integrators?


kyleabeauchamp:
It might be cute to use classes instead of functions, so that we can use inheritance to do further customizations.  This is not unprecedented, as Peter's RESPA integrator does this.

This would also let us write decorators that calculate various convergence statistics for monitoring.  Also, we could use member functions to help with tuning various mcmc parameters.

jchodera:
Good idea.  I've thought about this in earlier passes through this code.

Can you point me to Peter's RESPA integrator?  Is that pure python?

kyleabeauchamp:
https://simtk.org/svn/pyopenmm/trunk/simtk/pyopenmm/extras/respa.py

We should just copy this into openmmtools, so that we can actually use and test it.

jchodera:
Will do.

I hadn't realized you could subclass swig-wrapped C++ classes like `CustomIntegrator`.  Will do this for the others in a PR.

kyleabeauchamp:
Done in #36

PR #35, opened 2015-01-26
Add respa.py, convert HMC to class

kyleabeauchamp:
I figure we should convert one integrator first, finalize the API, then convert the other integrators.


jchodera:
I mentioned in #34 that I am already working on a PR for this.  Now we have competing PRs.

jchodera:
See #36.

kyleabeauchamp:
Sounds good, we can consolidate on yours

PR #36, opened 2015-01-26
Adding rRESPA and converting all integrators to classes.

jchodera:
Will need to finish converting integrators.py and tests.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -52,7 +52,7 @@ kB = units.BOLTZMANN_CONSTANT_kB * units.AVOGADRO_CONSTANT_NA
 # INTEGRATORS
 #=============================================================================================
 
-import respa
+from openmmtools import respa
 class MTSIntegrator(respa.MTSIntegrator):
     """
     MTSIntegrator implements the rRESPA multiple time step integration algorithm.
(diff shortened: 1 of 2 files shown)


jchodera:
Addresses #34

kyleabeauchamp:
So looks like the diff is pretty ugly from the way this conversion necessarily works (changes in tab structure).  Maybe you can briefly summarize how you changed things here in the discussion.

kyleabeauchamp:
I wonder if it makes sense to dump `respa.py` into a separate file and add `from .respa import MTSIntegrator` in `integrators.py`.  That might help preserve the provenance information.

jchodera:
OK, can do, but I need to do some gymnastics to make sure every integrator has default options (for testing purposes).

jchodera:
> So looks like the diff is pretty ugly from the way this conversion necessarily works (changes in tab structure).

This is the drawback of using a language where spacing is an integral part of syntax.

>  Maybe you can briefly summarize how you changed things here in the discussion.

* The `MTSIntegrator` from `respa.py` was integrated (adding default options)
* All functions that used to return a `CustomIntegrator` object are now subclasses of `CustomIntegrator` whose constructor initializes the object
* Some minor cleanup to docstrings

I may need to tinker with the tests, but it looks like they should work.

jchodera:
Seems like Python 3.x doesn't like the fact that this ratio can be a `float`:
https://github.com/choderalab/openmmtools/pull/36/files#diff-669299b3205f958c7662890143395b26R79

jchodera:
Tests pass. This should be ready to merge.

Issue #38, opened 2015-01-27
Unused parameter `flexibleConstraints` in testsystems


kyleabeauchamp:
A simple search for `flexibleConstraints` in testsystems suggests that this parameter is never actually used anywhere--it's present only in the constructor function header and nowhere else.

kyleabeauchamp:
Fixed in #39

jchodera:
To give context: This was used before we had a minimizer inside of OpenMM, and needed to use something like LBFGS from `scipy` that didn't know about constraints.

PR #39, opened 2015-01-27
Deleted flexibleConstraints

kyleabeauchamp:
See #38


jchodera:
I agree that this is no longer needed.

rmcgibbo:
I think you accidentally checked in a .pyc file.

PR #40, opened 2015-03-10
Adding 'topology' as an accessible parameter for testsystems

jchodera:
This PR adds access to the topologies for each test system via a `testsystem.topology` property.

### DIFF
--- a/openmmtools/testsystems.py
+++ b/openmmtools/testsystems.py
@@ -361,11 +361,11 @@ class TestSystem(object):
         """The simtk.openmm.app.Topology object corresponding to the test system."""
         return copy.deepcopy(self._topology)
 
-    @system.setter
+    @topology.setter
     def topology(self, value):
-        self._topology = copy.deepcopy(value)
+        self._topology = value
 
-    @system.deleter
+    @topology.deleter
     def topology(self):
         del self._topology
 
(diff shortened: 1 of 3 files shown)


jchodera:
OK, this seems to be working now.

Can someone else take a quick look?  @kyleabeauchamp or @jhprinz?

This adds a `.topology` property to each model system that should accurately provide an OpenMM `Topology` object. I've added a test to make sure that the `Topology` object has the same number of atoms as the OpenMM `System` object.

kyleabeauchamp:
looking now

openmmtools/testsystems.py
```diff
@@ -350,6 +357,19 @@ def positions(self):
         del self._positions
 
     @property
+    def topology(self):
```
kyleabeauchamp: So with the deepcopy here, it's essentially impossible to edit the underlying topology via `testsystem.topology`.  I guess this is probably a good idea here, but it's worth thinking about.

jchodera: Currently, `testsystem._topology` is mutable, `testsystem.topology` returns an immutable copy.

We do the same with `testsystem.system`.  

I'm happy to go either way on this.  In particular, I'd love to go with whichever behavior is the least surprising to users/developers.

kyleabeauchamp:
I'm fine with sticking with what you've got, but IMHO we want to describe this behavior in a Notes section in a docstring somewhere.

jchodera:
On further reflection, I think we should switch these back to being mutable, since that would likely cause less confusion.  But let's do that in a separate PR.

jchodera:
Everything else OK?  I realize there is some inconsistency in how these topology objects are constructed in each system, but we can probably clean that in the future.

kyleabeauchamp:
AFAIK this seems reasonable for now.

jchodera:
I just updated the `TestSystem` docstring to be clearer, but we can change this in the PR to address #41.

Issue #41, opened 2015-03-11
Make testsystem.system and testsystem.topology mutable rather than returning a deep copy


jchodera:
See discussion in #40

PR #42, opened 2015-03-12
TestSystem now returns mutable system, positions, topology

jchodera:
Implements #41.


kyleabeauchamp:
From the test failures, it looks like we've got the old behavior hard-coded somewhere.

jchodera:
I found an *actual* bug in `CustomLennardJonesFluidMixture` when no switch was used that caused the error, I think.  Not sure why this wasn't found before!

jchodera:
Seems to have fixed it!

Issue #48, opened 2015-04-20
travis build does not upload newest version and binstar version cannot be imported


jhprinz:
I just had to reinstall Anaconda from scratch. Completely new user. Newest Anaconda and installed openmmtools. It installs using conda but importing fails

`import openmmtools` with

```

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-5-113cf65ec230> in <module>()
----> 1 import openmmtools

/Users/jan-hendrikprinz/anaconda/lib/python2.7/site-packages/openmmtools-0.9.0-py2.7.egg/openmmtools/__init__.py in <module>()
      7 
      8 # Define global version.
----> 9 from openmmtools import version
     10 __version__ = version.version
     11 

ImportError: cannot import name version
```

Apparently the newest build did not get uploaded to binstar. In the after_success the `{PACKAGENAME}` seemed to be empty which results in the binstar command not being able to find the package. Has there been some changes recently?

jchodera:
Shoot.  Thanks for noticing this.  We've had this issue before, but I'm not sure what changed.  Will look into this.

jchodera:
Seems to be this error:
https://travis-ci.org/choderalab/openmmtools/jobs/55646811#L1442
```
[BinstarError] file /home/travis/miniconda/conda-bld/linux-64/-* does not exist 
```

jchodera:
`$PACKAGENAME` is not being set in `devtools/ci/after-success.sh`:
https://github.com/choderalab/openmmtools/blob/master/devtools/ci/after_sucess.sh#L16

jchodera:
Should be addressed by #49.  Please reopen this issue if the current travis build doesn't upload the latest `openmmtools`.

PR #49, opened 2015-04-20
Attempting fix for failed binstar upload due to missing $PACKAGENAME

jchodera:
Attempts to address #48


jchodera:
Looks like it should work now.  Merging.

PR #50, opened 2015-04-22
Added JAC DHFR testsystem, constants.py, and bugfix

kyleabeauchamp:


kyleabeauchamp:
Bugfix was a hard-coded nonbonded cutoff for ALA explicit testsystem.

I've also added a hydrogen masses flag to the ALA testsystem and the DHFR testsystem, which will facilitate HMR studies

jchodera:
Can you add provenance info (and ideally any scripts used to generate the `prmtop` files) along with this?

jchodera:
Oh, sorry---this was just a bugfix?

jchodera:
Ah, it's the actual first check-in of these files.  I see some provenance info in the `README.md`.

This is an old-style (pre-AMBER6) `prmtop` file.  Can the OpenMM `prmtop` reader handle these?

kyleabeauchamp:
It's using parmed to read these files as I don't think OpenMM supports those file formats.

jchodera:
> It's using parmed to read these files as I don't think OpenMM supports those file formats.

Correct, it doesn't.  I didn't see the ParmEd bits.  This sounds good, though!

Do we want to make ParmEd a required dependency now that it is available in `omnia`?

kyleabeauchamp:
It's not available just yet, so we can postpone the required dependency until it is.

jchodera:
Sounds good.

jchodera:
Does it make sense to add the other benchmarks form the Amber14 benchmark suite as well?

kyleabeauchamp:
I think DHFR is good enough for now, as adding the rest will bloat the repo quite a bit--they're somewhat larger systems.

jchodera:
There could be value in being able to easily iterate through a benchmark suite.

We could in principle gzip the `prmtop` and `inpcrd` files and feed streams to the readers, I think.

Compressed, I think the unique `prmtop` and `inpcrd` files (excluding the HMR 4 fs files) come to only 23M.

kyleabeauchamp:
If adding more benchmarks is a priority, I'd say we should create a new issue and pull request.  I don't think it's high priority right now, whereas DHFR is very commonly used.

jchodera:
OK, I'll create a separate issue to add the others.

Just thought it might be easier to add them all at once in a self-consistent way instead of adding everything piecemeal.

kyleabeauchamp:
OK, I think this is now working.

PR #53, opened 2015-05-08
Added old platform energy error testing script from old repex

jchodera:
I've added a script (under `openmmtools/examples/old-test-scripts/`) that compares systems from `testsystems` on all implemented platforms to look for energy deviations.

Eventually, we can make this a nosetest, but there is some utility in having this script around for now.


kyleabeauchamp:
thanks!

PR #54, opened 2015-05-09
test-openmm-platforms is now installed as a command-line script

jchodera:
`test-openmm-platforms` is now installed as a command-line script since this may be useful for debugging issues with hardware or software OpenMM installations.


kyleabeauchamp:
Looks good to me, can merge after travis

jchodera:
Should we have a nosetest for this as well?  I could even refactor this into a nosetest and a script that calls the nosetest...

kyleabeauchamp:
IMHO we can punt on nosetests for now

kyleabeauchamp:
Looks like we need to fix print statements for py3k

jchodera:
I'm getting this error:
```
DHFRExplicit
DHFR test system requires Parmed (`import chemistry`).
  File "/Users/choderaj/anaconda/bin//test-openmm-platforms", line 9, in <module>
    load_entry_point('openmmtools==0.6.2', 'console_scripts', 'test-openmm-platforms')()
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/openmmtools-0.6.2-py2.7.egg/openmmtools/scripts/test_openmm_platforms.py", line 211, in main
    testsystem = constructor()
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/openmmtools-0.6.2-py2.7.egg/openmmtools/testsystems.py", line 2569, in __init__
    raise(e)
ImportError: No module named chemistry
```
We should be requiring ParmEd as a dependency now.  @kyleabeauchamp : Can you create a PR to do that?

kyleabeauchamp:
It's not available as a conda recipe yet.  

All of our nose tests have try except clauses that take care of optional dependencies.

kyleabeauchamp:
I think we should just keep this as an optional dependency until parmed has a conda release.

jchodera:
> It's not available as a conda recipe yet.

I think this was just waiting on the (now complete) release of AmberTools 15, right?  Should we wrap this now?

jchodera:
> I think we should just keep this as an optional dependency until parmed has a conda release.

This would mean we have to rejigger our whole testing framework, would it not?  We need a different way of discriminating between failing test cases and optional tests that fail due to missing dependencies.  Any ideas?

jchodera:
> All of our nose tests have try except clauses that take care of optional dependencies.

Ah, OK.  I missed that.  Can you give a pointer to an example of the idiom for handling this?

kyleabeauchamp:
https://github.com/choderalab/openmmtools/blob/master/openmmtools/tests/test_testsystems.py#L55

jchodera:
Thanks!  That should do the trick for now.

Added a (simple) nosetest too, but not sure if this will cause a timeout.

jchodera:
The `--nocapture` flag doesn't seem to be working.

kyleabeauchamp:
Hmm, the tests seem OK otherwise--there are just some test failures with energy differences...

jchodera:
Oh, sorry!  I didn't see the output---you're right that it's there.

I'll go ahead and merge this then, and we can either fix the test systems that are causing problems or file bug reports for OpenMM if these are not `openmmtools` bugs.

PR #55, opened 2015-05-10
Fix problematic testsystems

jchodera:
I'm starting to fix some problematic test systems in `openmmtools.testsystems` that had large potential energies associated with them, since these would generally lead to large energy differences between platforms.

### DIFF
--- a/openmmtools/testsystems.py
+++ b/openmmtools/testsystems.py
@@ -1510,7 +1510,7 @@ class LennardJonesFluid(TestSystem):
     """
 
     def __init__(self,
-        nparticles=500,
+        nparticles=1000,
         reduced_density=0.05,
         mass=39.9 * unit.amu, # argon
         sigma=3.4 * unit.angstrom, # argon,
--- a/openmmtools/testsystems.py
+++ b/openmmtools/testsystems.py
@@ -789,8 +789,14 @@ class Diatom(TestSystem):
 class DiatomicFluid(TestSystem):
     """Create a diatomic fluid.
 
+    Note
+    ----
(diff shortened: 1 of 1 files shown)


jchodera:
Eventually, we can store serialized XML versions of minimized/equilibrated forms of these systems, but it would be useful to at least have reasonable programmatically-generated initial conditions (which is what we try to do now).

kyleabeauchamp:
So for the LJ box, I presume another way to build a starting box might be to treat the particles as hard spheres and use a sphere packer (for example https://pypi.python.org/pypi/spack).

rmcgibbo:
Isn't this what packmol is for?

kyleabeauchamp:
Yeah, I suppose I could do that too.

jchodera:
Is PackMol deterministic?

Currently, I use subrandom sequences or grids.

rmcgibbo:
I don't recall -- I suspect it is not, but that's mostly a guess.

rmcgibbo:
If it's non-deterministic only due the use of a PRNG, it may be possible to simply fix the seed, which would accomplish the same purpose.

kyleabeauchamp:
It looks like the only remaining test failure is explicit solvent src, which is already slated for renovation.  I'm going to merge the current work so I can file a new pr.

jchodera:
Whoops, this wasn't ready to merge yet---sorry for not marking it WIP!

jchodera:
> It looks like the only remaining test failure is explicit solvent src, which is already slated for renovation. I'm going to merge the current work so I can file a new pr.

I've got a more general fix for `src-explicit`.  Let me open this.

PR #57, opened 2015-05-10
Added CustomExternalForcesTestSystem and added options to LJCluster.

kyleabeauchamp:
CustomExternalForcesTestSystem will be useful for testing things related force groups---currently None of our testsystems do anything with force groups.  

The extra options in LJCluster allow us to use cutoffs and switches in the LJCluster test system.  This may be useful for debugging various switching and cutoff issues.


jchodera:
Awesome!

kyleabeauchamp:
Only test failure was src explicit, so merging.

PR #58, opened 2015-05-10
Removed switch flag

kyleabeauchamp:
I discovered that the `switch` flag had incorrect docstrings in several locations, leading to it being described as either a boolean or a length quantity.  

Rather than allow both a switch and a switch_width flag, here I've changed our testsystems to check whether `switch_width` is `None` for determining whether to apply a switching function.  IMHO it is easier for us to keep track of a single flag.  I think this will prevent further confusion down the line, although I am happy to reconsider.


jchodera:
Good catch!

kyleabeauchamp:
Looks like the only failure is again src explicit, merging.

PR #60, opened 2015-05-11
Fix for which precision models are tested when comparing platforms

jchodera:
This fixes the list of precision models tested when comparing platforms when trying to break down by force components.


jchodera:
This fixes the issue in my tests.

PR #62, opened 2015-05-12
Added explicit classes for truncated and switched LennardJonesFluid

jchodera:
Also corrected switch and cutoff distances.


kyleabeauchamp:
Looks good, only test failure is src

jchodera:
I wonder if we should be using `openmm-dev` in the travis tests instead of `openmm`?

kyleabeauchamp:
I have no strong preference either way.

PR #64, opened 2015-05-16
Added `shift` flag for LennardJonesFluid to enable shifted potential.

jchodera:
`LennardJonesFluid(shift=True)` now gives a shifted potential.


jchodera:
The only test failure is `SrcExplicit`, so merging.

Note that the `SrcExplicit` failure means that an updated `openmmtools-dev` is not being pushed to binstar.

PR #66, opened 2015-05-22
Ran pyflakes and autopep8 on constants, integrators, and testsystems

kyleabeauchamp:
My text editor was complaining at a lot of things, so I figured it was time to do a round of cleanup.

### DIFF
--- a/openmmtools/constants.py
+++ b/openmmtools/constants.py
@@ -1,4 +1,3 @@
-import numpy as np
 import simtk.unit as u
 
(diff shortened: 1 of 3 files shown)


jchodera:
Thanks!  

It's not recommended we automate this yet, right?

kyleabeauchamp:
I personally would rather run it manually once a year than maintain more CI scripts for running these things.

kyleabeauchamp:
OK, it looks like the only failure is src Explicit.

jchodera:
Maybe we should raise the threshhold for those failures?

PR #67, opened 2015-05-22
Charged LJ Fluid, Ewald tolerance, FCC grid

kyleabeauchamp:
1.  Allow alternating plus and minus charges in LJ fluid for testing PME in a simple liquid system
2.  Allow setting of Ewald tolerance in LJ Fluid and WaterBox
3.  Allow LJ Fluid to use a FCC grid for starting coordinates, which can be useful when trying to get a liquid testsystem that doesn't require pressure equilibration.

### DIFF
--- a/openmmtools/testsystems.py
+++ b/openmmtools/testsystems.py
@@ -48,6 +48,7 @@ import os
 import os.path
 import numpy as np
 import numpy.random
+import itertools
 
 import scipy
 import scipy.special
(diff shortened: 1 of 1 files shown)


jchodera:
Awesome.  Apologies for removing the grid code before---I thought that the subrandom initial conditions would be superior and allow any number of particles, but that doesn't appear to have been the case.

jchodera:
Is that really an FCC lattice, or just a primitive cubic lattice?

kyleabeauchamp:
Here is a screenshot from pymol of the nearest neighbors of a point.  The coordination number is 12, which should indicate FCC:

coord

kyleabeauchamp:
OK, the only travis failure is src explicit again.

PR #68, opened 2015-05-28
Added optional hydrogenmass to alanine testsystems.

kyleabeauchamp:
The option was already present in the ExplicitAlanine


kyleabeauchamp:
Tests pass locally and on travis except for the known energy difference

PR #69, opened 2015-05-29
Disabled OpenMM platform comparison tests

jchodera:
These are just slowing down travis and not helping us debug `openmmtools` issues, so I've disabled the nosetests (while leaving the command-line tool we can use to test OpenMM).


kyleabeauchamp:
LGTM

PR #80, opened 2015-07-05
Fixed parmed dependency for new parmed 2.0 release.

jchodera:
This adds `parmed` to the `devtools/conda-recipe` dependencies (for testing) and fixes the imports for ParmEd 2.0.


jchodera:
Will cut a 0.7.1 release for this after merging.

PR #84, opened 2015-08-23
Removed alchemically modified systems, which are now tested as part of Yank

jchodera:
The alchemical modification framework on [yank](http://github.com/choderalab/yank) has moved out of sync with the version in this repository, so I have removed the alchemical system tests in favor of having all of these in yank alone.


jchodera:
Still trying to get appveyor to work here.

PR #101, opened 2015-10-27
Various small bug fixes

jchodera:


Lnaden:
I'm going to close this PR, its a couple years old at this point and the codebase has changed a fair amount. If these bugs are still around, we can deal with that then.

Issue #85, opened 2015-08-24
Implement FIRE minimizer as a CustomIntegrator


jchodera:
The [FIRE](http://dx.doi.org/10.1103/PhysRevLett.97.170201) scheme from

> Structural relaxation made simple.
> PRL 97:170201, 2006
> http://dx.doi.org/10.1103/PhysRevLett.97.170201

seems like it might be an especially simple minimization scheme that can be implemented as a `CustomIntegrator`.

PR #102, opened 2015-11-03
[WIP] Implement FIRE minimizer as CustomIntegrator

jchodera:
This PR implements the [FIRE](http://dx.doi.org/10.1103/PhysRevLett.97.170201) minimizer (suggested in #85) from
> Structural relaxation made simple.
> PRL 97:170201, 2006
> http://dx.doi.org/10.1103/PhysRevLett.97.170201
which uses a simple heuristic scheme that respects constraints.


proteneer:
Interested to know if you guys are still working on this. I'd like to find a way to get rid of the conditional inertia check: P:=F.v < 0 to make it differentiable

jchodera:
I think @hannahbrucemacdonald and @jaimergp were looking into this last.

Can't you just use a soft tanh-like function that scales the timestep by a continuously-variable amount depending on the sign and value of the momentum?

proteneer:
I think that's definitely something we can look into, i.e. replacing the if/then (aka. heaviside function) with a tanh heaviside function.

PR #107, opened 2016-08-04
Add CB7:B2 host-guest test system

jchodera:
This PR adds a host-guest system in three different environments:
* `HostGuestVacuum`
* `HostGuestImplicit`
* `HostGuestExplicit`

The current model system uses cucurbit[7]uril (CB7) binding to the B2 ligand, from the excellent tutorial from Michael Gilson and Andrew Fenley:
http://ambermd.org/tutorials/advanced/tutorial21/


jhprinz:
Can we use this for OPS as well?

jchodera:
That's why I created it!

Still testing.

PR #112, opened 2016-08-08
Accumulate pseudowork in `VVVRIntegrator`

maxentile:
Adds a global variable that accumulates the pseudowork performed by the integrator. The pseudowork of a step is `W = ∆E - Q`: the total change in energy minus the heat exchanged with the bath during the velocity randomizations.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -670,9 +670,9 @@ class VVVRIntegrator(mm.CustomIntegrator):
         timestep : numpy.unit.Quantity compatible with femtoseconds, default: 1.0*simtk.unit.femtoseconds
            The integration timestep.
         monitor_heat : boolean, default: False
-           Compute the heat exchanged with the bath in each step, in the global `heat`.
+           Accumulate the heat exchanged with the bath in each step, in the global `heat`.
         monitor_work : boolean, default: False
-           Accumulate the shadow work of each step in the global `shadow_work`.
+           Accumulate the shadow work of each step, in the global `shadow_work`.
 
         Notes
         -----
@@ -718,18 +718,27 @@ class VVVRIntegrator(mm.CustomIntegrator):
         self.addPerDofVariable("sigma", 0)
         self.addPerDofVariable("x1", 0)  # position before application of constraints
         
-        # book-keeping variables
-        if monitor_work:
-            monitor_heat = True
-        
-        if monitor_heat:
+        # bookkeeping variables        
+        if monitor_heat and monitor_work:
             self.addGlobalVariable("heat", 0)
             self.addGlobalVariable("kinetic_energy_0", 0)
             self.addGlobalVariable("kinetic_energy_1", 0)
             self.addGlobalVariable("kinetic_energy_2", 0)
             self.addGlobalVariable("kinetic_energy_3", 0)
-            self.addGlobalVariable("old_energy", 0)
-            self.addGlobalVariable("new_energy", 0)
+            self.addGlobalVariable("energy_before_symplectic", 0)
+            self.addGlobalVariable("energy_after_symplectic", 0)
+            self.addGlobalVariable("shadow_work", 0)
+        elif monitor_heat:
+            self.addGlobalVariable("heat", 0)
+            self.addGlobalVariable("kinetic_energy_0", 0)
+            self.addGlobalVariable("kinetic_energy_1", 0)
+            self.addGlobalVariable("kinetic_energy_2", 0)
+            self.addGlobalVariable("kinetic_energy_3", 0)
+        elif monitor_work:
+            self.addGlobalVariable("kinetic_energy_1", 0)
+            self.addGlobalVariable("kinetic_energy_2", 0)
+            self.addGlobalVariable("energy_before_symplectic", 0)
+            self.addGlobalVariable("energy_after_symplectic", 0)
             self.addGlobalVariable("shadow_work", 0)
 
         #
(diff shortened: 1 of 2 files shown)


jchodera:
Thanks!

There are some big efficiency losses associated with accumulating the shadow work via additional energy computations. What about adding optional arguments (e.g.`monitor_heat` and `monitor_work`, defaulting to `False`) that add these extra variables and steps only if asked?

maxentile:
Good point! This adds 2 potential energy evaluations and 4 kinetic energy evaluations to each integrator step.

Will make bookkeeping optional momentarily

jchodera:
Maybe also add a test for these options [here](https://github.com/choderalab/openmmtools/blob/9e2ce71111be286837e5748ec0d772033a502c04/openmmtools/tests/test_integrators.py)?

jchodera:
Thanks!

Why the names `pseudowork` and `d_heat`, out of curiosity? Maybe just `shadow_work` and `heat` would work better? (I think we moved away from "pseudowork" before the papers were published.)

jchodera:
Tests look great!

maxentile:
Ahh, I'll change it to "shadow_work" then! I had just had named it "pseudowork" to be consistent with [the docstring](https://github.com/choderalab/openmmtools/blob/0.8.0/openmmtools/integrators.py#L677).

I had named the other variable "d_heat" to help indicate it was only storing the heat exchanged with the bath in the current step, rather than an accumulated value like the shadow work. Do you think that would be clearer as just "heat"?

openmmtools/integrators.py
```diff
@@ -746,5 +771,19 @@ def __init__(self, temperature=298.0 * simtk.unit.kelvin, collision_rate=91.0 /
...
+        if monitor_heat:
+            self.addComputeSum("kinetic_energy_3", "0.5 * m * v * v")
+        
+        if monitor_heat:
+            self.addComputeGlobal("heat", "(kinetic_energy_1 - kinetic_energy_0) + (kinetic_energy_3 - kinetic_energy_2)")
+        
+        if monitor_work:
+            # accumulate shadow_work, where W_step = ∆E_step - Q_step
+            self.addComputeGlobal("new_energy", "kinetic_energy_3 + energy")
+            self.addComputeGlobal("shadow_work", "shadow_work + (new_energy - old_energy) - heat")
```
jchodera: Instead of computing the shadow work as the total change in energy minus the total heat, why not just compute the total energy immediately before and after the velocity Verlet step (which is incorrectly labeled "Metropolized symplectic step" in my comment on line 737)? This may be numerically more stable.

jchodera:
>  had named the other variable "d_heat" to help indicate it was only storing the heat exchanged with the bath in the current step, rather than an accumulated value like the shadow work. Do you think that would be clearer as just "heat"?

Ah! Why not keep accumulating the heat, so that you can reset `shadow_work` and `heat` to zero, run `nsteps` steps, and then read off the values of accumulated `shadow_work` and `heat` at the end?

If you also need to store the per-step shadow work and heat, what about `step_shadow_work` and `step_heat`? But I'm not sure we need that.

openmmtools/integrators.py
maxentile: Ahh, good point! I had just implemented it that way following this comment in the To-Do's: https://github.com/choderalab/openmmtools/blob/0.8.0/openmmtools/integrators.py#L683 , but I think the alternative you suggest is simpler, and also lets me control `monitor_work` and `monitor_heat` separately.

maxentile:
> Ah! Why not keep accumulating the heat, so that you can reset shadow_work and heat to zero, run nsteps steps, and then read off the values of accumulated shadow_work and heat at the end?

Yes, that would be better! Fixing now...

>If you also need to store the per-step shadow work and heat, what about step_shadow_work and step_heat? But I'm not sure we need that.

There's no case I can think of where I need that.

openmmtools/tests/test_integrators.py
```diff
@@ -111,3 +111,33 @@ def test_integrator_decorators():
...
+   ''' When `monitor_work==True`, assert that global `shadow_work` is initialized to zero and
+   reaches a nonzero value after integrating a few dozen steps.
+   
+   By default (`monitor_work=False`), assert that there is no global name for `shadow_work`. '''
+   
+   # test `monitor_work=True` --> accumulation of a nonzero value in global `shadow_work`
+   testsystem = testsystems.HarmonicOscillator()
+   system, topology = testsystem.system, testsystem.topology
+   temperature = 298.0 * unit.kelvin
+   integrator = VVVRIntegrator(temperature, monitor_work=True)
```
jchodera: This should be `integrators.VVVRIntegrator(temperature, monitor_work=True)`

openmmtools/tests/test_integrators.py
```diff
@@ -111,3 +111,33 @@ def test_integrator_decorators():
...
+   context = mm.Context(system, integrator)
+   context.setPositions(testsystem.positions)
+   context.setVelocitiesToTemperature(temperature)
+   assert(integrator.getGlobalVariableByName('shadow_work') == 0)
+   integrator.step(25)
+   assert(integrator.getGlobalVariableByName('shadow_work') != 0)
+   
+   # test default (`monitor_work=False`, `monitor_heat=False`) --> absence of a global `shadow_work`
+   integrator = VVVRIntegrator(temperature)
```
jchodera: This should be
```python
integrator = integrators.VVVRIntegrator(temperature)
```

maxentile: Thanks! I was having trouble finding where the problem was

maxentile:
Hooray! Tests pass. Should be ready to merge?

jchodera:
Go for it!

PR #113, opened 2016-08-08
[WIP] Add LangevinSplittingIntegrator

maxentile:
- Adding a generic Langevin Integrator, that accepts a splitting scheme string (like "BAOAB") and produces the corresponding integrator
- Currently minimal (doesn't include restraints or bookkeeping), until we settle on a good way to represent these schemes
- For simplicity, this currently doesn't check if positions have been updated since last `B` step call, but checking this could save unnecessary force evaluations. E.g. the integrator constructed using the splitting `BAOAB` will make two force calls per timestep, when it only needs to make one.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -655,42 +655,139 @@ class GHMCIntegrator(mm.CustomIntegrator):
         self.addComputeGlobal("naccept", "naccept + accept")
         self.addComputeGlobal("ntrials", "ntrials + 1")
 
-class VVVRIntegrator(mm.CustomIntegrator):
+class LangevinSplittingIntegrator(mm.CustomIntegrator):
+    """Integrates Langevin dynamics with a prescribed operator splitting.
 
+    One way to divide the Langevin system is into three parts which can each be solved "exactly:"
+        - R: Linear "drift"
+            Deterministic update of *positions*, using current velocities
+        - V: Linear "kick"
+            Deterministic update of *velocities*, using current forces
+        - O: Ornstein-Uhlenbeck
+            Stochastic update of velocities, simulating interaction with a heat bath
+
+    We can then construct integrators by solving each part for a certain timestep in sequence.
+
+    (We can further split up the V step by force group, evaluating cheap but fast-fluctuating
+    forces more frequently than expensive by slow-fluctuating forces.)
+
+    Examples
+    --------
+        - VVVR:
+            splitting="O V R V O"
+        - BAOAB:
+            splitting="V R O R V"
+        - g-BAOAB, with K_p=3:
+            splitting="V R R R O R R R V"
+        - g-BAOAB with solvent-solute splitting, K_r=K_p=2:
+            splitting="V0 V1 R R O R R V1 V1 R R O R R V1 V0"
     """
-    Create a velocity Verlet with velocity randomization (VVVR) integrator.
 
-    """
+    def __init__(self,
+                 splitting="V0 R O R V1",
+                 temperature=298.0 * simtk.unit.kelvin,
+                 collision_rate=91.0 / simtk.unit.picoseconds,
+                 timestep=1.0 * simtk.unit.femtoseconds,
+                 constraint_tolerance=1e-8
+                 ):
+        """Create a discrete-time integrator of Langevin dynamics with a prescribed operator splitting.
 
-    def __init__(self, temperature=298.0 * simtk.unit.kelvin, collision_rate=91.0 / simtk.unit.picoseconds,
-                 timestep=1.0 * simtk.unit.femtoseconds, monitor_heat = False, monitor_work = False):
-        """
-        Create a velocity verlet with velocity randomization (VVVR) integrator.
+        The integrator will accumulate the heat exchanged with the bath in each step, in the global `heat`
 
         Parameters
         ----------
+        splitting : string
+            Sequence of R, V, O steps executed each timestep
+            Handle multiple force groups by V0, V1,
         temperature : numpy.unit.Quantity compatible with kelvin, default: 298.0*simtk.unit.kelvin
-           The temperature.
+           Fictitious "bath" temperature
         collision_rate : numpy.unit.Quantity compatible with 1/picoseconds, default: 91.0/simtk.unit.picoseconds
-           The collision rate.
+           Collision rate
         timestep : numpy.unit.Quantity compatible with femtoseconds, default: 1.0*simtk.unit.femtoseconds
-           The integration timestep.
-        monitor_heat : boolean, default: False
-           Accumulate the heat exchanged with the bath in each step, in the global `heat`.
-        monitor_work : boolean, default: False
-           Accumulate the shadow work of each step, in the global `shadow_work`.
+           Integration timestep
 
-        Notes
+        References
+        ----------
+        Benedict Leimkuhler, Charles Matthews, 2012
+        Rational Construction of Stochastic Numerical Methods for Molecular Sampling
+        https://arxiv.org/abs/1203.5428
+        """
+        # Compute constants
+        kT = kB * temperature
+        gamma = collision_rate
+
+        splitting = splitting.upper().split()
+        # there are at most 32 force groups in the system
+        assert (set(splitting).issubset(set("RVO").union(set(["V{}".format(i) for i in range(32)]))))
+        assert ("R" in splitting)
+        assert ("O" in splitting)
+
+        # Count how many times each step appears, so we know how big each R/V/O substep will be
+        n_R = sum([letter == "R" for letter in splitting])
+        n_O = sum([letter == "O" for letter in splitting])
+        # Each force group should be integrated for a total length equal to dt
+        fgs = set([step for step in splitting if step[0] == 'V'])
+        n_Vs = dict()
+        for fg in fgs: n_Vs[fg] = sum([step == fg for step in splitting])
+
+        # Define substep functions
+        def R_step():
+            self.addComputePerDof("x", "x + (dt / {}) * v".format(n_R))
+            self.addComputePerDof("x1", "x")  # save pre-constraint positions in x1
+            self.addConstrainPositions()  # x is now constrained
+            self.addComputePerDof("v", "v + (x - x1) / (dt / {})".format(n_R))
+            self.addConstrainVelocities()
+
+        def V_step(fg):
+            self.addComputePerDof("v", "v + ((dt / {}) * f{} / m)".format(n_Vs[fg], fg))
+            self.addConstrainVelocities()
+
+        kinetic_energy = "0.5 * m * v * v"
+        def O_step():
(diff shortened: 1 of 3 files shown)


jchodera:
> For simplicity, this currently doesn't check if positions have been updated since last B step call, but checking this could save unnecessary force evaluations. E.g. the integrator constructed using the splitting BAOAB will make two force calls per timestep, when it only needs to make one.

I think `energy` and `forces` already cache the energy and forces, only updating them if the positions or box state have changed, so no need to worry about this.

maxentile:
>I think energy and forces already cache the energy and forces, only updating them if the positions or box state have changed, so no need to worry about this.

Oh, that's awesome! I didn't realize OpenMM handled that automatically.

openmmtools/integrators.py
```diff
@@ -748,3 +748,144 @@ def __init__(self, temperature=298.0 * simtk.unit.kelvin, collision_rate=91.0 /
...
+        n_A = sum([letter == 'A' for letter in splitting])
+        n_B = sum([letter == 'B' for letter in splitting])
+        n_O = sum([letter == 'O' for letter in splitting])
+
+        # For now, don't split up the O step -- will relax this later!
+        assert (n_O == 1)
+
+        # get strings to pass to self.addComputePerDof(variable, expression) during each A/B/O step
+        update_equations = dict()
+        update_equations['A'] = ("x", "x + (dt / {n_A}) * v".format(n_A=n_A))
```
jchodera: We can probably generalize these update equations to include the potential for multiple lines so we can easily add constraints.

jchodera: We may also want to have aliases so that we can use 'VRORV' instead, etc.

jchodera:
In the [velocity Verlet integrator implementation](http://docs.openmm.org/7.0.0/api-python/generated/simtk.openmm.openmm.CustomIntegrator.html#simtk.openmm.openmm.CustomIntegrator), a correction is required as part of the RATTLE constraint:
```python
# velocity Verlet
integrator.addPerDofVariable("x1", 0);
integrator.addUpdateContextState();
integrator.addComputePerDof("v", "v+0.5*dt*f/m");
integrator.addComputePerDof("x", "x+dt*v");
integrator.addComputePerDof("x1", "x");
integrator.addConstrainPositions();
integrator.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt"); # note '(x-x1)/dt' correction to velocities
integrator.addConstrainVelocities();
```
Note the `(x-x1)/dt` correction to the velocities applied before `addConstrainVelocities`. See [Ben Leimkuhler's paper for details](http://rspa.royalsocietypublishing.org/content/472/2189/20160138) of where and why RATTLE enters. (I believe the book from [Gabriel Stoltz](https://www.amazon.com/Free-Energy-Computations-Mathematical-Perspective/dp/1848162472) also has a lucid treatment in Chapter 3.)

jchodera:
I'm not entirely sure how we generalize constraints to different splittings, but you should be able to figure this out from examining [the Leimkuhler paper](http://rspa.royalsocietypublishing.org/content/472/2189/20160138) since it presents the constraint update schemes for both BAOAB and OBABO.

openmmtools/integrators.py
maxentile: Thanks, John!

> We can probably generalize these update equations to include the potential for multiple lines so we can easily add constraints.

Hmm, I'll need to think about the best way to represent each step here. Here I've done two different things: represent the actual A/B/O updates as just update equation strings, and placed the logic for deciding when to apply constraints / collect bookkeeping info inside convenience functions that are called during each substep (`compute_pre_step_energies()`, `apply_constraints()`, `compute_post_step_energies()`, and `accumulate_heat_or_work()`).

> We may also want to have aliases so that we can use 'VRORV' instead, etc.

Okay! These would be synonymous, right?

Related issue: would there be much benefit in representing the S/A splitting also? (eq. 7 in http://arxiv.org/abs/1304.3269)

maxentile:
Thank you for the pointers! Okay, then the current implementation I added is almost certainly not correct when constraints are active. I'll take a closer look through these references to figure out when which constraints / corrections should be applied...

jchodera:
The [RESPA integrator example](https://github.com/choderalab/openmmtools/blob/master/openmmtools/respa.py) shows how a general multiple timestep scheme can be constructed in which different force groups and number of steps can be specified for each level of the MTS integrator. I think we could also generalize this to let us specify particle indices integrated in each level as well.

For example, you can create a per-dof variable for each integrator level that is `1` if the atoms should be propagated and `0` otherwise:
```python
        values = list()
        for index in range(nparticles):
            value = Vec3(1,1,1) if (index in atom_indices[mts_level]) else Vec3(0,0,0)
            values.append(value)
        varname = 'atomset%d' % mts_level
        integrator.addPerDofVariable(varname, 0)
        integrator.setPerDofVariableByName(varname, values)
```
You can then apply each step to just the atoms that are affected in this MTS level, e.g.
```python
integrator.addComputePerDof("v", "v+atomset"+str(mts_level)+"*0.5*dt*f"+str(force_groups)+"/m")
```

jchodera:
> Okay! These would be synonymous, right?

Yes, we'd just have `V` and `R` as synonyms for `B` and `A`.

> Related issue: would there be much benefit in representing the S/A splitting also? (eq. 7 in http://arxiv.org/abs/1304.3269)

No, I don't think we're interested in these splittings right now.

maxentile:
Before I can implement options like multiple timestepping with constraints generically and correctly, I'll need to really wrap my head around the case of [g-BAOAB](http://rspa.royalsocietypublishing.org/content/472/2189/20160138).

However, I'm having a hard time translating the algorithm definition from the Appendix into a `CustomIntegrator`.

Here's my current attempt: it blows up within a few hundred steps when simulating a constrained water box (`openmmtools.testsystems.WaterBox(constrained=True)`). I'm pretty sure I've made a mistake somewhere in the inner loop of `A()`, but I'm stuck trying to find/fix it. If someone can help spot where I'm going wrong here, I would greatly appreciate it!

```python
class GeodesicBAOABIntegrator(mm.CustomIntegrator):
    def __init__(self, K_r=3, temperature=298.0 * simtk.unit.kelvin, collision_rate=91.0 / simtk.unit.picoseconds, timestep=1.0 * simtk.unit.femtoseconds):
        kT = kB * temperature
        gamma = collision_rate
        a = numpy.exp(-gamma * timestep)
        inner_step_size = 1.0 / (2 * K_r)

        super(GeodesicBAOABIntegrator, self).__init__(timestep)
        self.addGlobalVariable("kT", kT)
        self.addGlobalVariable("a", a)
        self.addGlobalVariable("b", numpy.sqrt(kT * (1 - a**2)))
        self.addPerDofVariable("x1", 0)
            
        def B():
            self.addComputePerDof("v", "v + 0.5*dt*f/m")
            self.addConstrainVelocities()
        
        def A():
            for i in range(K_r):
                self.addComputePerDof("x", "x + v*(dt*{step_size})".format(step_size = inner_step_size))
                self.addComputePerDof("x1", "x")
                self.addConstrainPositions()
                self.addComputePerDof("v", "v + (x-x1)/(dt*{step_size})".format(step_size = inner_step_size))
                self.addConstrainVelocities()
                
        def O():
            self.addComputePerDof("v", "a * v + b * gaussian / sqrt(m)")
            self.addConstrainVelocities()
        
        self.addUpdateContextState()
        B()
        A()
        O()
        A()
        self.addUpdateContextState()
        B()
```
Here is the algorithm definition from the paper, for reference: 

gbaoab

where

a

jchodera:
Also, I should ask if you are using the WaterBox test system from openmmtools, and if you minimized it first!

jchodera:
Also, I'm not sure you need the extra velocity constraint step in A() since the component normal to the velocity was already removed and we are already correcting the velocities based on the position constraint, but I don't see how it should hurt (other than speed).

jchodera:
Thanks!

Is this he version currently checked into the PR?

jchodera:
One thing to watch out for: Unit conversion is not always handled automatically when you are forming expressions in CustomIntegrators. I would be sure to use .value_in_unit_system(unit.md_unit_system) to explicitly concert all your variables (like kT and timestep) to natural OpenMM units when setting these in CustomIntegrator expressions just to be safe. I've been bitten by this before.

jchodera:
One quick observation that probably isn't the cause here, but should be fixed: We should only call self.addUpdateContextState() at the very beginning of the integrator step. This allows forces like MonteCarloBarostat and the center of mass motion removal to do their thing. The expectation is that it will only be called once per step, but you have to be careful where you put it since it may change positions, velocities, box vectors, and invalidate forces (requiring their recomputation, which can be slow).

maxentile:
Thanks, John-- sorry for my slow response!

> Also, I should ask if you are using the WaterBox test system from openmmtools, and if you minimized it first!

Yes! `openmmtools.testsystems.WaterBox(constrained=True)` and minimized to the default tolerance before integrating.

> Also, I'm not sure you need the extra velocity constraint step in A() since the component normal to the velocity was already removed and we are already correcting the velocities based on the position constraint, but I don't see how it should hurt (other than speed).

I would think that without this, `K_r`>1 would always push us off the momentum-constraint manifold. I'll check numerically if this second velocity constraint step is actually affecting the velocities, and remove it if redundant.

> One thing to watch out for: Unit conversion is not always handled automatically when you are forming expressions in CustomIntegrators. I would be sure to use .value_in_unit_system(unit.md_unit_system) to explicitly concert all your variables (like kT and timestep) to natural OpenMM units when setting these in CustomIntegrator expressions just to be safe. I've been bitten by this before.

Great point! I'll include units.

>One quick observation that probably isn't the cause here, but should be fixed: We should only call self.addUpdateContextState() at the very beginning of the integrator step. This allows forces like MonteCarloBarostat and the center of mass motion removal to do their thing. The expectation is that it will only be called once per step, but you have to be careful where you put it since it may change positions, velocities, box vectors, and invalidate forces (requiring their recomputation, which can be slow).

Ahh, good point! I'll remove the second call. I had only included it to be doubly sure that forces in the second `B` step would be computed using the current positions, but that's unnecessary.

jchodera:
No need to check if removing the additional calls to constrain the velocity
is problematic----there should be no harm in leaving these in!

openmmtools/integrators.py
```diff
@@ -748,3 +748,144 @@ def __init__(self, temperature=298.0 * simtk.unit.kelvin, collision_rate=91.0 /
...
+        update_equations['B'] = ("v", "v + (dt / {n_B}) * f".format(n_B=n_B))
```
jchodera: `f` should be `f/m`

jchodera:
* Checked in a fix for omitted mass division in velocity update
* Added velocity correction for RATTLE

maxentile:
Completed in https://github.com/choderalab/openmmtools/pull/134

Issue #82, opened 2015-07-06
Add setTemperature() method to VVVRIntegrator (and other integrators)


jchodera:
We should be able to set the temperature for the `VVVRIntegrator` using a `setTemperature()` call, as in `LangevinIntegrator`. This can be done by ensuring these are context parameters.

PR #120, opened 2016-10-05
Implement `.getTemperature` and `setTemperature` for `VVVR` integrator

jhprinz:
This implements #82 for VVVR and the corresponding getter. This is more a convenience to get the temperature that is used.


jchodera:
Cool! 

I had lobbied for these functions to be put in at the `Integrator` level:

https://github.com/pandegroup/openmm/issues/1440

Maybe you want to chime in there if you think this would be useful?

andrrizzi:
Closing this since it was included as part of PR #130 .

PR #123, opened 2017-01-01
Add multiple GB model options to some systems

jchodera:
This PR extends some of the test systems to allow multiple `CustomGBForce`-based GB models supported by OpenMM's [`customgbforces.py`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/customgbforces.py) functionality to be used. This is a first step toward implementing general `CustomGBForce` support in [`alchemy`](http://github.com/choderalab/alchemy).

* `TolueneImplicit` and `HostGuestImplicit` now accept the `implicitSolvent` option
* Both classes also have variants like `TolueneImplicitHCT`, `TolueneImplicitOBC1`, etc. for all OpenMM-supported implicit solvent models.
* I've implemented the beginnings of a more flexible way of allowing parameters to be fed (as `kwargs`) to the class constructor. Only options that are present in the function signature for `AmberPrmtopFile.createSystem()` are passed on. This scheme (and some code) comes from @andrrizzi, who added this functionality to yank. Eventually, we can expand this scheme to all test systems once we refine it further, but for now, this is probably a good start.

### DIFF
--- a/openmmtools/tests/test_testsystems.py
+++ b/openmmtools/tests/test_testsystems.py
@@ -70,7 +70,7 @@ def check_properties(testsystem):
 def test_properties_all_testsystems():
     """Testing computation of analytic properties for all systems.
     """
-    testsystem_classes = testsystems.TestSystem.__subclasses__()
+    testsystem_classes = get_all_subclasses(testsystems.TestSystem)
     logging.info("Testing analytical property computation:")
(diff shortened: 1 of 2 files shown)


jchodera:
I've also generalized `LysozymeImplicit`, though I didn't create all the suffixed variations.

PR #127, opened 2017-02-07
LRU Context cache

andrrizzi:
This PR adds a module `cache.py` implementing a LRU cache for `Context`s that will be used by the new `MCMCMove` classes. It should be ready for review if all tests pass.

### DIFF
--- a/openmmtools/cache.py
+++ b/openmmtools/cache.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python
+
+# =============================================================================
+# MODULE DOCSTRING
+# =============================================================================
+
+"""
+Provide cache classes to handle creation of OpenMM Context objects.
+
+"""
+
+
+# =============================================================================
+# GLOBAL IMPORTS
+# =============================================================================
+
+import collections
+
+
+# =============================================================================
+# GENERAL LRU CACHE
+# =============================================================================
+
+class LRUCache(object):
+    """A simple LRU cache."""
+
+    def __init__(self, capacity, time_to_live=None):
+        self._data = collections.OrderedDict()
+        self._capacity = capacity
+        self._ttl = time_to_live
+        self._n_access = 0
+
+    def __getitem__(self, key):
+        self._n_access += 1
+
+        # When we access data, push element at the
+        # end to make it the most recently used.
+        entry = self._data.pop(key)
+        self._data[key] = entry
+
+        # Update expiration and cleanup expired values.
+        if self._ttl is not None:
+            entry.expiration = self._n_access + self._ttl
+            self._remove_expired()
+        return entry.value
+
+    def __setitem__(self, key, value):
+        self._n_access += 1
+
+        # When we access data, push element at the
+        # end to make it the most recently used.
+        try:
+            self._data.pop(key)
+        except KeyError:
+            # Remove first item if we hit maximum capacity.
+            if len(self._data) >= self._capacity:
+                self._data.popitem(last=False)
+
+        # Determine expiration and clean up expired.
+        if self._ttl is None:
+            ttl = None
+        else:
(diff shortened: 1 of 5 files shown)


andrrizzi:
Test fixed! This is ready for review.

openmmtools/cache.py
```diff
@@ -0,0 +1,352 @@
...
+    to the cache. Both read and write operations count as an access, but
+    only successful reads (i.e. those not raising KeyError) increment the
+    counter.
+
+    Parameters
+    ----------
+    capacity : int, optional
+        Maximum number of elements in the cache. When set to None, the
+        cache has infinite capacity (default is None).
+    time_to_live : int, optional
```
jchodera: Does it make sense to later be able to add an alternative `time_to_live` measured in actual wall clock time? I guess we could interpret this as number of accesses if it is an `int` or later interpret it as wall clock time if a `simtk.unit.seconds` compatible quantity is specified.

andrrizzi: > I guess we could interpret this as number of accesses if it is an int or later interpret it as wall clock time if a `simtk.unit.seconds` compatible quantity is specified

This sounds like a solution. I decided to go with number of accesses here because the wall clock time between each iterations is depends on the system and platform, while the number of accesses intuitively should be more or less the same (although it will scale with the total number of MC moves and thermodynamic states explored), but I am happy to implement a wall clock-based system if you think it's preferable.

jchodera: I think we can always add other behaviors later if we find they will be useful. I was originally wondering if it would be useful to have a few optional `time_to_live` arguments, but I think this single argument that could later have multiple interpretations based on the units attached to the value passed is just fine.

openmmtools/cache.py
```diff
@@ -0,0 +1,352 @@
...
+            context = self._lru[context_id]
+        except KeyError:
+            context = thermodynamic_state.create_context(integrator, platform)
+            self._lru[context_id] = context
+        context_integrator = context.getIntegrator()
+
+        # Update state of system and integrator of the cached context.
+        self._copy_integrator_state(integrator, context_integrator)
+        thermodynamic_state.apply_to_context(context)
+        return context, context_integrator
```
Lnaden: So I have an issue here. `Integrator` in this case is identical to the `context_integrator` any time a new context is made (they point to the same object). That would mean that you could not use the same integrator to make multiple contexts at slightly different thermodynamic states, you would need to recreate new integrators outside of this function every time. Is that the behavior we want?

jchodera: As soon as an `Integrator` is passed to a `Context`, it is owned by the created `Context`. You can't have an `Integrator` associated with multiple `Context` objects.

Lnaden: That's my point. By the logic here whatever code is passing integrators into this function has to know this behavior exists. What is the point of checking compatible integrators since the first and only time it could ever be used is in the creation of a context? Integrators does not seem like something that would be helpful to use for checking compatible contexts since they are 1:1

andrrizzi: Two integrators are compatible if they differ by only few parameters (e.g. step size, temperature). In this case `ContextCache` recycle the same `Context` and simply set the parameters of its `Integrator` to the ones requested.

This is why I made `get_context()` return a tuple `(Context, Integrator)` instead of just the `Context`. I thought this would make it clear (on top of the documentation) that the returned integrator must be used.

jchodera: I see your point. I think this behavior is documented, though. You always want to use the `context` and `integrator` objects *returned* by `get_context`, never the original integrator.

andrrizzi:
I've updated the `README` with a rough list of the new classes. If the integrator compatibility issues are clear, I'd merge this.

PR #128, opened 2017-02-08
[WIP] Distributed computing with celery

jchodera:
@andrrizzi : Here's the very basic test code I was playing with, in case you find it useful. This doesn't necessarily have to be merged, but might at least illustrate something along the lines of what I was thinking.

I haven't tested this on the cluster yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,24 @@ The `openmmtools.testsystems` module contains a large suite of test systems---in
 If differences in energies in excess of `ENERGY_TOLERANCE` (default: 0.06 kcal/mol) are detected, these systems will be serialized to XML for further debugging.
 
 This is installed onto the command line when the repository is installed.
+
+## Distributed computing with OpenMM
+
+You can use `openmmtools.distributed` to run various OpenMM tasks in a distributed manner using [`celery`](http://www.celeryproject.org/).
+
+```python
+from simtk import openmm, unit
+from openmmtools import testsystems, distributed
+
+# Set up a test system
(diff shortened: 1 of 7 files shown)


jchodera:
We'll still need a way to start celery workers on individual GPUs. I bet we could do this with something like
```bash
CUDA_VISIBLE_DEVICES=0 celery -A openmmtools.distributed worker -l info --concurrency=1 &
CUDA_VISIBLE_DEVICES=1 celery -A openmmtools.distributed worker -l info --concurrency=1 &
CUDA_VISIBLE_DEVICES=2 celery -A openmmtools.distributed worker -l info --concurrency=1 &
CUDA_VISIBLE_DEVICES=3 celery -A openmmtools.distributed worker -l info --concurrency=1 &
```
though @pgrinaway may have better ideas for how best to do this with multiple GPUs on a node.

It looks like there's also a way to specify worker queues with the `--queues` flag, like `--queues gpu` vs `--queues cpu`. Documentation on these can be found [here](http://docs.celeryproject.org/en/latest/userguide/routing.html#id2).

andrrizzi:
Thanks! I'll take a look at this tomorrow.

jchodera:
This is still very much test code for experimenting. I think the next steps are:
* Try to construct some sort of benchmark example that tests local vs remote execution of a replica exchange like operation on a realistic test system (e.g. Src in explicit solvent) to see how well things perform
* Make it easy to try both `celery` and `redis`

PR #130, opened 2017-02-10
Implement temperature setter and getter for integrators coupled to a heat bath

### LINKED ISSUE #129: Add getTemperature and setTemperature methods to integrators coupled to heat bath
When creating/applying itself to a `Context`, the new `ThermodynamicState` class currently adds an `AndersenThermostat` if the integrator does keep the temperature of the system constant. `ThermodynamicState` considers thermostated integrators those that expose `get/setTemperature()` methods. Any objection if I add these in the PR I'm currently working on?

andrrizzi:
Fix #129 .

I realize now this is a more general version of the old PR #120 . @jhprinz is it ok if I close your PR?

I'll merge this first into the `mcmoves` branch because I need it there.

I still need to add a test and some examples. Here I've added a new mixin class `MetropolizedIntegrator` that
- adds setter and getter for temperature to its children.
- implements a utility function `addTemperatureDependentComputePerDofs` which wraps the kT-dependent per-DOF constants into an if-block executed only when the temperature is set.

I thought the second would be useful to automate the computation of the per-DOF velocity stddev (i.e. sigma_v) only once, but I can revert those changes if you don't like it.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1,6 +1,6 @@
-#=============================================================================================
+# ============================================================================================
 # MODULE DOCSTRING
-#=============================================================================================
+# ============================================================================================
 
 """
 Custom integrators for molecular simulation.
@@ -31,9 +31,9 @@ this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """
 
-#=============================================================================================
+# ============================================================================================
 # GLOBAL IMPORTS
-#=============================================================================================
+# ============================================================================================
 
 import numpy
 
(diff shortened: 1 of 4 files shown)


andrrizzi:
This is ready for review!

The PR Travis tests are expected not to pass because I'm merging to the `mcmoves` branch which is still undone.

I still have a couple of stylistic changes I want to do in `test_integrators.py` (there are mixed 3 and 4-spaces tabs), but I'll wait after the review to not mess the diff up.

Are there better names for the base class (i.e. `MetropolizedIntegrator`)?

jchodera:
Thanks for putting this together!

> Are there better names for the base class (i.e. MetropolizedIntegrator)?

Hm. `VVVRIntegrator` isn't actually Metropolized, so it is awkward for it to inherit from a class called `MetropolisIntegrator`. What about using mixins where `MetropolizedIntegrator` means it is Metropolized and has get/reset methods for acceptance statistics, while `ThermostattedIntegrator` means it has temperature set/get methods?

andrrizzi:
> `ThermostattedIntegrator` means it has temperature set/get methods?

Sounds better, thanks! I'll leave the common class for get/reset methods for a future PR since I don't need it right now, but I'll change the name of the existing base class to `ThermostatedIntegrator`.

jchodera:
OK! Presumably `is_metropolized` will become `is_thermostatted` then?

andrrizzi:
Yep! I'll change everything consistently to the new name.

andrrizzi:
Done! I'll merge when tests will pass.

jchodera:
Thanks!

PR #131, opened 2017-02-14
MCMC framework module, fixes to ContextCache and states, merge #130 to master

andrrizzi:
Ready for review.

Beside the changes in PR #130 to integrators, this PR adapts the MCMC framework to the new states and caching classes. The changes w.r.t. `openmmmcmc` are the following.
- Instead of taking lists and dict, the `MCMCSampler` takes `SequenceMove`s and `WeightedMove`s. I thought this could be a more sustainable design and allow us to easily experiment with different protocols in the future (e.g. adaptive).
- `MCMCMove.apply` does not return a new `SamplerState` but modifies it in place. This is because the framework theoretically allows changes to the `ThermodynamicState` too. If we need it to return the results as a tuple for Celery we can very easily write a wrapper.
- Added a new base class `IntegratorMove` that removes a lot of duplicate code and makes it easy to define new `MCMCMove`s based on OpenMM integrators.
- The `Platform` to use is now bound to a `ContextCache` instance so that we won't have to pass it around from object to object all the time. If for some reason the user wants to use two different platform in the same run, he can set different `ContextCache`s to different `MCMCMove`s.
- PEP8 fixes and reworked documentation and examples with the new states and cache classes.

I have already updated the README.md with the MCMC classes.

For the record, I had to play with the travis script for tests because when installing pymbar, it was considering build `np19` > `np111` and I was getting a `libgfortran.so not found` issue because numpy would get downgraded for incompatibility with `pymbar`.


README.md
```diff
@@ -22,6 +22,15 @@ This repository contains a number of additional integrators for OpenMM in `openm
 
 The `openmmtools.testsystems` module contains a large suite of test systems---including many with simple exactly-computable properties---that can be used to test molecular simulation algorithms
 
+## Monte Carlo moves
```
jchodera: Let's call this "Markov chain Monte Carlo proposal schemes and move compositions" or something.

README.md
```diff
@@ -22,6 +22,15 @@ This repository contains a number of additional integrators for OpenMM in `openm
...
+An implementation of an `MCMCMove` encodes how to propagate an OpenMM `System` to generate a new sample. Different `MCMCMove`s can be combined for more advanced schemes.
+- `LangevinDynamicsMove`: Langevin dynamics segment as a (pseudo) Monte Carlo move.
```
jchodera: Maybe add "(WARNING: Does not preserve the true target distribution.)"

README.md
```diff
@@ -22,6 +22,15 @@ This repository contains a number of additional integrators for OpenMM in `openm
...
+- `HMCMove`: Assigns velocities from the Maxwell-Boltzmann distribution and propagate through velocity Verlet steps.
+- `GHMCMove`: Generalized hybrid Monte Carlo Markov chain Monte Carlo.
+- `MonteCarloBarostatMove`: Attempts to update the box volume using Monte Carlo updates.
```
jchodera: "using Metropolis-Hastings Monte Carlo updates"

andrrizzi:
Thanks! I'll fix your comments and merge.

PR #132, opened 2017-02-14
Import alchemy module and add .gitignore file

andrrizzi:
Ready to be reviewed.

Here is the import from the `alchemy` repo as it is. I'd refactor this in a second PR to make the diff clearer since there are already a lot of commits here.

The only thing I changed are some PEP8 stuff and the addition of a `.gitignore` file that was missing in this repo.

The only failing test is already fixed in #131 .


jchodera:
Thanks! Let's make sure to add a deprecation warning to the `alchemy` repo that people should now use `from openmmtools import alchemy` and that development is continuing here!

andrrizzi:
Good idea! I'll add it after I've completed the refactoring here.

PR #133, opened 2017-02-22
New alchemy API

andrrizzi:
This is a good point for a first review since the diff will become messier after this. Tests are not expected to pass right now.

This adds to `AlchemicalState` an implementation of the `IComposableState` interface. After this I plan to add a bunch of test fixes and changes to `AbsoluteAlchemicalFactory` that will break backwards compatibility with alchemy 1.X. In particular
- Fix PEP8, which means convert camel case functions to underscore names.
- Remove alchemical functions from `AbsoluteAlchemicalFactory` which have been implemented more generally in `AlchemicalState` for all the parameters.
- Remove all the functions that have been transferred to `AlchemicalState` such as `perturbSystem` (now `AlchemicalState.apply_to_system`) and `perturbContext` (now `AlchemicalState.apply_to_context`).
- Some refactoring to avoid depending on a cached copy of the alchemical system in the alchemical factory that can become inconsistent with the other internal variables.

Any objections to these changes?


jchodera:
Awesome! Reviewing now, but some comments:

> Fix PEP8, which means convert camel case functions to underscore names.

Definitely sounds good.

> Remove alchemical functions from AbsoluteAlchemicalFactory which have been implemented more generally in AlchemicalState for all the parameters.
> Remove all the functions that have been transferred to AlchemicalState such as perturbSystem (now AlchemicalState.apply_to_system) and perturbContext (now AlchemicalState.apply_to_context).

These sound necessary given the refactoring of `AlchemicalState`, but is this API sufficiently general to allow us to once again introduce alchemical states that differ not by a global context variable but instead by a scaling of charges? We had discussed potentially adding this optional capability back as a means of both (1) avoiding the statistical efficiency loss for PME by not being able to include the reciprocal space contributions for the ligand, and (2) avoiding some of the speed loss from recoding all the electrostatics as `CustomNonbondedForce` terms.

Ideally, we'd be able to support both kinds of `AlchemicalState` updates---those that involve just changing `Context` parameters, and those that involve `updateParametersInContext` to modify charges---but it sounds like there wouldn't necessarily be a way to convert a system that uses scaled charges into, say, a system that softens the Lennard-Jones interactions via `CustomNonbondedForce` via `apply_to_context`.

> Some refactoring to avoid depending on a cached copy of the alchemical system in the alchemical factory that can become inconsistent with the other internal variables.

This is left over from when `deepcopy` was excruciatingly slow and there was no way to delete `Force` objects, so I'm happy to see this go if it simplifies the code.

openmmtools/states.py
```diff
@@ -1702,6 +1702,11 @@ def _read_context_state(self, context, check_consistency):
 # COMPOUND THERMODYNAMIC STATE
 # =============================================================================
 
+class ComposableStateError(Exception):
+    """Error raised by a ComposableState."""
+    pass
+
+
 class IComposableState(utils.SubhookedABCMeta):
```
jchodera: What does the `I` prefix notation mean here?

jchodera:
Can you remind me if there is an example of the new proposed API for this somewhere already?

openmmtools/states.py
andrrizzi: `I` usually stands for `Interface`. It's not standard but very common: http://stackoverflow.com/questions/10723839/python-naming-conventions-for-interfaces-and-abstract-classes

andrrizzi:
> is this API sufficiently general to allow us to once again introduce alchemical states that differ not by a global context variable but instead by a scaling of charges?

I don't think there will be problems in extending it to this. At least, I don't see extra challenges that come with the new API that would not be present if we maintained everything inside the alchemical factory. To be clearer, `AlchemicalState.lambda_electrostatics` is a property, and the mechanism used by `apply_to_system` and `apply_to_context` to translate the value of that property to the system/context is hidden from the user. It can be both a global variable or a a scale factor. Which behavior to adopt could be either controlled by an argument in the constructor or simply inferred from the `System` (e.g. if you can't find a global parameter for `lambda_electrostatics` just scale the charges).

Did you have a particular case in mind that you thought could be tricky?

> Can you remind me if there is an example of the new proposed API for this somewhere already?

The main API proposal was in choderalab/yank#535, but it's just a skeleton. At this point the examples in the `AlchemicalState` docstrings are probably the best place to figure out how things work.

andrrizzi:
And just to make sure there are not misunderstandings, the method `AbsoluteAlchemicalFactory.perturbSystem` that I've moved to `AlchemicalState.apply_to_system` just sets the parameters of the forces. The conversion to alchemical system would still happen in `AbsoluteAlchemicalFactory.createPerturbedSystems`.

andrrizzi:
I'm finalizing the changes to alchemy today. Two questions:
1. Do we want to have `annihilate_sterics`, `annihilate_electrostatics` and softcore parameters per `AlchemicalRegion` or per factory?
2. Do we want to keep the default complex/solvent implicit/explicit protocols? Are these still encouraged?

jchodera:
1. let's keep these by region 
2. Let's ditch these

andrrizzi:
Ok! With the softcore parameters too in the `AlchemicalRegion` though, `AbsoluteAlchemicalFactory` becomes a class with no attributes and a bunch of static methods. What about at this point renaming it `AlchemicalFactory` and expose different methods for single/dual topology (in the future)? The complete final API would become something like
```python
alchemical_region1 = AlchemicalRegion(alchemical_atoms=range(0, 22), alchemical_bonds=True,
                                      alchemical_torsions=[1, 2, 3])
alchemical_region2 = AlchemicalRegion(alchemical_atoms=range(52, 100), annihilate_sterics=True,
                                      softcore_beta=1.0)
alchemical_regions = {'region1': alchemical_region1, 'region2': alchemical_region2}
alchemical_system = AlchemicalFactory.create_alchemical_system(system, alchemical_regions)

# Infer all the regions' alchemical states from the system.
alchemical_state1, alchemical_state2 = AlchemicalState.from_system_regions(alchemical_system)

# Slave region2 sterics/electrostatics to lambda.
alchemical_state2.set_alchemical_variable('lambda', 1.0)
alchemical_state2.lambda_sterics = AlchemicalFunction('step_hm(lambda - 0.5) + 2*lambda * step_hm(0.5 - lambda)')
alchemical_state2.lambda_electrostatics = AlchemicalFunction('2*(lambda - 0.5) * step(lambda - 0.5)')

# Set region2 to non-interacting.
alchemical_state2.set_alchemical_variable('lambda', 0.0)
alchemical_state2.apply_to_system(alchemical_system)
```
and in the future, when we'll have a dual topology
```python
alchemical_system = AlchemicalFactory.dual_topology_alchemical_system(system1, system2,
                                              atom_mapping=atom_mapping, alchemical_regions=None)
```
where `alchemical_regions` are extra regions to alchemically modify (beside the relative transformation).

Sorry I'm proposing a lot of changes here, but this is effectively an alchemy API 2.0 and I'd like it to be stable for a long time.

andrrizzi:
Ah, no I forgot about the `consistent_exceptions` argument in the `__init__` that shouldn't probably go in the alchemical regions, so the `AlchemicalFactory` will have to be instantiated. I have edited the previous post.

andrrizzi:
Tests pass! This is ready to be merged.

Beside the things I've mentioned before, I've reworked the tests to separate slow from fast test cases and remove redundant code/tests.

andrrizzi:
This is blocking so I had to push my new changes on this PR. The last few commits add `__getstate__` and `__setstate__` methods to all the state classes, and a couple of serialization utilities in `utils.py`. This will simplify a lot the way our samplers store their attributes for resuming and generalize it to arbitrary state classes.

Still ready to be reviewed if tests pass.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+class AlchemicalStateError(states.ComposableStateError):
+    """Error raised by an AlchemicalState."""
+    pass
 
-#=============================================================================================
-# MODULE UTILITIES
-#=============================================================================================
 
-def _is_periodic(system):
+class AlchemicalFunction(object):
```
Lnaden: So I noticed these are not really used yet. Is the plan to make basically a master `lambda` variable that all other `lambda_X` variables map onto? e.g. you define the whole alchemical path by 3 functions defining restraints over the first third of the master, electrostatics over the second, and LJ over the last third?

openmmtools/alchemy.py
```diff
@@ -1415,397 +1701,114 @@ def _alchemicallyModifyGBSAOBCForce(self, system, reference_force, force_labels,
...
-        Returns
-        -------
-        system : simtk.openmm.System
-            Alchemically-modified version of self.reference_system
-        force_labels : dict of int : str
-            force_labels[name] is the force index in the alchemically modified system of the modified force `name`
+    @staticmethod
+    def _find_force_components(alchemical_system):
+        """Return force labels and indices for each force."""
+        def add_label(label, index):
```
Lnaden: Does this even work with `force_labels` targeting something in its outer scope that has not yet been defined? For at least clarity, could you move the variable initialization above the internal functions?

openmmtools/mcmc.py
```diff
@@ -530,16 +530,16 @@ def apply(self, thermodynamic_state, sampler_state):
         self._before_integration(context, thermodynamic_state)
 
         # Run dynamics.
-        timer.start("{}: step()".format(move_name))
+        timer.start("{}: step({})".format(move_name, self.n_steps))
```
Lnaden: This is fine for minimalist output, but would it be more clear to have: 

`"{}: Take {} steps".format(...)`

?

openmmtools/alchemy.py
andrrizzi: Yep! This is a feature that was already implemented in alchemy. Here I just moved them from the `CustomForce` to `AlchemicalState`.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
-    Non-periodic Lennard-Jones cluster.
+    # -------------------------------------------------------------------------
+    # Constructors
+    # -------------------------------------------------------------------------
 
-    >>> cluster = testsystems.LennardJonesCluster()
-    >>> print(_is_periodic(cluster.system))
-    False
+    def __init__(self, **kwargs):
+        self._alchemical_variables = {}
```
andrrizzi: Alchemical variables are those extra parameters that can be used inside an expression in `AlchemicalFunction`.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+                if alchemical_parameters[parameter_name] != parameter_value:
+                    err_msg = ('Parameter {} has been found in the force {} with two values: '
+                               '{} and {}').format(parameter_name, force.__class__.__name__,
+                                                   parameter_value, alchemical_parameters[parameter_name])
+                    raise AlchemicalStateError(err_msg)
+            else:
+                alchemical_parameters[parameter_name] = parameter_value
+
+        # Check that the system is alchemical.
+        if len(alchemical_parameters) == 0:
```
andrrizzi: Yep! If the system is not alchemical we can handle the exception.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+                parameter_name = force.getGlobalParameterName(parameter_id)
+                if parameter_name in supported_parameters:
+                    yield force, parameter_name, parameter_id
+
+
+# =============================================================================
+# ALCHEMICAL REGION
+# =============================================================================
+
+_ALCHEMICAL_REGION_ARGS = collections.OrderedDict(
```
andrrizzi: Sure!

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+
+    Parameters
+    ----------
+    alchemical_atoms : list of int, optional
+        List of atoms to be designated for which the nonbonded forces (both
+        sterics and electrostatics components) have to be alchemically
+        modified (default is None).
+    alchemical_bonds : bool or list of int, optional
+        If a list of bond indices are specified, these HarmonicBondForce
+        entries are softened with 'lambda_bonds'. If set to True, this list
```
andrrizzi: The argument handles both `None` and `False` (see line 915) since it can take both a list and a bool. I think setting this to `True` will be the most common way to use this, so I wouldn't raise warning. I can change the default to `False` if you think it's better.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+        Eq. 13 of Ref. [1] (default is 1).
+    softcore_d, softcore_e, softcore_f : float, optional
+        Parameters modifying softcore electrostatics form (default is 1).
+
+    Notes
+    -----
+    The parameters softcore_e and softcore_f determine the effective distance
+    between point charges according to
+
+    r_eff = sigma*((softcore_beta*(lambda_electrostatics-1)^softcore_e + (r/sigma)^softcore_f))^(1/softcore_f)
```
andrrizzi: Yep! Only they're for the sterics, here I just moved the existing documentation but it specifies this is the distance between charges. I can add the same formula for sterics.

jchodera:
Do we want to bump the version number to 2.0.0 here, since these are fairly major changes and work from here on out will break backwards compatibility?

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+        timer.stop('Create alchemically modified system')
+        timer.report_timing()
+
+        return alchemical_system
 
     @classmethod
-    def _getSystemGlobalParameters(cls, system, prefix='lambda_'):
-        """
-        Get a list of available alchemical parameters beginning with 'prefix'
+    def get_energy_components(cls, alchemical_system, alchemical_state, positions, platform=None):
```
andrrizzi: We use this only in tests currently. It may be useful to debug if we are hunting down problems, but this won't be used normally.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+        # Find and label all forces.
+        force_labels = cls._find_force_components(alchemical_system)
+        assert len(force_labels) <= 32, ("The alchemical system has more than 32 force groups; "
+                                         "can't compute individual force component energies.")
+
+        # Create deep copy of alchemical system.
+        system = copy.deepcopy(alchemical_system)
+
+        # Separate all forces into separate force groups.
+        for force_index, force in enumerate(system.getForces()):
```
andrrizzi: Yeah, that's a limitation, but since this is just code for testing, I'd implement only if needed.

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
+            force.setForceGroup(force_index)
+
+        # Create a Context in the given state.
         integrator = openmm.VerletIntegrator(1.0 * unit.femtoseconds)
-        context = openmm.Context(system, integrator, platform)
-        alchemical_parameters = [ parameter for parameter in context.getState(getParameters=True).getParameters().keys() if parameter.startswith(prefix) ]
+        if platform is None:
```
andrrizzi: The platform will be centralized in the `ContextCache` everywhere else. This is just a debug function.

openmmtools/alchemy.py
Lnaden: Actually, I can think of a reason why. If the alchemical region ONLY defined 2 atoms such as making an artificial or alchemical covalent bond. You could emulate a covalent binding event. Since we are moving to a multiple alchemical region framework, this feature may actually see some use.

For single molecule non-covalent free energy calculations though, horrible idea

openmmtools/alchemy.py
```diff
@@ -12,413 +12,905 @@
...
-        """
-        kwargs = { parameter : 1 for parameter in self.alchemical_parameters }
-        return AlchemicalState(**kwargs)
+    # -------------------------------------------------------------------------
+    # Internal usage: AlchemicalRegion
+    # -------------------------------------------------------------------------
 
     @classmethod
-    def _tabulateBonds(cls, system):
+    def _resolve_alchemical_region(cls, system, alchemical_region):
```
andrrizzi: Pretty much it is to ensure that all `alchemical_X` are `set`s and not `bool`s. If the passed arguments are `bool`s this construct the list of indices from the `alchemical_atoms` list.

openmmtools/alchemy.py
Lnaden: Good thought, keep it in

openmmtools/alchemy.py
Lnaden: Fai

openmmtools/alchemy.py
```diff
@@ -473,35 +971,42 @@ def _buildAlchemicalTorsionList(self, alchemical_atomset):
...
+        bonds = cls._tabulate_bonds(system)
+
+        def is_bonded(i, j):
             if j in bonds[i]:
                 return True
             return False
-        def is_proper_torsion(i,j,k,l):
-            if is_bonded(i,j) and is_bonded(j,k) and is_bonded(k,l):
+
+        def is_proper_torsion(i, j, k, l):
```
andrrizzi: @jchodera may be able to answer this question (it was his original implementation I believe).

Lnaden:
Probably, but bump it to a beta or at least a `2.0.0dev0` build since I think this is changing far too rapidly for us to say "2.0.0"

andrrizzi:
Technically we are not breaking backwards compatibility because this will be a whole other package. `openmmtools` is still on version `0.8.4` so we could bump it to `1.0`, or is there a specific module version? In both cases, before releasing I'd like to integrate these changes to Yank, because I've already found a couple of small things to add while making the changes to `repex`.

Lnaden:
Do we plan to release other changes before integrating this stuff into Yank? If so, then keep the lower revision, but once we start needing all the new API stuff in yank, we will want to bump the release version.

@andrrizzi lgtm beyond the small quality of life changes I requested and it looks like you have in the pipe. I'll switch to approved

jchodera:
Ah, OK! We can at least go to `0.9.x`. I'd like to get some real docs started before we start a `1.0.x` revision.

andrrizzi:
> Do we plan to release other changes before integrating this stuff into Yank?

Yeah, I already have a couple locally, minor stuff anyway. I've bumped to 0.9.0. I'll merge when tests pass.

PR #135, opened 2017-03-07
New storage system with auto-magical variable handling

Lnaden:
This is the new storage system module which handles hard drive IO for arbitrary data sets with minimal user input. Currently handles NetCDF storage, but can be extended to add new storage modules with from the abstract class.

Handles the following types:
* String
* Int
* Float
* List (any size)
* Tuple (any size)
* Quantity (of any of the previous types)
* Dict (entries can be any of the previous types with string keywords)
* numpy.ndarray (and Quantity(ndarray))

The `StorageInterface` module allows auto-magical data IO such that you define your directories and variables on-the-fly with commands such as 

`StorgeInterfaceInstance.my_directory.my_variable.write(data)` 

where a folder called `my_directory` and `my_variable` are created for you with `data` being automatically encoded to the storage medium and saved. Accessing it is done with 

`StorgeInterfaceInstance.my_directory.my_variable.read()`

and the data at that same location is fetched and automatically decoded for you. No template nor pre-determining structure required.

@andrrizzi Main review points are the changes to the `math_eval` function, the main `__init__`, and the new `quantity_from_string` function. Functionality of storage code unchanged from the YANK PR, unless there are other issues you see with it.

----

This is a migration of the code from choderalab/yank#617 to make this module stand separate from YANK as others may want to use it.

Tagging those that have expressed interest in this @bas-rustenburg @gregoryross @ChayaSt, I would appreciate any feedback you have on the code.


Lnaden:
One problem with this is that NetCDF4 on the main channel is only available on numpy 1.11 and not 1.12 like the one on `conda-forge`. So this package now sometimes causes numpy downgrades.

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1741 @@
...
+# =============================================================================
+# ABSTRACT TYPE HANDLER
+# =============================================================================
+
+
+class NCVariableTypeHandler(ABC):
+    """
+    Pointer class which provides instructions on how to handle a given nc_variable
+    """
+    def __init__(self, parent_handler, target, storage_object=None):
```
andrrizzi: I may be confused here, but could you rename `parent_handler` to `netcdf_driver`? And is there a use case in which we want to have a different `storage_object` than `netcdf_driver.ncfile`?

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1741 @@
...
+    else:
+        data = data.astype(int)
+    return data
+
+
+def nc_int_encoder(data):
+    return data
+
+
+def nc_float_decoder(nc_variable):
```
andrrizzi: One way you could remove code redundancy here is to use something like:
```python
# This works is for everything but strings.
def dummy_encoder(data):
    return data

# This works for ints and floats.
def scalar_encoder(casting_type):
    def _scalar_encoder(nc_variable):
        data = nc_variable[:]
        if data.shape == (1,):
            data = casting_type(data[0])
        else:
            data = data.astype(casting_type)
        return data
    return _scalar_encoder
```

openmmtools/utils.py
```diff
@@ -177,9 +180,12 @@ def math_eval(expression, variables=None):
                  ast.Pow: operator.pow, ast.USub: operator.neg}
 
     # Supported functions, not defined in math.
-    functions = {'step': lambda x: 1 * (x >= 0),
-                 'step_hm': lambda x: 0.5 * (np.sign(x) + 1),
-                 'sign': lambda x: np.sign(x)}
+    if functions is None:
+        functions = {}
+    functions.update({'step': lambda x: 1 * (x >= 0),
```
andrrizzi: Did you want to do the opposite?
```python
functions = {'step': ..., 'step_hm': ...}
if user_functions is not None:
    functions.update(user_functions)
```
or did you wanted the default functions to overwrite a custom `step` function?

openmmtools/utils.py
```diff
@@ -219,6 +225,15 @@ def _math_eval(node):
 # QUANTITY UTILITIES
 # =============================================================================
 
+# List of simtk.unit methods that are actually units and functions instead of base classes
+# Pre-computed to reduce run-time cost
+# Get the built-in units
+_VALID_UNITS = {method: getattr(unit, method) for method in dir(unit) if type(getattr(unit, method)) is unit.Unit}
+# Get the built in unit functions and make sure they are not just types
+_VALID_UNIT_FUNCTIONS = {method: getattr(unit, method) for method in dir(unit)
```
andrrizzi: If it can simplify your life, you could also make use of `inspect.getmembers(unit, predicate=inspect.isfunction)`.

openmmtools/storage/storageinterface.py
```diff
@@ -0,0 +1,521 @@
...
+
+    @property
+    def directory(self):
+        """
+        Checks if the object can be used as a directory for accepting other SIDV objects. Once this is set, the
+        .write, .append, and .fetch functions are locked out.
+
+        Returns
+        -------
+        directory_pointer : None, or storage_system specific directory of storage
```
Lnaden: I would agree with this except for the part where there is a 3rd state: `unbound`. When no IO has actually taken place yet, the SIDV does not know if it is a variable or a directory, it could be either. The resulting logic would require me to either return `None` or some informative string as a 3rd option and coding around that. By having both `.variable` and `.directory` return None or the target its bound to, it lets me manipulate this 3rd option. 

The `bound_target` object still acts like you want it and the `.variable` and `.directory` variables will probably not be used much in practice, but are there if people want them. I can still refactor it if you think its a good idea.

openmmtools/storage/storageinterface.py
```diff
@@ -0,0 +1,521 @@
...
+        Pointer to the object which actually handles read/write operations
+
+        Returns
+        -------
+        storage_driver :
+            Instance of the module which handles IO actions to specific storage type requested by storage_system
+            string at initialization.
+
+        """
+        return self._storage_driver
```
Lnaden: I had originally created that variable as a protected name so you could not assign it as one of the variables on disk. You are right, I can just derive this and all it did was expose a hidden objects' property. I did a small refactor to remove this and just derive the object.

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1741 @@
...
+                else:
+                    self.ncfile = nc.Dataset(self.file_name, 'w')
+            else:
+                self.ncfile = nc.Dataset(self.file_name, self.access_mode)
+
+    def __init__(self, file_name, access_mode=None):
+        super(NetCDFIODriver, self).__init__(file_name, access_mode=access_mode)
+        # Initialize the file bind variable. All actions involving files
+        self.ncfile = None
+        self._groups = {}
```
Lnaden: So this is actually something I am going to keep around to actually make interfacing with the NetCDF file easier. The `_groups` and `_variables` objects do not just house groups and variables at the top level, they serve as a way to flatten ALL the groups and ALL the variables in the netcdf file into a single dict. 

You are also going to have the problem any time someone directly acts on the netcdf file. the idea is that the `.ncfile` variable is there if you want it/need it, but in general everything can be handled through the IODriver. I also wrote it so that you can attempt to fetch variables on the disk on the fly, so even if someone makes changes, the system can find those changes and try to work with them, just like what we have to do now where we manually have to write both Create/Resume functionality.

Specifically because that exact problem above is a pain, I made the StorageInterface to remove that.

Lnaden: I will, however, make a pass and unify the naming convention.

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1741 @@
...
+        self.set_codec(int, NCInt)  # Int
+        self.set_codec(dict, NCDict)  # Dict
+        self.set_codec(float, NCFloat)  # Float
+        # List/tuple
+        self.set_codec(list, NCIterable)
+        self.set_codec(tuple, NCIterable)
+        self.set_codec(np.ndarray, NCArray)  # Array
+        self.set_codec(unit.Quantity, NCQuantity)  # Quantity
+        # Bind the metadata reader types based on the dtype string of each class
+        self._IOMetaDataReaders = {self._type_maps[key].dtype_string(): self._type_maps[key] for key in self._type_maps}
```
Lnaden: Now this one I can fix.

openmmtools/storage/iodrivers.py
Lnaden: Renamed them to `parent_driver` to remove confusion.

Yes! The storage object can either be the top level `ncfile` OR a group/subgroup within that file.

openmmtools/utils.py
Lnaden: Eh, this yields the same result and avoids another import. Its also not too hard to read, so I think I'll keep it.

openmmtools/storage/storageinterface.py
andrrizzi: I had imagined both `is_variable` and `is_directory` to return `False` if the variable wasn't bound. But a better check for the user to see if it bound at all would be to check that `bound_target` is `None`. Both solutions seems good to me anyway. The point was more about fixing the docstring.

Lnaden: That I can fix. We are stuck in limbo for a bit however since either `conda` or `conda-build` BROKE LOCAL PACKAGE AUTOMATION. And it also looks like `conda-build` broke our `conda-build-all` script on `conda-recipes`so we couldn't upload changes right now even if we wanted to.

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1653 @@
...
+    """PyYAML Loader that recognized !Quantity nodes, converts YAML output -> Python type"""
+    def __init__(self, *args, **kwargs):
+        super(NCYamlLoader, self).__init__(*args, **kwargs)
+        self.add_constructor(u'!Quantity', self.quantity_constructor)
+
+    @staticmethod
+    def quantity_constructor(loader, node):
+        loaded_mapping = loader.construct_mapping(node)
+        data_unit = quantity_from_string(loaded_mapping['NCUnit'])
+        data_value = loaded_mapping['NCValue']
```
andrrizzi: Maybe we should call these two just `unit` and `value`? This way we'll be able to use the same serialization in other formats without tying it to netcdf.

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1653 @@
...
+            return_type = self._on_disk_dtype
+        return return_type
+
+    @property
+    def _on_disk_dtype(self):
+        """
+        Allow overwriting the dtype for storage for extending this method to cast data as a different type on disk
+        This is the property to overwrite the cast dtype
+        """
+        return None
```
andrrizzi: If you just return `self.dtype` instead of None, you can remove the hybrid property/getter `_get_on_disk_dtype` and you would still be able to overwrite the behavior.

openmmtools/tests/test_storage_iodrivers.py
```diff
@@ -0,0 +1,209 @@
...
+
+def test_netcdf_dictionary_type_codec():
+    """Test that the dictionary type codec can read/write/append with various unit and _value types"""
+    input_data = {
+        'count': 4,
+        'ratio': 0.4,
+        'name': 'four',
+        'repeated': [4, 4, 4],
+        'temperature': 4 * unit.kelvin,
+        'box_vectors': (np.eye(3) * 4.0) * unit.nanometer
```
andrrizzi: Can you add a `Quantity`-wrapped numpy array as a test case here to make sure it works if it's not somewhere else?

openmmtools/storage/iodrivers.py
Lnaden: Can't do that. The `self.dtype` property is how the IODriver identifies what codec to use based purely on the input data. It processes `type(data)` and reads the codec from the internal database. The problem with the `dict` type is that its a subset of `NCScalar` which used to pass `self.dtype` to the NetCDF variable creation routine, but `dict` is not a type NetCDF can handle, so I had to come up with a way to define a mappable type in the codec, without breaking the IODriver's auto-data-detection methods.

openmmtools/storage/iodrivers.py
Lnaden: Are you suggesting generalizing the dict YAML Loaders/Dumper for future drivers?

openmmtools/tests/test_storage_iodrivers.py
Lnaden: Line 207, already does it!

openmmtools/storage/iodrivers.py
andrrizzi: Sorry, I think I wasn't clear. I'm just suggesting to substitute this:
```python
    @property
    def _get_on_disk_dtype(self):
        """Function to process None for _on_disk_dtype"""
        if self._on_disk_dtype is None:
            return_type = self.dtype
        else:
            return_type = self._on_disk_dtype
        return return_type

    @property
    def _on_disk_dtype(self):
        """
        Allow overwriting the dtype for storage for extending this method to cast data as a different type on disk
        This is the property to overwrite the cast dtype
        """
        return None
```
with a single
```python
    @property
    def _on_disk_dtype(self):
        """
        Allow overwriting the dtype for storage for extending this method to cast data as a different type on disk
        This is the property to overwrite the cast dtype
        """
        return self.dtype
```
i.e. get rid of the hybrid `_get_on_disk_dtype` property. Unless there's something I'm missing.

Lnaden: I see now, yes, that I can do

openmmtools/storage/iodrivers.py
andrrizzi: No no this is fine for now, but if in the future we want to do this somewhere else, we'll need a dict to str serialization that has not `NC` in the name and we'll have to duplicate the code.

openmmtools/tests/test_storage_iodrivers.py
andrrizzi: Doh! Didn't realize, sorry.

Lnaden:
@andrrizzi I have brought in the changes we discussed yesterday which should be reviewed. You can look at the delta between the last review (4e4770a) and now [at this link](https://github.com/choderalab/openmmtools/compare/4e4770a...b1fd593)

openmmtools/storage/iodrivers.py
```diff
@@ -0,0 +1,1612 @@
...
+            self.add_metadata('IODriver_Type', self.dtype_string())
+            self.add_metadata('type', typename(self.dtype))
+            self._unit = 'NoneType'
+            self.add_metadata('IODriver_Unit', self._unit)
+            # Specify the type of storage object this should tie to
+            self.add_metadata('IODriver_Storage_Type', self.storage_type)
+            self.add_metadata('IODriver_Appendable', 0)
+        self._dump_metadata_buffer()
+        # Set the output mode by calling the variable
+        self._output_mode
```
andrrizzi: Here too (and in few other places).

Lnaden:
I've refactored this a bit more to make `_bind_read` a non-abstract method, also moved parts of `_bind_append` and `_bind_write` functions to a common call.

andrrizzi:
Looks great! I just realized I forgot to import the changes to the dictionary codecs from yank. I'll do it asap.

jchodera:
Feel free to review and merge this without my involvement!

Lnaden:
There are some bugs I have been working out and @andrrizzi got the review in, once I finish up the last few conflicts (and travis stops fighting me), I'll be able to merge this

Lnaden:
So the windows builds now fail because `tqdm` and `openmm` conflict. There is no Windows Python 3.4 cut for `tqdm`, and I don't think there was one for `OpenMM` either, the windows tests have been using OpenMM 6.3 from what it looks like. This also strikes me as odd.

jchodera:
This will be fixed as soon as we add [conda-forge](https://anaconda.org/conda-forge/tqdm/files) as a dependency.

Lnaden:
I'm assuming that will also fix the Linux Python 3.4 build not finishing either, since I assume its `tqdm` missing again (not on the `anaconda` channel). A few points though:

1. The Python 3.4 tests are garbage since they dont build and instead pull packages from the net
2. Do we really need that progress bar on [one test](https://github.com/choderalab/openmmtools/blob/ac11feadfe4428da970bf4fe6cdab45aa580ff55/openmmtools/tests/test_integrators.py#L309) thus requiring a whole new package?
3. We should drop Python 3.4 support anyways once we get a stable OpenMM on all platforms for python 3.6

jchodera:
> Do we really need that progress bar on one test thus requiring a whole new package?

We can ditch that.

> We should drop Python 3.4 support anyways once we get a stable OpenMM on all platforms for python 3.6

I can't stress this enough: We MUST do whatever conda-forge does regarding package support for now. Let's not drop support for anything until conda-forge does.

Lnaden:
Okay, this is back into a state it can be merged in. I don't have it handle the compressed, fixed length dictionaries like we do in Yank's dicts when writing to NetCDF, but I don't think we quite need that yet. Did you want to take a look at this to make sure it does all you need to, @jchodera?

jchodera:
Awesome! Thanks for adding docs!

Let's go ahead and merge this in and we can improve it from here.

Lnaden:
Will do if this last test does not time out again!

PR #140, opened 2017-03-17
Several small improvements to existing framework

andrrizzi:
Ready for review. Several bug fixes and improvements. Here is a list of the main changes:
- `ContextCache` can now be used to retrieve a context in a thermodynamic state with any integrator, which is useful when we only have to compute energies.
- Added an `IntegratorMove` class to easily instantiate an `MCMCMove` object from an OpenMM `Integrator`.
- Implemented `__getstate__` and `__setstate__` methods for all states and MCMC moves for future-compatible serialization.
- Split the restorable interface features for `CustomIntegrator`s from `ThermostatedIntegrator` to a new base class `RestorableIntegrator`
- Any MCMCMove inheriting from `BaseIntegratorMove` now checks for NaNs energy after `apply()`, and optionally restart the move for a configurable number of times. An `IntegratorMoveError` that can serialize the `Context` is raised if it cannot recover.
- Added new MCMCMoves `MetropolizedMove` (partially abstract), `MCRotationMove` and `MCDisplacementMove` which were previously implemented as part of `ModifiedHamiltonianExchange` in `yank/sampling.py`.

### DIFF
--- a/openmmtools/tests/test_cache.py
+++ b/openmmtools/tests/test_cache.py
@@ -221,6 +221,25 @@ class TestContextCache(object):
         assert len(cache) == 4
         assert n_contexts == 4
 
+    def test_get_context_any_integrator(self):
+        """ContextCache.get_context first search the cache when integrator is unspecified."""
+        cache = ContextCache()
+        state1, state2 = self.incompatible_states[:2]
+
+        # First we create a Context in state1.
+        cache.get_context(state1, copy.deepcopy(self.verlet_2fs))
+        assert len(cache) == 1
+
+        # When we don't specify the integrator, it first looks for cached Contexts.
+        context, integrator = cache.get_context(state1)
+        assert len(cache) == 1
+        assert state1.is_context_compatible(context)
+        assert isinstance(integrator, openmm.VerletIntegrator)
+
+        # With an incompatible state, a new Context is created.
+        cache.get_context(state2)
+        assert len(cache) == 2
+
     def test_cache_capacity_ttl(self):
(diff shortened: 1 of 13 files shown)


openmmtools/tests/test_cache.py
```diff
@@ -221,6 +221,25 @@ def test_get_incompatible_context(self):
...
         assert n_contexts == 4
 
+    def test_get_context_any_integrator(self):
+        """ContextCache.get_context first search the cache when integrator is unspecified."""
+        cache = ContextCache()
+        state1, state2 = self.incompatible_states[:2]
+
+        # First we create a Context in state1.
+        cache.get_context(state1, copy.deepcopy(self.verlet_2fs))
+        assert len(cache) == 1
```
pgrinaway: Why do we assert that the length of the cache is 1?

pgrinaway: I mean, what if it were 20? Would that be bad?

andrrizzi: Oh, sorry, that's just a particular test case. We instantiate a new local, empty `ContextCache` above (I'm not using the global cache in this test), so that `assert` is just checking that the cache (originally empty) has indeed created and stored a `Context` object after the call to `get_context`.

If we were using the global cache in `cache.global_context_cache` that could be 20 depending on what has happened before.

pgrinaway: Oh, wow, I totally misread that. Who will review the reviewers?

andrrizzi:
Thanks for going through it! Merging then.

PR #142, opened 2017-03-22
Custom ghmc

maxentile:
Metropolized version of `LangevinSplittingIntegrator`. Metropolizes each step using `shadow_work`. Allows user to:
* specify order of deterministic integration steps using a string such as `V R V`, 
* specify order of all stochastic and deterministic steps using a string such as `O V R V O`,
* specify a MTS GHMC scheme using a string such as `V0 R V1 R V1 R V1 R V0`

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1262,7 +1262,204 @@ class GeodesicBAOABIntegrator(LangevinSplittingIntegrator):
                                              measure_shadow_work=measure_shadow_work,
                                              measure_heat=measure_heat
                                              )
+class CustomizableGHMC(mm.CustomIntegrator):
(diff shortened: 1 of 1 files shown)


jchodera:
This is great for now, but at some point in the future I bet we can refactor `LangevinSplittingIntegrator` and this version to minimize code duplication and inherit from the same integrator (or call the same base integrator with or without a Metropolization flag)!

maxentile:
Agreed! And there's probably a better way to define integrator "building blocks" (like the V, R, O steps) that can be used across multiple integrators, instead of defining them in each class...

maxentile:
Obviated by https://github.com/choderalab/openmmtools/pull/146 !

PR #143, opened 2017-03-22
Fix travis segfault

### LINKED ISSUE #138: Stochastic Seg Fault on ThermoState.create_context For Python 3+
@andrrizzi 

There [appear to be tests](https://travis-ci.org/choderalab/openmmtools/jobs/211134982#L2180) which stochastically cause segmentation faults on Python 3. The [test itself is here](https://github.com/choderalab/openmmtools/blob/98d8b0614ee1b8208a38924894969cca923acc24/openmmtools/tests/test_states.py#L543). If you repeat the tests enough times, it will successfully build. As the one who originally wrote this code, could you take a look and see if there is some obvious reason we are seeing a segmentation fault?

andrrizzi:
Fix #138 .

This seems to have worked. In the test, I was creating a bunch of `Context`s on different platforms and probably the garbage collector was not quick enough to delete them without an explicit `del` statement.

I'll rerun the tests 2-3 times before merging just to be sure.

Thanks to @Lnaden for directing my attention to the faulty test.

### DIFF
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -20,7 +20,7 @@ import copy
 import numpy as np
 from simtk import openmm, unit
 
-from openmmtools import utils, integrators
+from openmmtools import utils, integrators, constants
 
 
 # =============================================================================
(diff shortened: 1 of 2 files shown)


Lnaden:
You will want to delete the `integrator`s as well, the C++ garbage collection does not happen until all references to the `context` are gone, which includes the integrator

andrrizzi:
Hmm. I'm actually getting another error now in a different test, but I'll try that. Thanks for the tip!

andrrizzi:
Seems to have fixed it! It was probably the leftovers of the `create_context` test that was making the other one fail. Thanks!

PR #144, opened 2017-03-23
Track external protocol work in LangevinSplittingIntegrator

bas-rustenburg:
Why:

Some codebases, for instance protons and saltswap, modify context
parameters externally. To save expensive energy computations,
we want to accumulate the protocol work inside the integrator.
This commit adds a `protocol_work` variable to keep track of it.

It also adds a unit test to make sure the new feature works adequately.

This change addresses the need by:

* Adding the `measure_protocol_work` parameter, default false, and
including the machinery that calculates it.
* Adding a unit test to ensure it works as expected.

Please double-check that the tests passes before merging. I haven't been able to run it locally yet.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -964,7 +964,6 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
         measure_protocol_work : boolean
             Accumulate the protocol work, in the global `protocol_work`.
             Assumes that context parameters have been perturbed externally.
-            Note that it adds one energy computation per simulation step.
         """
 
         # Compute constants
@@ -1104,19 +1103,21 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
         if measure_protocol_work:
             self.addGlobalVariable("protocol_work", 0)
             self.addGlobalVariable("perturbed_pe", 0)
+            self.addGlobalVariable("unperturbed_pe", 0)
 
         # Integrate
         self.addUpdateContextState()
         self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
 
         # Protocol work is calculated by taking the potential energy in the perturbed system
-        # and subtracting the previous, unperturbed potential energy
-        # before any steps are taken.
+        # and subtracting the previous, unperturbed potential energy after the last iteration.
         if measure_protocol_work:
             self.addComputeGlobal("perturbed_pe", "energy")
-            self.addComputeGlobal("protocol_work", "protocol_work + (perturbed_pe - new_pe)")
+            self.addComputeGlobal("protocol_work", "protocol_work + (perturbed_pe - unperturbed_pe)")
         for i, step in enumerate(splitting):
             substep_function(step)
+        if measure_protocol_work:
+            self.addComputeGlobal("unperturbed_pe", "energy")
 
 
 class VVVRIntegrator(LangevinSplittingIntegrator):
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1128,7 +1128,8 @@ class VVVRIntegrator(LangevinSplittingIntegrator):
                  timestep=1.0 * simtk.unit.femtoseconds,
                  constraint_tolerance=1e-8,
                  measure_shadow_work=False,
-                 measure_heat=True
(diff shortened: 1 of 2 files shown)


pgrinaway:
🔥  nice

bas-rustenburg:
It would be appreciated if @maxentile and @gregoryross take a look at this to make sure the implementation is correct, since they should be most familiar with the integrator and procedure respectively.

bas-rustenburg:
Thanks Josh, I'll fix that. It uses the energy at the end of the _previous_ step, and at the start of the upcoming step.

openmmtools/tests/test_integrators.py
```diff
@@ -206,6 +206,39 @@ def test_vvvr_shadow_work_accumulation():
...
+    context = openmm.Context(system, integrator)
+    context.setPositions(testsystem.positions)
+    context.setVelocitiesToTemperature(temperature)
+    assert(integrator.getGlobalVariableByName('protocol_work') == 0), "Protocol work should be 0 initially"
+    integrator.step(25)
+    assert(integrator.getGlobalVariableByName('protocol_work') == 0), "There should be no protocol work."
+    perturbed_K=99.0 * unit.kilocalories_per_mole / unit.angstroms**2
+    context.setParameter('testsystems_HarmonicOscillator_K', perturbed_K)
+    integrator.step(1)
+    assert (integrator.getGlobalVariableByName('protocol_work') != 0), "There should be protocol work after perturbing."
```
gregoryross: There are multiple `assert` statements in this functions that each test a different aspect. It's certainly worth considering splitting these into separate tests.

bas-rustenburg: Perhaps not optimal. I was just following the original test structure that Josh used. It should at least fail with a specific error message if it does fail.

openmmtools/tests/test_integrators.py
```diff
@@ -206,6 +206,39 @@ def test_vvvr_shadow_work_accumulation():
     assert('shadow_work' not in names_of_globals)
 
 
+def test_vvvr_protocol_work_accumulation():
```
gregoryross: I strongly recommend testing the accumulated work explicitly by comparing with `getState(getEnergy=True)`. I *thought* I had a test for this in `saltswap` here:

https://github.com/choderalab/saltswap/blob/master/saltswap/tests/test_integrators.py#L185-L206

But it's not exactly the same as testing the internally accumulated work. Anyway, something that tests for that would be good. (I'll certainly be adding to my `saltswap` tests.)

bas-rustenburg: I'll add that, thanks for the suggestion!

bas-rustenburg:
Just put in a new commit that hopefully addresses @maxentile's comments.

openmmtools/integrators.py
```diff
@@ -1093,14 +1098,31 @@ def substep_function(step_string):
...
         # Integrate
         self.addUpdateContextState()
         self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
+
+        # Protocol work is calculated by taking the potential energy in the perturbed system
+        # and subtracting the previous, unperturbed potential energy
+        # before any steps are taken.
+        if measure_protocol_work:
+            self.addComputeGlobal("perturbed_pe", "energy")
+            self.addComputeGlobal("protocol_work", "protocol_work + (perturbed_pe - new_pe)")
```
gregoryross: I find it difficult to wrap my head around `CustomIntegrators` at the best of times, but is there potential for conflict if `measure_shadow_work = True`, as `new_pe` is set to zero?

bas-rustenburg: You might be right there. Latest fix takes care of it.

bas-rustenburg:
Fixed Greg's concern and no longer using new_pe, thanks for suggesting it Josh!

maxentile:
Looks great to me!

bas-rustenburg:
Update to test against external potential energy included now.

bas-rustenburg:
Bugfix

bas-rustenburg:
This should hopefully be the last bug. It passes for me locally now.

bas-rustenburg:
I hereby declare this PR, merged!

PR #145, opened 2017-03-24
Remove lambda_restraints from parameters controlled by AlchemicalState

andrrizzi:
I have removed `lambda_restraints` from parameters controlled by `AlchemicalState` since the `alchemy` module does not offer features related to restraints. The parameter is now controlled in YANK by a dedicated composable `RestraintState`.

I've also added some context cleanup in the tests that hopefully will fix the last memory-related problem with Travis tests.

@bas-rustenburg this is the last API-breaking change I plan to make before YANK 1.0, so `openmmtools` should be ready for release on my side.


jchodera:
Thanks! This seems fine.

PR #146, opened 2017-04-03
[WIP] Integrator extensions

pgrinaway:
This is a refactor of the `LangevinSplittingIntegrator`, as well as a subclass that can do NCMC. I think there are a few things left, but I think I am forgetting some:

- [ ] Rename AlchemicalLangevinSplittingIntegrator to something (suggestions?)
- [ ] Simplify shadow work computation for efficiency
- [ ] Add more sanity checks to prevent unworkable combinations of parameters

Anything else? Otherwise, please review and let me know if something is not clear or needs to change.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1058,13 +1058,17 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
         measure_protocol_work : boolean
             Accumulate the protocol work, in the global `protocol_work`.
             Assumes that context parameters have been perturbed externally.
-        metropolize : bool, optional
-            Whether to metropolize each step, based on the shadow work. If true, resets shadow work.
         """
 
         # Compute constants
         kT = kB * temperature
         gamma = collision_rate
+
+        #check if integrator is metropolized by checking for M step:
+        if splitting.find("M") > -1:
+            self._metropolized_integrator = True
+        else:
+            self._metropolized_integrator = False
 
         ORV_counts, mts, force_group_nV = self.parse_splitting_string(splitting)
 
@@ -1106,17 +1110,13 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
             self.addGlobalVariable("unperturbed_pe", 0)
 
         #If we metropolize, we have to keep track of the before and after (x, v)
-        if metropolize:
+        if self._metropolized_integrator:
             self.addPerDofVariable("vold", 0)
             self.addPerDofVariable("xold", 0)
 
         # Integrate
         self.addUpdateContextState()
         self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
-
-        if metropolize:
-            self.addComputePerDof("xold", "x")
-            self.addComputePerDof("vold", "v")
 
         # Protocol work is calculated by taking the potential energy in the perturbed system
         # and subtracting the previous, unperturbed potential energy after the last iteration.
@@ -1131,8 +1131,6 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
         if measure_protocol_work:
             self.addComputeGlobal("unperturbed_pe", "energy")
 
-        if metropolize:
-            self.metropolize()
 
     def sanity_check(self, splitting, allowed_characters="M()RVO0123456789"):
         """
@@ -1153,7 +1151,7 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
         splitting_no_space = splitting.replace(" ", "")
         # Make sure we contain at least one of R, V, O steps
         assert ("R" in splitting_no_space)
-        assert ("V" in [s[0] for s in splitting_no_space])
+        assert ("V" in splitting_no_space)
         assert ("O" in splitting_no_space)
 
         # Make sure it contains no invalid steps
@@ -1366,11 +1364,51 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
 class AlchemicalLangevinSplittingIntegrator(LangevinSplittingIntegrator):
     """
     This class performs nonequilibrium switching using the Perses-style lambda functions.
+
+    The type of propagator can be specified via the splitting string argument, described below.
+
+        Parameters
+        ----------
+        alchemical_functions : dict of strings
+            key: value pairs such as "global_parameter" : function_of_lambda where function_of_lambda is a Lepton-compatible
+            string that depends on the variable "lambda"
+
+        splitting : string
+            Sequence of R, V, O (and optionally V{i}), and M( ) substeps to be executed each timestep.
+
+            Forces are only used in V-step. Handle multiple force groups by appending the force group index
+            to V-steps, e.g. "V0" will only use forces from force group 0. "V" will perform a step using all forces.
+            M( will cause metropolization, and must be followed later by a ).
+
+
+        temperature : numpy.unit.Quantity compatible with kelvin, default: 298.0*simtk.unit.kelvin
+           Fictitious "bath" temperature
+
+        collision_rate : numpy.unit.Quantity compatible with 1/picoseconds, default: 91.0/simtk.unit.picoseconds
+           Collision rate
+
+        timestep : numpy.unit.Quantity compatible with femtoseconds, default: 1.0*simtk.unit.femtoseconds
+           Integration timestep
+
+        constraint_tolerance : float
+            Tolerance for constraint solver
+
+        override_splitting_checks : boolean
+            If True, skip sanity-checks on `splitting` string.
+
+        measure_shadow_work : boolean
+            Accumulate the shadow work performed by the symplectic substeps, in the global `shadow_work`
+
+        measure_heat : boolean
+            Accumulate the heat exchanged with the bath in each step, in the global `heat`
+
+        measure_protocol_work : boolean
+            Accumulate the protocol work, in the global `protocol_work`.
+            Assumes that context parameters have been perturbed externally.
     """
 
     def __init__(self,
                  alchemical_functions,
-                 system,
                  splitting="V R O R V",
                  temperature=298.0 * simtk.unit.kelvin,
                  collision_rate=91.0 / simtk.unit.picoseconds,
@@ -1379,8 +1417,7 @@ class AlchemicalLangevinSplittingIntegrator(LangevinSplittingIntegrator):
                  override_splitting_checks=False,
                  measure_shadow_work=False,
                  measure_heat=True,
-                 measure_protocol_work=False,
-                 metropolize=False,
+                 measure_protocol_work=True,
                  direction="forward",
                  nsteps_neq=100):
 
@@ -1388,13 +1425,21 @@ class AlchemicalLangevinSplittingIntegrator(LangevinSplittingIntegrator):
         self._direction = direction
         self._n_steps_neq = nsteps_neq
 
-        self._system_parameters = set()
+        #add some global variables relevant to the integrator
+        self.addGlobalVariable('lambda', 0.0) # parameter switched from 0 <--> 1 during course of integrating internal 'nsteps' of dynamics
+        self.addGlobalVariable('kinetic', 0.0) # kinetic energy
+        self.addGlobalVariable('nsteps', self._n_steps_neq) # total number of NCMC steps to perform
+        self.addGlobalVariable('step', 0) # current NCMC step number
 
-        for force_index in range(system.getNumForces()):
-            force = system.getForce(force_index)
-            if hasattr(force, 'getNumGlobalParameters'):
-                for parameter_index in range(force.getNumGlobalParameters()):
-                    self.system_parameters.add(force.getGlobalParameterName(parameter_index))
+        #collect the system parameters.
+        self._system_parameters = {system_parameter for system_parameter in alchemical_functions.keys()}
+
+        #call the base class constructor
+        super(AlchemicalLangevinSplittingIntegrator, self).__init__(splitting=splitting, temperature=temperature,
+                                                                    collision_rate=collision_rate, timestep=timestep,
+                                                                    constraint_tolerance=constraint_tolerance, override_splitting_checks=override_splitting_checks,
+                                                                    measure_shadow_work=measure_shadow_work, measure_heat=measure_heat,
+                                                                    measure_protocol_work=measure_protocol_work)
 
     def addUpdateAlchemicalParametersStep(self):
         """
@@ -1423,6 +1468,64 @@ class AlchemicalLangevinSplittingIntegrator(LangevinSplittingIntegrator):
         # Accumulate protocol work
         self.addComputeGlobal("Enew", "energy")
         self.addComputeGlobal("protocol_work", "protocol_work + (Enew-Eold)/kT")
+
+    def sanity_check(self, splitting, allowed_characters="HM()RVO0123456789"):
+        super(AlchemicalLangevinSplittingIntegrator, self).sanity_check(splitting, allowed_characters=allowed_characters)
+
+    def substep_function(self, step_string, measure_shadow_work, measure_heat, n_R, force_group_nV, mts):
+        """
+        Take step string, and add the appropriate R, V, O, M, or H step with appropriate parameters.
+        The step string input here is a single character (or character + number, for MTS)
+
+        Parameters
+        ----------
+        step_string : str
+            R, O, V, or Vn (where n is a nonnegative integer specifying force group)
+        measure_shadow_work : bool
+            Whether the steps should measure shadow work
+        measure_heat : bool
+            Whether the O step should measure heat
+        n_R : int
+            The number of R steps per integrator step
+        n_V : int
+            The number of V steps per integrator step
+        force_group_nV : dict
+            The number of V steps per integrator step per force group. {0: nV} if not mts
+        mts : bool
+            Whether the integrator is a multiple timestep integrator
+        """
+
+        if step_string == "O":
+            self.O_step(measure_heat)
+        elif step_string == "R":
+            self.R_step(measure_shadow_work, n_R)
+        elif step_string == "M(":
+            self.addComputePerDof("xold", "x")
+            self.addComputePerDof("vold", "v")
+        elif step_string == ")":
+            self.metropolize()
+        elif step_string[0] == "V":
+            #get the force group for this update--it's the number after the V
+            force_group = step_string[1:]
+            self.V_step(force_group, measure_shadow_work, force_group_nV, mts)
+        elif step_string == "H":
+            self.addAlchemicalPerturbationStep()
+
+    def addGlobalVariables(self, nsteps):
+        """
+        Add the appropriate global parameters to the CustomIntegrator. nsteps refers to the number of
+        total steps in the protocol.
+
+        Parameters
+        ----------
+        nsteps : int, greater than 0
+            The number of steps in the switching protocol.
+        """
+        self.addGlobalVariable('lambda', 0.0) # parameter switched from 0 <--> 1 during course of integrating internal 'nsteps' of dynamics
+        self.addGlobalVariable('kinetic', 0.0) # kinetic energy
+        self.addGlobalVariable('nsteps', nsteps) # total number of NCMC steps to perform
+        self.addGlobalVariable('step', 0) # current NCMC step number
+
 
 
 class VVVRIntegrator(LangevinSplittingIntegrator):
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -999,6 +999,11 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
         - g-BAOAB with solvent-solute splitting, K_r=K_p=2:
             splitting="V0 V1 R R O R R V1 R R O R R V1 V0"
 
+    Attributes
+    ----------
+    _kinetic_energy : str
+        This is 0.5*m*v*v by default, and is the expression used for the kinetic energy
+
     References
     ----------
     [Leimkuhler and Matthews, 2015] Molecular dynamics: with deterministic and stochastic numerical methods, Chapter 7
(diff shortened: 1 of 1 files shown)


openmmtools/integrators.py
```diff
@@ -1013,17 +1020,19 @@ def __init__(self,
                  override_splitting_checks=False,
```
bas-rustenburg: While we're at it, we should change the default collision rate to 1 ps.

openmmtools/integrators.py
```diff
@@ -1054,104 +1063,14 @@ def __init__(self,
...
 
-        # Do a couple sanity checks on the splitting string
-        if override_splitting_checks == False:
-            # Make sure we contain at least one of R, V, O steps
-            assert ("R" in splitting)
-            assert ("V" in [s[0] for s in splitting])
-            assert ("O" in splitting)
-
-            # Make sure it contains no invalid steps
-            assert (set(splitting).issubset(set("RVO").union(set(["V{}".format(i) for i in range(32)]))))
```
maxentile: I think the test in L1086 got dropped in the new sanity_check function

openmmtools/integrators.py
pgrinaway: done

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+        self.addComputePerDof("xold", "x")
+        self.addComputePerDof("vold", "v")
+
+class AlchemicalLangevinSplittingIntegrator(LangevinSplittingIntegrator):
+    """
+    This class performs nonequilibrium switching using the Perses-style lambda functions.
+
+    The type of propagator can be specified via the splitting string argument, described below.
+
+        Parameters
```
maxentile: Should the argument descriptions be in the docstring of the constructor?

Also, list of arguments in docstring is missing `direction`, `nsteps_neq`

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+            Whether the integrator is a multiple timestep integrator
+        """
+
+        if step_string == "O":
+            self.O_step(measure_heat)
+        elif step_string == "R":
+            self.R_step(measure_shadow_work, n_R)
+        elif step_string == "M(":
+            self.addComputePerDof("xold", "x")
+            self.addComputePerDof("vold", "v")
```
maxentile: Should `shadow_work` also be reset to zero here?

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+        ----------
+        step_string : str
+            R, O, V, or Vn (where n is a nonnegative integer specifying force group)
+        measure_shadow_work : bool
+            Whether the steps should measure shadow work
+        measure_heat : bool
+            Whether the O step should measure heat
+        n_R : int
+            The number of R steps per integrator step
+        n_V : int
```
maxentile: `n_V` listed in docstring but isn't an argument

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+
+    def substep_function(self, step_string, measure_shadow_work, measure_heat, n_R, force_group_nV, mts):
+        """
+        Take step string, and add the appropriate R, V, O, M, or H step with appropriate parameters.
+        The step string input here is a single character (or character + number, for MTS)
+
+        Parameters
+        ----------
+        step_string : str
+            R, O, V, or Vn (where n is a nonnegative integer specifying force group)
```
maxentile: Missing `M`

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+        measure_shadow_work : bool
+            Whether the steps should measure shadow work
+        measure_heat : bool
+            Whether the O step should measure heat
+        n_R : int
+            The number of R steps per integrator step
+        n_V : int
```
maxentile: `n_V` listed in docstring but not an argument

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+
+    def substep_function(self, step_string, measure_shadow_work, measure_heat, n_R, force_group_nV, mts):
+        """
+        Take step string, and add the appropriate R, V, O, M, or H step with appropriate parameters.
+        The step string input here is a single character (or character + number, for MTS)
+
+        Parameters
+        ----------
+        step_string : str
+            R, O, V, or Vn (where n is a nonnegative integer specifying force group)
```
maxentile: Missing `M`

openmmtools/integrators.py
```diff
@@ -1199,12 +1123,411 @@ def substep_function(step_string):
...
+        mts : bool
+            Whether the integrator is a multiple timestep integrator
+        """
+
+        if step_string == "O":
+            self.O_step(measure_heat)
+        elif step_string == "R":
+            self.R_step(measure_shadow_work, n_R)
+        elif step_string == "M(":
+            self.addComputePerDof("xold", "x")
```
maxentile: Should `shadow_work` be reset to zero here?

maxentile:
Oops, sorry for the duplicated comments at L1483, L1490, L1502-- had refreshed the browser during the Review and thought they had disappeared. Deleting one each of the duplicates now...

pgrinaway:
Just reorganized a tiny bit to allow the subclass which would work with `constph` et al. to be simpler. Taking a look at the other comments now.

openmmtools/integrators.py
pgrinaway: Hm, sorry--I interpreted "no invalid steps" to mean that it contained something other than RVO and the digits 0-9. Is that not correct?

maxentile: OpenMM limits the number of force groups, so the allowable steps should be {`R`, `O`, `V0`, `V1`, ..., `V31`}, I think (plus metropolization and hamiltonian updates now)

pgrinaway: Ok, will fix

openmmtools/integrators.py
pgrinaway: I think so.

openmmtools/integrators.py
```diff
@@ -1185,25 +1099,464 @@ def substep_function(step_string):
...
+    def metropolize(self):
+        """
+        Add a Metropolization (based on shadow work) step to the integrator. When Metropolization occurs, shadow work
+        is reset.
+        """
+        self.addComputeGlobal("accept", "step(exp(-(shadow_work)/kT) - uniform)")
+        self.beginIfBlock("accept != 1")
+        self.addComputePerDof("x", "xold")
+        self.addComputePerDof("v", "-vold")
+        self.addComputeGlobal("shadow_work", 0)
```
maxentile: Should this be moved out of the if block?

maxentile: Also, should `metropolize()` increment `naccept`, `ntrials`, as GHMCIntegrator did?

openmmtools/integrators.py
pgrinaway: Done

openmmtools/integrators.py
pgrinaway: Done

openmmtools/integrators.py
pgrinaway: Done

openmmtools/integrators.py
pgrinaway: Actually, shouldn't it be reset on the accept/reject step?

openmmtools/integrators.py
pgrinaway: Because of how the Metropolis step is structured, I'm going to have an `nreject` instead.

pgrinaway: done

openmmtools/integrators.py
pgrinaway: done

openmmtools/integrators.py
```diff
@@ -1185,25 +1094,536 @@ def substep_function(step_string):
...
+                                                                              measure_heat=measure_heat)
+
+    def add_integrator_steps(self, splitting, measure_shadow_work, measure_heat, ORV_counts, force_group_nV, mts):
+        self.addComputeGlobal("perturbed_pe", "energy")
+        self.addComputeGlobal("protocol_work", "protocol_work + (perturbed_pe - unperturbed_pe)")
+        super(ExternalPerturbationLangevinSplittingIntegrator, self).add_integrator_steps(splitting, measure_shadow_work, measure_heat, ORV_counts, force_group_nV, mts)
+        self.addComputeGlobal("unperturbed_pe", "energy")
 
 class VVVRIntegrator(LangevinSplittingIntegrator):
     """Create a velocity Verlet with velocity randomization (VVVR) integrator."""
```
bas-rustenburg: Be sure to remove the measure_protocol work flag. 

Also, under test_integrators.py, for the `test_vvvr_protocol_work_accumulation`  you will need to change the code to use the `ExternalPerturbationLangevinSplittingIntegrator` integrator (and probably just initialize it with the VVVR splitting).

pgrinaway: removed flag.

openmmtools/integrators.py
```diff
@@ -1185,25 +1095,562 @@ def substep_function(step_string):
...
+    def __init__(self,
+                 splitting="V R O R V",
+                 temperature=298.0 * simtk.unit.kelvin,
+                 collision_rate=1.0 / simtk.unit.picoseconds,
+                 timestep=1.0 * simtk.unit.femtoseconds,
+                 constraint_tolerance=1e-8,
+                 measure_shadow_work=False,
+                 measure_heat=True):
+
+        self.addGlobalVariable("protocol_work", 0)
```
bas-rustenburg: We can't define global variables before calling the CustomIntegrator init function. 

I'm running into this error:
```pytb
File "C:\GitHub\openmmtools\openmmtools\tests\test_integrators.py", line 219, in test_vvvr_protocol_work_accumulation
    integrator = integrators.ExternalPerturbationLangevinSplittingIntegrator(splitting="O V R V O", temperature=temperature)
  File "C:\GitHub\openmmtools\openmmtools\integrators.py", line 1638, in __init__
    self.addGlobalVariable("protocol_work", 0)
  File "C:\Users\rustenba\AppData\Local\Continuum\Miniconda3\envs\py35\lib\site-packages\simtk\openmm\openmm.py", line 10574, in addGlobalVariable
    return _openmm.CustomIntegrator_addGlobalVariable(self, name, initialValue)
Exception: in method 'CustomIntegrator_addGlobalVariable', argument 1 of type 'OpenMM::CustomIntegrator *'
```

openmmtools/integrators.py
```diff
@@ -999,31 +999,38 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
...
                  measure_heat=True,
-                 measure_protocol_work=False
                  ):
         """Create a Langevin integrator with the prescribed operator splitting.
 
         Parameters
         ----------
         splitting : string
-            Sequence of R, V, O (and optionally V{i}) substeps to be executed each timestep.
+            Sequence of R, V, O (and optionally V{i}), and { } substeps to be executed each timestep.
```
jchodera: Should we choose a different symbol from `{` or somehow denote these symbols separately, since `V{i}` is used to represent V0, V1, etc?

Do we also need to list the default `splitting` as 
```
splitting : str, optional, default='V R O O R V'
```

openmmtools/integrators.py
```diff
@@ -999,31 +999,38 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
...
+
     References
     ----------
     [Leimkuhler and Matthews, 2015] Molecular dynamics: with deterministic and stochastic numerical methods, Chapter 7
     """
 
+    _kinetic_energy = "0.5 * m * v * v"
+
     def __init__(self,
                  splitting="V R O R V",
```
jchodera: Should this be `"V R O O R V"`?

openmmtools/integrators.py
```diff
@@ -1037,121 +1044,24 @@ def __init__(self,
         constraint_tolerance : float
```
jchodera: Should this be `constraint_tolerance : float, optional, default=1.0e-8`?

openmmtools/integrators.py
```diff
@@ -999,31 +999,38 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
...
         splitting : string
-            Sequence of R, V, O (and optionally V{i}) substeps to be executed each timestep.
+            Sequence of R, V, O (and optionally V{i}), and { } substeps to be executed each timestep.
 
             Forces are only used in V-step. Handle multiple force groups by appending the force group index
             to V-steps, e.g. "V0" will only use forces from force group 0. "V" will perform a step using all forces.
+            { will cause metropolization, and must be followed later by a }.
+
 
         temperature : numpy.unit.Quantity compatible with kelvin, default: 298.0*simtk.unit.kelvin
```
jchodera: Should this have an `optional`?

openmmtools/integrators.py
```diff
@@ -1037,121 +1044,24 @@ def __init__(self,
         constraint_tolerance : float
             Tolerance for constraint solver
 
-        override_splitting_checks : boolean
```
jchodera: Should this have `optional, default=False`?

openmmtools/integrators.py
```diff
@@ -1037,121 +1044,24 @@ def __init__(self,
...
-            If True, skip sanity-checks on `splitting` string.
-
         measure_shadow_work : boolean
             Accumulate the shadow work performed by the symplectic substeps, in the global `shadow_work`
 
         measure_heat : boolean
```
jchodera: Should this have `optional, default=True`?

openmmtools/integrators.py
```diff
@@ -1037,121 +1044,24 @@ def __init__(self,
...
             Accumulate the heat exchanged with the bath in each step, in the global `heat`
-
-        measure_protocol_work : boolean
```
jchodera: Should this have `optional, default=False`?

openmmtools/integrators.py
```diff
@@ -999,31 +999,38 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
...
 
         Parameters
         ----------
         splitting : string
-            Sequence of R, V, O (and optionally V{i}) substeps to be executed each timestep.
+            Sequence of R, V, O (and optionally V{i}), and { } substeps to be executed each timestep.
 
             Forces are only used in V-step. Handle multiple force groups by appending the force group index
             to V-steps, e.g. "V0" will only use forces from force group 0. "V" will perform a step using all forces.
+            { will cause metropolization, and must be followed later by a }.
```
jchodera: Maybe note that the portion in between `{` and `}` will be Metropoloized, and must include all steps which generate shadow work?

openmmtools/integrators.py
```diff
@@ -1185,25 +1095,565 @@ def substep_function(step_string):
...
+        """
+
+        if step_string == "O":
+            self.O_step(measure_heat)
+        elif step_string == "R":
+            self.R_step(measure_shadow_work, n_R)
+        elif step_string == "{":
+            self.addComputePerDof("xold", "x")
+            self.addComputePerDof("vold", "v")
+        elif step_string == "}":
```
jchodera: Should this include a check to see if `self._metropolize = True`, and if `{` has already appeared?

openmmtools/integrators.py
maxentile: @jchodera : Is there are reason to prefer this here? A single `O` solve of length dt will have the same effect (in distribution) as two consecutive `O` solves of length dt/2 -- `O O` just takes 2x as many gaussian r.v.'s as `O`...

openmmtools/integrators.py
pgrinaway: I can change to a different character. That's probably a good idea.

>Do we also need to list the default splitting as

Perhaps, I just left the old docstring mostly unchanged.

openmmtools/integrators.py
pgrinaway: I can pre-check the string for that.

pgrinaway:
Ok, I've changed the metropolization syntax to `()` and made sure it happens only once. 

@maxentile would you mind taking a look at the docstring stuff John mentioned/committing an appropriate fix? I am not as familiar.

maxentile:
>@maxentile would you mind taking a look at the docstring stuff John mentioned/committing an appropriate fix? I am not as familiar.

Sure! Taking a look now.

openmmtools/integrators.py
jchodera: No reason to prefer this except for simplification and standardization of terminology to coincide with our Langevin paper, but it sounds like everything works fine if either `V R O R V` or `V R O O R V` are specified, so this is not critical.

pgrinaway:
Remaining tasks:

- [ ] Tests for `Alchemical*` integrator
- [ ] Change camel case to pythonic naming
- [ ] Replace GHMC integrator with subclass
- [ ] Revert to curly brace formulation for Metropolized block.

bas-rustenburg:
Spotted a potential bug. https://travis-ci.org/choderalab/openmmtools/jobs/218558296#L1601

Tagging @andrrizzi who wrote the test.

I'm guessing the alchemical integrator taking 1 required positional argument is breaking this. 
https://github.com/choderalab/openmmtools/blame/master/openmmtools/tests/test_integrators.py#L255

What would make sense, making a special case for the Alchemical integrator, or skipping it? I'm not sure what the purpose of the test is. Does it really need to test _all_ integrators?

bas-rustenburg:
Justed pushed my fixes to some of the unit tests so that they handle the AlchemicalLangevinSplittingIntegrator API, and I've updated its `__init__` to work.

pgrinaway:
Thanks @bas-rustenburg @maxentile

openmmtools/tests/test_integrators.py
```diff
@@ -312,3 +313,117 @@ def test_thermostated_integrator_hash():
...
+    """
+
+    #make a ghmc integrator with the default parameters
+    ghmc_integrator = GHMCIntegrator()
+
+    #use the reference platform, since this is for a test
+    platform = openmm.Platform.getPlatformByName("Reference")
+
+    #make a context
+    context = openmm.Context(system, ghmc_integrator, platform)
```
maxentile: Can probably move this out of the inner-loop function -- the test creates an OpenMM context 20,000 times, when the only thing that's changing between iterations is the initial positions, I think

openmmtools/tests/test_integrators.py
```diff
@@ -312,3 +313,117 @@ def test_thermostated_integrator_hash():
...
+
+    alchemical_ctx_forward = openmm.Context(modified_system, alchemical_integrator_forward, platform)
+    alchemical_ctx_reverse = openmm.Context(modified_system, alchemical_integrator_reverse, platform)
+
+    #get the forward work values:
+    positions = alanine_dipeptide.positions
+    w_f = numpy.zeros([n_iterations])
+    for i in range(n_iterations):
+        w_f[i], eq_positions = run_nonequilibrium_switching(modified_system, positions,
+                                              default_functions, alchemical_integrator_forward, 100, alchemical_ctx_forward,
```
maxentile: `100` was probably supposed to reference `nsteps` from L324?

openmmtools/tests/test_integrators.py
```diff
@@ -312,3 +313,117 @@ def test_thermostated_integrator_hash():
...
+                                              direction="forward")
+        positions = eq_positions
```
maxentile: It might make sense to sample a collection of `eq_positions` first, then draw from that collection with replacement.

openmmtools/tests/test_integrators.py
```diff
@@ -312,3 +313,117 @@ def test_thermostated_integrator_hash():
...
+    for parameter in alchemical_functions.keys():
+        context.setParameter(parameter, 0.0) if direction == "forward" else context.setParameter(parameter, 1.0)
+
+    #run some steps to equilibrate
+    ghmc_integrator.step(100)
+
+    eq_positions = context.getState(getPositions=True).getPositions(asNumpy=True)
+
+    alchemical_ctx.setPositions(eq_positions)
+
```
maxentile: Also `alchemical_ctx.setVelocitiesToTemperature(298 * unit.kelvin)` ?

jchodera:
>  Revert to curly brace formulation for Metropolized block.
 
Why revert to curly braces? Sorry if I missed this earlier!

maxentile:
>Why revert to curly braces? Sorry if I missed this earlier!

My request -- I would like to reserve `(`, `)` for nested / multitimestep syntax to be added later.

openmmtools/integrators.py
```diff
@@ -999,31 +1304,41 @@ class LangevinSplittingIntegrator(ThermostatedIntegrator):
...
+
     References
     ----------
+    [Nilmeier, et al. 2011] Nonequilibrium candidate Monte Carlo is an efficient tool for equilibrium simulation
     [Leimkuhler and Matthews, 2015] Molecular dynamics: with deterministic and stochastic numerical methods, Chapter 7
     """
 
     def __init__(self,
+                 alchemical_functions,
                  splitting="V R O R V",
```
sgill2: It might be helpful to change the default splitting to include an alchemical example using H. On related note, the splitting docstring should be updated to include H as an option.

pgrinaway: Will do, thanks.

bas-rustenburg:
I think this commit may have caused a conflict with master somehow.  https://github.com/choderalab/openmmtools/commit/4320f7fb2413a572d9b1feea9f494969bc6cd151

maxentile:
Ahh, my bad, should I revert that? I had updated from master there...

maxentile:
Ok, resolved -- conflict was version "0.9.1" --> "0.9.2"

pgrinaway:
Ok merging, tests pass.

jchodera:
Woohoo! Thanks!

PR #147, opened 2017-04-03
Fix serialization of MCMC container moves and CompoundThermodynamicState

andrrizzi:
Ready for review.

Previously, I converted lists inside the serialized representation of MCMCMoves and compound states to dictionaries to work around a problem of serializing `dict`s containing `list`s of `dict`s in netcdf. Now that we serialize `dict`s as strings instead, we can make this a little cleaner.

### DIFF
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -378,16 +378,12 @@ class SequenceMove(object):
         return iter(self.move_list)
 
     def __getstate__(self):
-        serialization = dict()
-        for i, move in enumerate(self.move_list):
-            serialization['move' + str(i)] = utils.serialize(move)
-        return serialization
+        serialized_moves = [utils.serialize(move) for move in self.move_list]
+        return dict(move_list=serialized_moves)
 
     def __setstate__(self, serialization):
-        self.move_list = [None for _ in range(len(serialization))]
-        # Restore moves in the correct order.
-        for i in range(len(serialization)):
-            self.move_list[i] = utils.deserialize(serialization['move' + str(i)])
+        serialized_moves = serialization['move_list']
+        self.move_list = [utils.deserialize(move) for move in serialized_moves]
 
 
 class WeightedMove(object):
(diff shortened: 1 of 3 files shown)


openmmtools/mcmc.py
```diff
@@ -378,16 +378,12 @@ def __iter__(self):
         return iter(self.move_list)
 
     def __getstate__(self):
-        serialization = dict()
-        for i, move in enumerate(self.move_list):
-            serialization['move' + str(i)] = utils.serialize(move)
-        return serialization
+        serialized_moves = [utils.serialize(move) for move in self.move_list]
```
Lnaden: So this is a dict with a single key? That seems odd. Did you instead me a dict of moves?

 Why not just dictionary comprehension then:
```python
return {move: utils.serialize(move) for move in self.move_list}
```

openmmtools/mcmc.py
```diff
@@ -378,16 +378,12 @@ def __iter__(self):
...
+        return dict(move_list=serialized_moves)
 
     def __setstate__(self, serialization):
-        self.move_list = [None for _ in range(len(serialization))]
-        # Restore moves in the correct order.
-        for i in range(len(serialization)):
-            self.move_list[i] = utils.deserialize(serialization['move' + str(i)])
+        serialized_moves = serialization['move_list']
+        self.move_list = [utils.deserialize(move) for move in serialized_moves]
```
Lnaden: This may need to be changed based on the above comment

openmmtools/mcmc.py
```diff
@@ -469,19 +465,19 @@ def apply(self, thermodynamic_state, sampler_state):
...
 
     def __getstate__(self):
-        serialization = dict()
-        for i, (move, weight) in enumerate(self.move_set):
-            serialization['move' + str(i)] = utils.serialize(move)
-            serialization['move' + str(i) + '_weight'] = weight
-        return serialization
+        serialized_moves = [utils.serialize(move) for move, _ in self.move_set]
+        weights = [weight for _, weight in self.move_set]
+        return dict(moves=serialized_moves, weights=weights)
```
Lnaden: This uses a different key to store the move list than in the previous case. If these are meant to be used together, or at least passed into each other, then I think you will want to change `moves` to `move_list`

openmmtools/states.py
```diff
@@ -2066,25 +2066,26 @@ def __getstate__(self):
         """Return a dictionary representation of the state."""
         # Create original ThermodynamicState to serialize.
         thermodynamic_state = object.__new__(self.__class__.__bases__[1])
-        thermodynamic_state.__dict__ = copy.deepcopy(self.__dict__)
+        thermodynamic_state.__dict__ = self.__dict__
```
Lnaden: Is there a reason this is not a deep copy any more? Do you want it pointing right at the same object?

openmmtools/mcmc.py
andrrizzi: Yeah, `__getstate__` should return a dictionary representation of the object. That's the signature for all `__getstate__` methods. The key must be a constant, otherwise `__setstate__` doesn't know what to read.

openmmtools/states.py
andrrizzi: I removed it because `utils.serialize` doesn't modify the passed arguments, and doing a `copy.deepcopy` involves copying a whole system, which may take time. It's just a small optimization.

openmmtools/mcmc.py
andrrizzi: They're not meant to be passed to each other, so it should be fine. Every object only cares about its own representation.

Lnaden:
okie-dokie, LGTM

andrrizzi:
Thanks! Switching `ISRELEASED` flag to `True` before merging.

Issue #164, opened 2017-04-25
Reset protocol work in `ExternalPerturbationLangevinIntegrator`


gregoryross:
I think the application of `ExternalPerturbationLangevinIntegrator` will be simplified if the 'protocol_work' is set to zero when 'first_step'=0. Currently, to correctly calculate the protocol work after many NCMC attempts, an instance of `ExternalPerturbationLangevinIntegrator` requires the following before running a NCMC attempt:
```
ncmc_integrator.setGlobalVariableByName("first_step", 0)
ncmc_integrator.setGlobalVariableByName("protocol_work", 0)
```
@bas-rustenburg, what do you think of wrapping `setGlobalVariableByName("protocol_work", 0)` inside [this if statement](https://github.com/choderalab/openmmtools/blob/master/openmmtools/integrators.py#L1551-L1554)? That way, before each NCMC protocol, one only needs to call `ncmc_integrator.setGlobalVariableByName("first_step", 0)` to reset the integrator completely.

gregoryross:
This issue was addressed in PR #165.

PR #165, opened 2017-04-25
Simplifying NCMC with ExternalPerturbationLangevinIntegrator

gregoryross:
This PR addresses Issue #164. With this commit, the protocol work variable in `ExternalPerturbationLangevinIntegrator` is set to zero when first_step = 0. This simplifies the implementation of `ExternalPerturbationLangevinIntegrator` for NCMC, which requires multiple sequential protocol work calculations. 

@bas-rustenburg, could you make sure that I've used `self.addComputeGlobal("protocol_work", "0.0")` correctly? I get confused with this versus  `self.setGlobalVariableByName("protocol_work", "0.0")` . Thanks!

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1520,7 +1520,13 @@ class NonequilibriumLangevinIntegrator(LangevinIntegrator):
 
 
 class ExternalPerturbationLangevinIntegrator(LangevinIntegrator):
-    """LangevinSplittingIntegrator that accounts for external perturbations and tracks protocol work."""
+    """
+    LangevinSplittingIntegrator that accounts for external perturbations and tracks protocol work.
+
+    If used before an NCMC attempt, initialize the protocol work with
+    >>> integrator.setGlobalVariableByName("first_step", 0)
+    where integrator is an instance of ExternalPerturbationLangevinIntegrator.
+    """
 
     def __init__(self,
                  splitting="V R O R V",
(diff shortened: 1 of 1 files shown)


openmmtools/integrators.py
```diff
@@ -1520,7 +1520,13 @@ def reset_integrator(self):
 
 
 class ExternalPerturbationLangevinIntegrator(LangevinIntegrator):
-    """LangevinSplittingIntegrator that accounts for external perturbations and tracks protocol work."""
+    """
+    LangevinSplittingIntegrator that accounts for external perturbations and tracks protocol work.
+
+    If used before an NCMC attempt, initialize the protocol work with
```
bas-rustenburg: This description is a bit vague. Why not write something along the lines of

`Reinitialize protocol work by using...` and leave NCMC out of it. 

This might also be a good opportunity to add the following to the docstring, right before this segment.

`The externally performed protocol work is accumulated in the protocol_work global variable.`

openmmtools/integrators.py
```diff
@@ -1551,6 +1557,7 @@ def add_integrator_steps(self, splitting, measure_shadow_work, measure_heat, ORV
         self.beginIfBlock("first_step < 1")
         self.addComputeGlobal("first_step", "1")
         self.addComputeGlobal("unperturbed_pe", "energy")
+        self.addComputeGlobal("protocol_work", "0.0")
```
bas-rustenburg: Looks good to me!

openmmtools/integrators.py
gregoryross: Good points!

bas-rustenburg:
Nice addition to the docstring!

bas-rustenburg:
[Fix open PR](https://trello.com/c/T7MBv94J/5-fix-open-pr)

bas-rustenburg:
Thanks Greg!

PR #168, opened 2017-04-29
Add more tests for ExternalPerturbationLangevinIntegrator

jchodera:
This adds some additional tests to try to figure out what is going on with @gregoryross's observation that there is a discrepancy between the protocol work measured from computing potential energy changes and reading from the integrator's accumulated protocol work.

So far, I am detecting no significant discrepancies.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,6 +18,7 @@ requirements:
     - setuptools
     - openmm
     - parmed
+    - tqdm
 
   run:
     - python
@@ -28,12 +29,13 @@ requirements:
     - six
     - openmm
     - parmed
-
+    - tqdm
 
 test:
   requires:
     - nose
     - pymbar
+    - tqdm
   imports:
     - openmmtools
 
--- a/openmmtools/tests/test_integrators.py
+++ b/openmmtools/tests/test_integrators.py
@@ -273,10 +273,11 @@ class TestExternalPerturbationLangevinIntegrator(TestCase):
         parameter_name = 'lambda_electrostatics'
         parameter_initial = 1.0
         parameter_final = 0.0
+        platform_names = [ openmm.Platform.getPlatform(index).getName() for index in range(openmm.Platform.getNumPlatforms()) ]
         for nonbonded_method in ['CutoffPeriodic', 'PME']:
             testsystem = testsystems.AlchemicalWaterBox(nonbondedMethod=getattr(app, nonbonded_method))
-            name = '%s with nonbondedMethod=%s' % (testsystem.name, nonbonded_method)
-            for platform_name in ['CPU', 'OpenCL', 'Reference']:
+            for platform_name in platform_names:
+                name = '%s %s %s' % (testsystem.name, nonbonded_method, platform_name)                
                 self.compare_external_protocol_work_accumulation(testsystem, parameter_name, parameter_initial, parameter_final, platform_name=platform_name, name=name)
 
     def compare_external_protocol_work_accumulation(self, testsystem, parameter_name, parameter_initial, parameter_final, platform_name='Reference', name=None):
--- a/openmmtools/tests/test_integrators.py
+++ b/openmmtools/tests/test_integrators.py
@@ -294,7 +294,7 @@ class TestExternalPerturbationLangevinIntegrator(TestCase):
 
         # TODO: Set precision and determinism if platform is ['OpenCL', 'CUDA']
 
-        nsteps = 100
+        nsteps = 20
         kT = kB * temperature
         integrator = integrators.ExternalPerturbationLangevinIntegrator(splitting="O V R V O", temperature=temperature)
         context = openmm.Context(system, integrator, platform)
--- a/openmmtools/tests/test_integrators.py
+++ b/openmmtools/tests/test_integrators.py
@@ -267,23 +267,32 @@ class TestExternalPerturbationLangevinIntegrator(TestCase):
             self.compare_external_protocol_work_accumulation(testsystem, parameter_name, parameter_initial, parameter_final, platform_name=platform_name)
 
     def test_protocol_work_accumulation_waterbox(self):
-        """Testing protocol work accumulation for ExternalPerturbationLangevinIntegrator with AlchemicallyDischargedWaterBox
+        """Testing protocol work accumulation for ExternalPerturbationLangevinIntegrator with AlchemicalWaterBox
         """
-        testsystem = testsystems.AlchemicallyDischargedWaterBox()
+        from simtk.openmm import app
         parameter_name = 'lambda_electrostatics'
         parameter_initial = 1.0
         parameter_final = 0.0
-        for platform_name in ['Reference', 'CPU']:
-            self.compare_external_protocol_work_accumulation(testsystem, parameter_name, parameter_initial, parameter_final, platform_name=platform_name)
+        for nonbonded_method in ['CutoffPeriodic', 'PME']:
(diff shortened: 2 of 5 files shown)


jchodera:
@gregoryross : I've now found a failure case!

By adding a variant of `WaterBox` in which the first water can be discharged via the context parameter `lambda_electrostatics`:
```
FAIL: Testing protocol work accumulation for ExternalPerturbationLangevinIntegrator with AlchemicallyDischargedWaterBox
  File "/Users/choderaj/github/choderalab/openmmtools/openmmtools.choderalab/openmmtools/tests/test_integrators.py", line 277, in test_protocol_work_accumulation_waterbox
    self.compare_external_protocol_work_accumulation(testsystem, parameter_name, parameter_initial, parameter_final, platform_name=platform_name)
  File "/Users/choderaj/github/choderalab/openmmtools/openmmtools.choderalab/openmmtools/tests/test_integrators.py", line 313, in compare_external_protocol_work_accumulation
    self.assertAlmostEqual(external_protocol_work, integrator_protocol_work, msg=message)
AssertionError: 0.3348818945502135 != 0.33486957771584547 within 7 places : 
protocol work discrepancy noted for AlchemicallyDischargedWaterBox on platform CPU
step     0 : external     3.348819e-01 kT | integrator     3.348696e-01 kT | difference     1.231683e-05 kT

Ran 2 tests in 76.463s
```
Notably, the work magnitudes for the harmonic oscillator test (which doesn't fail) has no problem, so I'm guessing this *might* be an OpenMM bug associated with `NonbondedForce`.

The [OpenMM 7.1.1 bugfix release](https://github.com/pandegroup/openmm/issues/1793) is imminently due, so we would want to open an issue soon if it is an OpenMM bug.

jchodera:
I've expanded the tests to test PME and `CutoffPeriodic`. Only PME shows the discrepancy.

I wonder if the CPU implementation of PME is simply non-deterministic because of the use of multiple threads.

jchodera:
Setting `OPENMM_CPU_THREADS` to 1 eliminates the issue. I imagine it will for your other tests too, @gregoryross. I think this is simply the non-determinism issue with CPU multithreading.

These tests should be ready to merge, and I think the problem is solved.

jchodera:
I've opened an OpenMM issue to see if there is a way to avoid the nondeterminism without dropping down to a single CPU thread: https://github.com/pandegroup/openmm/issues/1799

jchodera:
Can someone review this?

The py34 error is the same as in #169 and is still unexplained (and presumably unrelated):
```
6.08s$ conda build devtools/conda-recipe
BUILD START: openmmtools-dev-0.0.0-py34_0
updating index in: /home/travis/miniconda/conda-bld/linux-64
updating index in: /home/travis/miniconda/conda-bld/noarch
  File "/home/travis/miniconda/bin/conda-build", line 6, in <module>
    sys.exit(conda_build.cli.main_build.main())
  File "/home/travis/miniconda/lib/python2.7/site-packages/conda_build/cli/main_build.py", line 322, in main
    execute(sys.argv[1:])
  File "/home/travis/miniconda/lib/python2.7/site-packages/conda_build/cli/main_build.py", line 313, in execute
    noverify=args.no_verify)
  File "/home/travis/miniconda/lib/python2.7/site-packages/conda_build/api.py", line 97, in build
    need_source_download=need_source_download, config=config)
  File "/home/travis/miniconda/lib/python2.7/site-packages/conda_build/build.py", line 1536, in build_tree
    .format(recipe, e.pkgs) + "\n\n" + extra_help)
AttributeError: 'UnsatisfiableError' object has no attribute 'pkgs'
```

openmmtools/tests/test_integrators.py
```diff
@@ -247,7 +253,76 @@ def test_external_protocol_work_accumulation():
...
+        """Testing protocol work accumulation for ExternalPerturbationLangevinIntegrator with HarmonicOscillator
+        """
+        testsystem = testsystems.HarmonicOscillator()
+        parameter_name = 'testsystems_HarmonicOscillator_x0'
+        parameter_initial = 0.0 * unit.angstroms
+        parameter_final = 10.0 * unit.angstroms
+        for platform_name in ['Reference', 'CPU']:
+            self.compare_external_protocol_work_accumulation(testsystem, parameter_name, parameter_initial, parameter_final, platform_name=platform_name)
+
+    def test_protocol_work_accumulation_waterbox(self):
```
gregoryross: I'm running this test locally, which often fails for the `PME CPU on platform CPU` iteration. The difference between the integrator and external measurements of external work differs only by `1.231683e-05 kT`, which is greater than the `1e-07 kT` default for the `assertAlmostEqual()` in `compare_external_protocol_work_accumulation()`. How about we relax the precision to something like `1e-04 kT`?

openmmtools/tests/test_integrators.py
```diff
@@ -247,7 +253,76 @@ def test_external_protocol_work_accumulation():
...
+            name = testsystem.name
+
+        from openmmtools.constants import kB
+        system, topology = testsystem.system, testsystem.topology
+        temperature = 298.0 * unit.kelvin
+        platform = openmm.Platform.getPlatformByName(platform_name)
+
+        # TODO: Set precision and determinism if platform is ['OpenCL', 'CUDA']
+
+        nsteps = 100
```
gregoryross: These tests are very slow because `compare_external_protocol_work_accumulation()` is called several times. How about lowering `nsteps` to something like 20?

devtools/conda-recipe/meta.yaml
```diff
@@ -28,12 +29,13 @@ requirements:
     - six
     - openmm
     - parmed
-
+    - tqdm
```
gregoryross: I don't have a strong feeling either way, but do we really need this dependence? It's helpful when running the tests locally, but the new tests could be re-written without it.

jchodera:
> These tests look good, but one is failing locally due to the high degree of precision required in the assertions. @jchodera, I'm happy to relax those if you like.

I don't see that failure on travis. Could you link to the failure?

openmmtools/tests/test_integrators.py
jchodera: Sure, we can relax the precision, but are you sure you are setting `OPENMM_CPU_THREADS=1` in your shell BEFORE you run the tests? The [travis tests do this](https://github.com/choderalab/openmmtools/pull/168/files#diff-354f30a63fb0907d4ad57269548329e3R41).

openmmtools/tests/test_integrators.py
jchodera: Sure!

devtools/conda-recipe/meta.yaml
jchodera: It's on conda, so there's no reason not to include it, is there?

openmmtools/tests/test_integrators.py
gregoryross: Ah! You're right! The test passes fine locally with `OPENMM_CPU_THREADS=1`.

devtools/conda-recipe/meta.yaml
gregoryross: Leaving it in is absolutely fine with me.

gregoryross:
@jchodera, I've reduced the number of NCMC steps in the new unit tests so that they run much faster now. Once the continuous integration steps have passed I'm happy to merge if we're okay with ignoring the weird python 3.4 errors.

openmmtools/tests/test_integrators.py
jchodera: This is just related to the same issue as https://github.com/pandegroup/openmm/issues/1799 then

jchodera:
Thanks! 

We can figure out the py34 error in a separate PR, but this is low priority since it doesn't seem to affect the science.

PR #174, opened 2017-05-01
Fix step-length error in add_integrator_steps

maxentile:
Each step was being integrated for `dt / ORV_counts["R"]`


maxentile:
Ahh, sorry, I had misread-- `substep_function` accepts `n_R` regardless of `step_string`.

Closing.

Issue #173, opened 2017-05-01
ExternalPerturbationLangevinIntegrator produces strange energies with `updateParametersInContext`


gregoryross:
I've encountered large discrepancies in calculated energies when using `ExternalPerturbationLangevinIntegrator`. The code below compares protocol work that is accumulated inside `ExternalPerturbationLangevinIntegrator` and outside the integrator using `getState()`. The following example is a model of an NCMC procedure where no perturbation is performed on the system but `updateParametersInContext()` is still called. When `updateParametersInContext()` is called, the different methods of calculating protocol work produce values that differ on the order of 1kJ/mol.

Tagging @jchodera, @bas-rustenburg, @pgrinaway and @maxentile.
``` python
from simtk import openmm, unit
from simtk.openmm import app
import numpy as np

# Using one CPU thread
import os
os.environ['OPENMM_CPU_THREADS'] = '1'

# Long range method
nonbonded_method = 'CutoffPeriodic'

# Creating a waterbox
from openmmtools.testsystems import WaterBox
wbox = WaterBox(box_edge=20.0*unit.angstrom , nonbondedMethod=getattr(app, nonbonded_method))
wbox.system.addForce(openmm.MonteCarloBarostat(1*unit.atmospheres, 300*unit.kelvin))

# Extracting the nonbonded force
non_bonded_force = wbox.system.getForce(2)

# The integrator to perform the equilibrium dynamics
from openmmtools.integrators import ExternalPerturbationLangevinIntegrator
integrator = ExternalPerturbationLangevinIntegrator(temperature=300*unit.kelvin, collision_rate=50.0 / unit.picosecond, timestep=1.0 * unit.femtosecond)

# Creating the context
platform = openmm.Platform.getPlatformByName('CPU')
context = openmm.Context(wbox.system, integrator, platform)
context.setPositions(wbox.positions)

# Running some equilibrium dynamics
integrator.step(100)

# The number of NCMC type iterations and NCMC steps per iteration.
niterations = 20
ncmc_steps = 10

internal_work = np.zeros(niterations)
external_work = np.zeros(niterations)

# Whether to call updateParametersInContext. If True, then assertion below will fail.
update_parameters = True

# A model of NCMC without perturbation but using updateParametersInContext
for i in range(niterations):
    integrator.setGlobalVariableByName('first_step',0)
    integrator.setGlobalVariableByName('protocol_work',0)
    for s in range(ncmc_steps):
        integrator.step(1)
[...]

PR #175, opened 2017-05-02
ExternalPerturbationLangevinIntegrator barostat fix

gregoryross:
Fix made to address Issue #173. This fix addresses the fact that a barostat, when present, was erroneously affecting the protocol work when calculated with the `ExternalPerturbationLangevinIntegrator`. The work-around was to move 
```
self.addUpdateContextState()
self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
````
to inside `add_integrator_steps()`. This makes it possible to separate the protocol work measurement from energy changes due to the barostat or the center of mass remover.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1454,10 +1454,6 @@ class NonequilibriumLangevinIntegrator(LangevinIntegrator):
         Override the base class to insert reset steps around the integrator.
         """
 
-        # Integrate
-        self.addUpdateContextState()
-        self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
-
         #if the step is zero,
         self.beginIfBlock('step = 0')
         self.addConstrainPositions()
@@ -1576,9 +1572,6 @@ class ExternalPerturbationLangevinIntegrator(LangevinIntegrator):
         self.addComputeGlobal("protocol_work", "protocol_work + (perturbed_pe - unperturbed_pe)")
 
         # Computing context updates, such as from the barostat, _after_ computing protocol work.
-        self.addUpdateContextState()
-        self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
-
         super(ExternalPerturbationLangevinIntegrator, self).add_integrator_steps(splitting, measure_shadow_work, measure_heat, ORV_counts, force_group_nV, mts)
         self.addComputeGlobal("unperturbed_pe", "energy")
 
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -954,10 +954,6 @@ class LangevinIntegrator(ThermostatedIntegrator):
             self.addPerDofVariable("vold", 0)
             self.addPerDofVariable("xold", 0)
 
-        # Integrate
-        self.addUpdateContextState()
-        self.addComputeTemperatureDependentConstants({"sigma": "sqrt(kT/m)"})
-
         self.add_integrator_steps(splitting, measure_shadow_work, measure_heat, ORV_counts, force_group_nV, mts)
 
     def add_integrator_steps(self, splitting, measure_shadow_work, measure_heat, ORV_counts, force_group_nV, mts):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: openmmtools/integrators.py [hunk 1]
```diff
@@ -1453,6 +1453,11 @@ def add_integrator_steps(self, splitting, measure_shadow_work, measure_heat, ORV
         """
         Override the base class to insert reset steps around the integrator.
         """
+
+        # Integrate
```
jchodera: Calling `self.addUpdateContextState()` here is not appropriate because you call `super(NonequilibriumLangevinIntegrator, self).add_integrator_steps()` a few lines later, which also calls `self.addUpdateContextState()`, which would mean you call this twice per step.

openmmtools/integrators.py
```diff
@@ -1566,7 +1571,14 @@ def add_integrator_steps(self, splitting, measure_shadow_work, measure_heat, ORV
         self.addComputeGlobal("unperturbed_pe", "energy")
         self.addComputeGlobal("protocol_work", "0.0")
         self.endBlock()
+
+        # Calculate the protocol work
         self.addComputeGlobal("protocol_work", "protocol_work + (perturbed_pe - unperturbed_pe)")
+
+        # Computing context updates, such as from the barostat, _after_ computing protocol work.
```
jchodera: Same thing here.

gregoryross:
Okay thanks @jchodera, my mistake. Working on a fix now.

jchodera:
Thanks for adding the test!

I wonder if we should add a test to make sure there is one and only one `addUpdateContextState()` in the entire integrator sequence. You can use a [scheme like this](https://github.com/choderalab/integrator-benchmark/blob/master/benchmark/serialization/serialize_integrators.py#L31-L35) to make sure there is exactly one [step type 5](https://github.com/choderalab/integrator-benchmark/blob/master/benchmark/serialization/serialize_integrators.py#L17) in the integrator.

gregoryross:
An extra test similar to the example you posted sounds like a good idea. I'll add one for `NonequilibriumLangevinIntegrator` and `ExternalPerturbationLangevinIntegrator`.

jchodera:
> An extra test similar to the example you posted sounds like a good idea. I'll add one for NonequilibriumLangevinIntegrator and ExternalPerturbationLangevinIntegrator.

Maybe you could just have that test [loop over all integrators in `openmmtools.integrators` like this](https://github.com/choderalab/openmmtools/blob/master/openmmtools/tests/test_integrators.py#L159-L162)?
```
@andrrizzi has conveniently [provided a class](https://github.com/choderalab/openmmtools/blob/master/openmmtools/tests/test_integrators.py#L42-L65) that gives a list of all integrators.

**Comment by gregoryross** [comment]
Oh yeah, that's nice!

**Comment by gregoryross** [comment]
Assuming the continuous-integration tests pass (with the exception for python 3.4) would you mind approving this PR @jchodera and @bas-rustenburg?

File: openmmtools/tests/test_integrators.py [hunk 3]
```diff
@@ -186,6 +187,27 @@ def test_integrator_decorators():
...
 
+def test_single_force_update():
+    """
+    Ensures that addUpdateContextState() is called only once for all custom integrators, except the
+    NonequilibriumLangevinIntegrator which requires an alchmical system.
+    """
+    custom_integrators = get_all_custom_integrators()
+    for integrator_name, integrator_class in custom_integrators:
+        # The NonequilibriumLangevinIntegrator requires an alchemical function.
+        if issubclass(integrator_class, integrators.NonequilibriumLangevinIntegrator):
```
jchodera: There's no reason not to test all integrators, since all should have exactly one instance of `addUpdateContextState()`. I'll change this before I merge.

jchodera: I misread this! It *does* test all integrators. Thanks!

jchodera:
Merging this, since the py3.4 issue is the only one in the travis tests.

PR #169, opened 2017-04-29
Improve PME and RF support in alchemy

jchodera:
This PR adds a number of new features to `alchemy`, along with some useful updates to `testsystems` to help improve testing of electrostatics handling:

#### Unified defaults for `ewaldErrorTolerance`, `cutoff`, and `switch_width` for `testsystems`

Solvated protein test systems previously used a variety of different default parameters for `ewaldErrorTolerance`, `cutoff`, and `switch_width`, sometimes even relying on OpenMM defaults (which can give poor reweighting behavior). These systems now utilize a single global set of options defined at the top of `testsystems.py` with improved default parameters:
```python
DEFAULT_EWALD_ERROR_TOLERANCE = 1.0e-5 # default Ewald error tolerance
DEFAULT_CUTOFF_DISTANCE = 10.0 * unit.angstroms # default cutoff distance
DEFAULT_SWITCH_WIDTH = 1.5 * unit.angstroms # default switch width
```
This may benefit other projects as well, such as @maxentile's Langevin integrator benchmark.

#### Handling of alchemical region electrostatics for PME

The new `alchemical_pme_treatment` argument to `AlchemicalFactory` allows the user to select two different behaviors for how alchemical region electrostatics should be treated when the system uses PME:
* `direct-space` (default): Only the direct-space interactions, which behaves as `erfc(alpha*r)/r`, are included for interactions involving the alchemical region. This is the current behavior, and is retained as the default.
* `coulomb`: The standard Coulomb interaction, which behaves as `1/r`, is used for interactions involving the alchemical region. A switch is added to ensure force continuity. The switch width is controlled by the optional `switch_width` parameter.

After extensive testing with `TestAlchemicalFactorySlow.test_overlap()`, it appears that the current behavior (`direct-space`) has much lower energy stddev than `coulomb`, which is surprising to me. More study is needed, I imagine, but I am retaining the current behavior for now but leaving the new option is as an optional choice.

#### Handling of alchemical region electrostatics for reaction field (RF)

The new `alchemical_rf_treatment` argument to `AlchemicalFactory` allows the user to select two different behaviors for how alchemical region electrostatics should be treated when the system uses RF:
[...]

PR #178, opened 2017-05-04
DictYamlLoader/Dumper explicitly handle ndarrays

andrrizzi:
Here are the changes imported from Yank. The Python 3.4 error is due to a conda conflict that I'm fixing in #169 .


Lnaden:
Sure

PR #187, opened 2017-05-11
Unit test for ExternalPerturbationLangevinIntegrator.reset_protocol_work()

gregoryross:
A new feature was added in the previous release. This PR adds a unit test that ensures that protocol work is correctly set to zero with the command reset_protocol_work().

The unit tests for ExternalPerturbationLangevinIntegrator previously shared a lot of the same code, and multiple unit tests were performed in one "test". This PR refactors the shared code to make the unit tests for ExternalPerturbationLangevinIntegrator simpler and more extensible.

### DIFF
--- a/openmmtools/tests/test_integrators.py
+++ b/openmmtools/tests/test_integrators.py
@@ -375,9 +375,9 @@ class TestExternalPerturbationLangevinIntegrator(TestCase):
         testsystem.system.addForce(openmm.MonteCarloBarostat(1 * unit.atmospheres, temperature, 2))
         context, integrator = self.create_system(testsystem, parameter_name, parameter_initial, temperature, platform_name)
 
-        assert (integrator.getGlobalVariableByName('protocol_work') == 0)
+        assert (integrator.get_protocol_work(dimensionless=True) == 0)
         integrator.step(5)
-        assert(integrator.getGlobalVariableByName('protocol_work') == 0)
+        assert(integrator.get_protocol_work(dimensionless=True) == 0)
 
     def test_reset_protocol_work(self):
         """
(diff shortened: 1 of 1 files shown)


jchodera:
> This PR adds a unit test that ensures that protocol work is correctly set to zero with the command reset_protocol_work().

You probably want to use the more general `integrator.reset()`, which ensures that *all* quantities (step count, protocol work, shadow work) are reset together.

jchodera:
I see---this is intended just to check the `reset_protocol_work` functionality. It's fine as-is then!

openmmtools/tests/test_integrators.py
```diff
@@ -310,6 +310,105 @@ def test_external_protocol_work_accumulation():
...
+        for step in range(nsteps):
+            lambda_value = float(step + 1) / float(nsteps)
+            parameter_value = parameter_initial * (1 - lambda_value) + parameter_final * lambda_value
+            initial_energy = context.getState(getEnergy=True).getPotentialEnergy()
+            context.setParameter(parameter_name, parameter_value)
+            final_energy = context.getState(getEnergy=True).getPotentialEnergy()
+            external_protocol_work += (final_energy - initial_energy) / kT
+            integrator.step(1)
+
+        integrator_protocol_work = integrator.getGlobalVariableByName('protocol_work') * unit.kilojoules_per_mole / kT
```
jchodera: Don't use the global variables to access work! Always use `integrator.get_protocol_work(dimensionless=True)` if you want the dimensionless work!

jchodera:
Whoops---I found one issue with your test. See the inline comment above.

openmmtools/tests/test_integrators.py
gregoryross: Okay, it's added.

jchodera:
Thanks!

jchodera:
It looks like this test is frequently timing out:
```
Testing protocol work accumulation for ExternalPerturbationLangevinIntegrator with AlchemicalWaterBox ... 
```
Can you mark that as `@attr('slow')` perhaps so that it is skipped on travis?

gregoryross:
That test (`test_protocol_work_accumulation_waterbox()`) is repeated for all platforms, including `Reference` which is super slow. Instead of skipping it on travis, how about testing only `CPU` platform. As long as the other platforms remain tested in `test_protocol_work_accumulation_harmonic_oscillator()` the unit tests will have good coverage.

Issue #190, opened 2017-05-13
Implement keyword tagging system for testsystems to allow for auto-discovery of test systems


jchodera:
There are so many test systems and variations in `openmmtools.testsystems` that it would be very useful to implement a keyword-tagging system to allow for test system auto-discovery based on keyword for making it easier to shove a lot of tests through code.

Example keywords:
* `vacuum`, `NoCutoff` - test systems that don't have a cutoff
* `fast` - test systems that are small and fast
* `explicit-solvent` - test systems in explicit solvent
* `protein` - test systems that contain a protein
* `artificial` - test systems that are mainly meant to test weird OpenMM forces
* `custom-forces` - test systems that use OpenMM `Custom*Forces`
* `waterbox` - test systems that are variations on a water box

and so on.

PR #197, opened 2017-05-14
[WIP] Implement basic tagging system for testsystems

jchodera:
Addresses #190 

This PR begins to implement a system where we can add keyword tags to each system to enable autodiscovery of testsystems that have useful properties, like are quick to run on travis (or too slow to run on travis).

Instead of having the initializer define the tags, it would be useful to be able to query this without having to actually construct the object, since that construction itself can be slow. I'm still trying to think of how best to tackle that.

### DIFF
--- a/openmmtools/tests/test_testsystems.py
+++ b/openmmtools/tests/test_testsystems.py
@@ -85,6 +85,7 @@ def test_properties_all_testsystems():
         logging.info(f.description)
         yield f
 
+# TODO: Convert these to 'fast' tags within testsystems
 fast_testsystems = [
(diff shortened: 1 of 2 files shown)


jchodera:
Any thoughts on this approach from @Lnaden @andrrizzi @pgrinaway?

andrrizzi:
Nothing particularly smart. The only thing that comes to mind is to maintain a dictionary `tuple_of_tags: test_system`, but you'd still have to implement the infrastructure to cycle over all test systems if you only need to match a subset of tags that compose the key.

PR #200, opened 2017-05-18
Memory optimization for ThermodynamicState and reset velocities

andrrizzi:
Ready for review.

- Memory optimization of system hold by `ThermodynamicState`.
  - `ThermodynamicState` now uses a `weakref` cache to keep in memory only one instance of standard system per compatible state.
  - Copy and deepcopy defaults have been overridden to avoid copying the internal system.
  - Deserialization has been optimized to avoid deserializing the system when possible and using the cache instead.
- Optimization of how we reassign velocities in `BaseIntegratorMove`.
  - I realized today (thanks to @maxentile) that the current approach to `setVelocitiesToTemperature` ended up setting the velocities twice: first with `SamplerState.apply_to_context()` and then with `setVelocitiesToTemperature()`. This adds a `reassign_velocities` argument in the constructor of `BaseIntegratorMove` and fixes the problem.
- Switch to gBAOAB as fallback integrator in `ContextCache` when the user doesn't care about the integrator and there are no cached `Context` in the given `ThermodynamicState`.

### DIFF
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -671,20 +671,28 @@ class BaseIntegratorMove(object):
             # Subclasses may implement _before_integration().
             self._before_integration(context, thermodynamic_state)
 
-            # Run dynamics.
-            timer.start("{}: step({})".format(move_name, self.n_steps))
-            integrator.step(self.n_steps)
-            timer.stop("{}: step({})".format(move_name, self.n_steps))
+            try:
+                # Run dynamics.
+                timer.start("{}: step({})".format(move_name, self.n_steps))
+                integrator.step(self.n_steps)
+            except Exception:
+                # Catches particle positions becoming nan during integration.
+                restart = True
+            else:
+                timer.stop("{}: step({})".format(move_name, self.n_steps))
 
-            # We get also velocities here even if we don't need them because we
-            # will recycle this State to update the sampler state object. This
-            # way we won't need a second call to Context.getState().
-            context_state = context.getState(getPositions=True, getVelocities=True, getEnergy=True,
-                                             enforcePeriodicBox=thermodynamic_state.is_periodic)
+                # We get also velocities here even if we don't need them because we
+                # will recycle this State to update the sampler state object. This
+                # way we won't need a second call to Context.getState().
+                context_state = context.getState(getPositions=True, getVelocities=True, getEnergy=True,
+                                                 enforcePeriodicBox=thermodynamic_state.is_periodic)
 
-            # Check for NaNs in energies.
-            potential_energy = context_state.getPotentialEnergy()
-            if np.isnan(potential_energy.value_in_unit(potential_energy.unit)):
+                # Check for NaNs in energies.
+                potential_energy = context_state.getPotentialEnergy()
+                restart = np.isnan(potential_energy.value_in_unit(potential_energy.unit))
+
+            # Restart the move if we found NaNs.
+            if restart:
                 err_msg = ('Potential energy is NaN after {} attempts of integration '
                            'with move {}'.format(attempt_counter, self.__class__.__name__))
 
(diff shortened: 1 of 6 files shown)


jchodera:
I'm hoping @Lnaden has time to review this in detail since I'm a bit swamped.

Lnaden:
Yeah, I'll take a look at it

openmmtools/mcmc.py
```diff
@@ -572,6 +572,9 @@ class BaseIntegratorMove(object):
     context_cache : openmmtools.cache.ContextCache, optional
         The ContextCache to use for Context creation. If None, the global cache
         openmmtools.cache.global_context_cache is used (default is None).
+    reassign_velocities : bool, optional
+        If True, the velocities will be reassigned from the Maxwell-Boltzmann
+        distribution at the beginning of the move (default is False).
```
Lnaden: Out of curiosity, is there any reason why we *wouldn't* want to re-assign velocities?

openmmtools/mcmc.py
```diff
@@ -656,24 +662,37 @@ def apply(self, thermodynamic_state, sampler_state):
 
         # Perform the integration.
         for attempt_counter in range(self.n_restart_attempts + 1):
-            sampler_state.apply_to_context(context)
 
+            # If we reassign velocities, we can ignore the ones in sampler_state.
+            sampler_state.apply_to_context(context, ignore_velocities=self.reassign_velocities)
```
Lnaden: This isn't a problem, but why does this function need to be able to ignore velocities at all? In asks the question "Why can't we ignore positions" (from a user perspective). Should this function instead just apply to context the whole sampler state, then the user can shuffle velocities like you do here in the next line?

openmmtools/states.py
```diff
@@ -419,33 +410,43 @@ def get_system(self, remove_thermostat=False, remove_barostat=False):
...
 
     @temperature.setter
     def temperature(self, value):
         if value is None:
             raise ThermodynamicsError(ThermodynamicsError.NONE_TEMPERATURE)
-        self._set_system_thermostat(self._system, value)
-        barostat = self._barostat
-        if barostat is not None:
-            self._set_barostat_temperature(barostat, value)
+        self._temperature = value
```
Lnaden: Why is the Barostat protected with an old barostat and not the temperature? (I'm assuming its because we dont allow NVE sims, so T will not change, but its worth asking)

openmmtools/states.py
```diff
@@ -457,17 +458,22 @@ def pressure(self):
...
-        return barostat.getDefaultPressure()
+        return self._pressure
 
     @pressure.setter
-    def pressure(self, value):
-        # Invalidate cache if the ensemble changes.
-        if (value is None) != (self._barostat is None):
-            self._cached_standard_system_hash = None
-        self._set_system_pressure(self._system, value)
+    def pressure(self, new_pressure):
```
Lnaden: Would it be easier and cleaner to instead store the "old state" as a complete object? Or do you think storing each component is safer?

openmmtools/mcmc.py
andrrizzi: Fair point. The main two cases I had in mind here were in a sequence of moves (in case you wanted to use the final velocities of the previous moves) or for MC integrators that don't sample velocities. The original implementation of `GHMCMove` also didn't reassign velocities so I didn't modify it. In practice, we probably won't use this very often.

openmmtools/mcmc.py
andrrizzi: This is just an optimization thing. It should save some time passing information to the GPU at each iteration.

openmmtools/states.py
andrrizzi: > its because we dont allow NVE sims

Partially, yes. The main reason for protecting the `barostat` and `pressure` properties with their old values is that non periodic systems cannot have a barostat, so if this goes wrong (or anything else like setting an unsupported barostat) we need to recover before raising the exception. This is not a problem when setting the temperature.

openmmtools/states.py
andrrizzi: This is related to the previous comment. We're only changing the pressure here but we need to hold temporarily the old pressure to keep the object consistent in case the exception is catched when trying to set a barostat to a non-periodic system. The rest of the state is left untouched.

Lnaden:
Thanks for answering that, looks fine otherwise

andrrizzi:
Sure! Thanks for the review!

PR #203, opened 2017-05-22
[WIP] Initial samplers infrastructure

jchodera:
Initial import of samplers infrastructure with a start at documentation of the intended interface.


jchodera:
@pgrinaway : The only useful thing to look at so far is the [documentation of how to use the samplers](https://github.com/choderalab/openmmtools/blob/d6b46150d6a1c5e120f5bfd4f2a87a1f9362d401/docs/sampling.rst).

Also, should we call them `openmmtools.samplers.ExpandedEnsembleSampler` and `openmmtools.samplers.SAMSSampler`, or just ``openmmtools.samplers.ExpandedEnsemble` and `openmmtools.samplers.SAMS`?

jchodera:
And is this idea of "you have to initialize the stack of samplers yourselves, feeding each layer to the next" still the best approach, or should we have the sampler always initialize the stack below it? Maybe we could have the best of both worlds with convenience factories to do that if you really want it to be fully automated?

jchodera:
@pgrinaway : I have a [bunch of `SAMS` test systems](https://github.com/choderalab/openmmtools/blob/0f5742f75e51bdcc14452bbff2957537fe45786a/openmmtools/tests/test_samplers.py#L88-L588) that currently would live in `openmmtools.tests.test_samplers`, but it seems like these might be useful for docstrings, examples, and papers. Is there someplace more natural that these should live? For example, I could potentially move them to `openmmtools.testsystems.sams` so you could, for example, use:
```python
from openmmtools.testsystems.sams import AblImatinibExplicitAlchemical
testsystem = AblImatinibExplicitAlchemical()
testsystem.sams_sampler.run()
```
Does that make sense?

jchodera:
I'm actually not sure which of these locations is best:
1. `from openmmtools.testsystems.sams import AblImatinibExplicitAlchemical`
2. `from openmmtools.testsystems.samplers.sams import AblImatinibExplicitAlchemical`
3. `from openmmtools.samplers.sams.testsystems import AblImatinibExplicitAlchemical`

jchodera:
Just a heads-up to @andrrizzi that I also got a head-start on pulling in the `mpi` and `repex` modules into `openmmtools` as `openmmtools.distributed.mpi` and `openmmtools.samplers.repex`. I'm still fixing up the tests and docs, but code has all been pulled in.

andrrizzi:
> Just a heads-up to @andrrizzi that I also got a head-start on pulling in the `mpi` and `repex` modules

I would have preferred to see if we could keep at least part of the commit history, but that's ok. I think you'll have to import also `test_repex.py` and add `netcdf` dependency to make things work then.

jchodera:
> I would have preferred to see if we could keep at least part of the commit history, but that's ok. I think you'll have to import also test_repex.py and add netcdf dependency to make things work then.

Oof, sorry about this. I can strip it out again and just leave the infrastructure in place to import `repex.py` and its commit history directly in a separate PR, if you would prefer!

andrrizzi:
> I can strip it out again

Let me check first if it's possible to do it, thanks. It should be straightforward for the MPI module but for repex it's a little more complicated because it has more promiscuous history. If it is, I can do it and push it directly into this PR.

A simpler alternative could be maybe to point to the history of the imported file in the initial commit message.

Sorry if I'm being weird about this. The reason is that I got the impression that commit history is the only thing that counts for companies when they want to verify your contribution to some module during the hiring process, and this becomes harder to do if the history is scattered.

Lnaden:
Transferring history between GitHub packages is really hard, especially if there are multiple files involved and very much so if those files also need to be moved. I had this problem when I moved the storage module from YANK to OpenMMTools. I can try to help if need be.

jchodera:
> Sorry if I'm being weird about this. The reason is that I got the impression that commit history is the only thing that counts for companies when they want to verify your contribution to some module during the hiring process, and this becomes harder to do if the history is scattered.

Not a problem at all! It would be great to preserve this, or at least have you handle the commits.

How about this: I'll open a new PR with just the SAMS half of things, absent the modifications elsewhere? You can follow the structure by adding the repex side of things in a separate PR. This was still a useful exercise for me to figure out what the structure should look like, however.

jchodera:
Closed in favor of #203

andrrizzi:
Sounds great @jchodera , thanks so much! 

> especially if there are multiple files involved 

If you figured out a way, I'd love to learn how to do that!

PR #205, opened 2017-05-25
System-less getstate

Lnaden:
Allow __getstate__ to accept kwargs so we can customize the state objects returned from it
This generally is an optimization thing which should not be used carte blanche

Only option this adds is for `ThermodynamicState` to accept `skip_system` as a bool and returns `None` for the system object. This can be used in later optimizations to other code (like YANK) 

e.g. speed up caching many, many states of compatible states by not re-serializing the system every time.

Tagging @andrrizzi for review

### DIFF
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -350,7 +350,7 @@ class AlchemicalState(object):
     def __str__(self):
         return str(self._parameters)
 
-    def __getstate__(self):
+    def __getstate__(self, **kwargs):
         """Return a dictionary representation of the state."""
         serialization = dict(parameters={}, alchemical_variables={})
 
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -350,7 +350,7 @@ class AlchemicalState(object):
     def __str__(self):
         return str(self._parameters)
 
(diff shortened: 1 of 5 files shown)


andrrizzi:
Sorry! I don't I explained myself well. I didn't mean that all `__getstate__` methods should have `**kwargs`. Only `utils.serialize()` should, and `ThermodynamicState.getstate()` should have an explicit `skip_system` argument, but I'd leave everything else untouched.

Lnaden:
Sorry, I forgot that if no kwargs are supplied to the top function (`serialize`), then none are passed to the subfunction `__getstate__`, so is only thrown if you try to pass kwargs to serializing objects without `**kwargs` as part of get state. Fixing

Lnaden:
Okay, fixed it. I left the calls in `ThermodynamicState` and the `CompoundState` since it could aslo hold a `ThermodynamicState`

andrrizzi:
> and the `CompoundState`

Nice catch! Can you make the argument explicit in `ThermodynamicState` (it's fine in `CompoundState`)? I don't like `**kwargs` when they are not forwarded to any other function because they fail silently with typos. Feel free to merge otherwise!

PR #206, opened 2017-05-25
Rename AlchemicalFactory to AbsoluteAlchemicalFactory

### LINKED ISSUE #204: Rename AlchemicalFactory to AbsoluteAlchemicalFactory
We'll soon have a `RelativeAlchemicalFactory` or `HybridAlchemicalFactory` too, so it would be useful to make this change sooner rather than later.

andrrizzi:
Fix #204.

@jchodera is this what you had in mind for the API warning (second commit)?

### DIFF
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -12,7 +12,7 @@ DESCRIPTION
 This module contains enumerative factories for generating alchemically-modified System objects
 usable for the calculation of free energy differences of hydration or ligand binding.
 
-* `AlchemicalFactory` uses fused elecrostatic and steric alchemical modifications.
+* `AbsoluteAlchemicalFactory` uses fused elecrostatic and steric alchemical modifications.
 
 """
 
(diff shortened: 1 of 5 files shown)


openmmtools/alchemy.py
```diff
@@ -889,6 +889,10 @@ def get_energy_components(cls, alchemical_system, alchemical_state, positions, p
     def replace_reaction_field(self, reference_system):
         """Replace reaction-field electrostatics with Custom*Force terms to ensure c_rf = 0.
 
+        .. warning:: Unstable API.
+            This method is still experimental. It could be moved in some
```
jchodera: Change "moved in" to "moved to"

jchodera:
Yes!

Maybe we should use this opportunity to change `alchemy` into a subpackage as well?

andrrizzi:
> Change "moved in" to "moved to"

Will do! Thanks for catching.

> Maybe we should use this opportunity to change `alchemy` into a subpackage as well?

Can we postpone this until we'll incorporate the other factories? I'd like to merge everything/release everything and restart the calculations before the weekend.

jchodera:
Sure, looks good otherwise.

andrrizzi:
Thanks! Merging.

PR #207, opened 2017-05-25
Release 0.10.0

andrrizzi:
Prepare for release 0.9.5.


jchodera:
Should this be 0.10.0 since we broke `AlchemicalFactory`?

andrrizzi:
Good point! I'll change it.

Issue #217, opened 2017-06-09
Add option to AbsoluteAlchemicalFactory to disable long-range correction in CustomNonbondedForce when alchemical switching will be used


jchodera:
In cases where we want to alchemically switch `lambda_sterics` by updating its value every timestep, a huge (100x) slowdown is noted because the dispersion correction is numerically integrated (on the CPU) each time `lambda_sterics` is updated. To avoid this, we should add an option to `AbsoluteAlchemicalFactory` to disable the long-range correction in `CustomNonbondedForce` to allow efficient integration in these cases.

cc: https://github.com/pandegroup/openmm/issues/1832
cc: https://github.com/choderalab/openmmtools/issues/201#issuecomment-306965778

andrrizzi:
This was added in #218 .

PR #218, opened 2017-06-09
Add option to AbsoluteAlchemicalFactory to disable long-range dispersion correction for alchemical region

jchodera:
Addresses #217 

This PR adds the option `disable_alchemical_dispersion_correction` to `AbsoluteAlchemicalFactory`. If set to `True`, it will override the dispersion correction for the alchemical region, forcing it off. This is important when speed is desired for nonequilibrium switching where `lambda_sterics` is changing frequently; otherwise, if the long-range dispersion correction is active, the dispersion correction will be recomputed on the CPU every time `lambda_sterics` is altered, incurring a ~0.5 s overhead/timestep, potentially slowing things down by 100x.

Example:
```python
new_factory = AbsoluteAlchemicalFactory(consistent_exceptions=False, disable_alchemical_dispersion_correction=True)
reference_system = testsystems.WaterBox().system
alchemical_region = AlchemicalRegion(alchemical_atoms=[0, 1, 2])
alchemical_system = new_factory.create_alchemical_system(reference_system, alchemical_region)
```

An overlap test is added to ensure that a water box and LJ fluid retain good overlap with their unmodified systems when this option is in use. We can't test energy differences because disabling the dispersion correction for some atoms introduces a non-constant additive term.

### DIFF
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -673,10 +673,16 @@ class AbsoluteAlchemicalFactory(object):
         'direct-space' only models the direct space contribution
         'coulomb' includes switched Coulomb interaction
     alchemical_rf_treatment : str, optional, default = 'switched'
-         Controls how alchemical region electrostatics are treated when RF is used
-         Options are ['switched', 'shifted']
-         'switched' sets c_rf = 0 for all reaction-field interactions and ensures continuity with a switch
-         'shifted' retains c_rf != 0 but can give erroneous results for hydration free energies
+        Controls how alchemical region electrostatics are treated when RF is used
+        Options are ['switched', 'shifted']
+        'switched' sets c_rf = 0 for all reaction-field interactions and ensures continuity with a switch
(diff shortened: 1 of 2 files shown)


jchodera:
> Do you think this is something we should disable in yank too since we keep a single context and keep applying the ThermodynamicState with a different lambda_sterics at each iteration?

I've been trying to think about whether this is a good idea or not, but it's not completely clear which setting is optimal. Setting this option to `True` will reduce overlap, but I'm not sure how much---it will depend on the size of the alchemical region. On the other hand, since we take ~500 steps, the overhead is only ~0.5 seconds added onto the integration time. In some cases, that's huge; in others, its minimal. 

Maybe we can make it an option?

We might also see if we can lobby to speed up the recomputation in OpenMM by having it migrate to a GPU kernel.

jchodera:
Looks like we're bumping into the testing time limit.

jchodera:
I've removed the expensive waterbox test. The LJ fluid should be sufficient.

PR #223, opened 2017-06-14
New test system: the Drew-Dickenson B-DNA dodecamer

gregoryross:
The test system is a palindromic sequence that is useful for studying ion-DNA interactions. This will be used as a `saltswap` test system.

**Notes from the class doctstrings**
Drew-Dickerson B-DNA dodecamer (CGCGAATTCGCG) in explicit solvent. Structure taken from the RCSB PDB accession code 4C64 (1.3 Angstrom resolution). Solvated using the TIP3P water model in a cuboidal box with at least a 10 Angstrom clearance between the edge of the box and the DNA. The DNA is parametrized with the AMBER 14SB forcefield. The system has a total charge of -22 as no neutralizing counterions have been included.


jchodera:
Thanks for adding this!

We need information on the provenance of these files. Can you add a `README.md` to the `ddd-dodecamer/` directory containing the `{prmtop,inpcrd}` files with complete provenance information? If you prepared this yourself, the scripts you used to do so should also be checked in. You might add information about the provenance to the docstring as well.

openmmtools/testsystems.py
```diff
@@ -3473,6 +3474,82 @@ def __init__(self, constraints=app.HBonds, rigid_water=True, nonbondedCutoff=DEF
 
         self.system, self.positions = system, positions
 
+# =============================================================================================
+# Drew-Dickerson B-DNA dodecamer in explicit solvent
+# =============================================================================================
+
+class DDDExplicit(TestSystem):
```
jchodera: How about we call it `DNADodecamerExplicit` or `DrewDickersonDNADodecamerExplicit`?

jchodera:
If this was the protocol you followed with LEaP, you can just add that to the `README.md` and dump your input scripts into the directory.
> Drew-Dickerson B-DNA dodecamer (CGCGAATTCGCG) in explicit solvent. Structure taken from the RCSB PDB accession code
>  +    4C64 (1.3 Angstrom resolution). Solvated using the TIP3P water model in a cuboidal box with at least a 10 Angstrom
>  +    clearance between the edge of the box and the DNA. The DNA is parametrized with the AMBER 14SB forcefield. The
>  +    system has a total charge of -22 as no neutralizing counterions have been included.

gregoryross:
@jchodera, I've renamed the test system `DNADodecamerExplicit` as per your suggestion and added the preparation files, description of preparation, and provence information. Please let me know if your happy for me to merge this (once the tests pass).

openmmtools/testsystems.py
```diff
@@ -3473,6 +3474,82 @@ def __init__(self, constraints=app.HBonds, rigid_water=True, nonbondedCutoff=DEF
...
+    """
+
+    def __init__(self, constraints=app.HBonds, rigid_water=True, nonbondedCutoff=DEFAULT_CUTOFF_DISTANCE,
+                 use_dispersion_correction=True, nonbondedMethod=app.PME, hydrogenMass=None,
+                 switch_width=DEFAULT_SWITCH_WIDTH, ewaldErrorTolerance=DEFAULT_EWALD_ERROR_TOLERANCE, **kwargs):
+
+        TestSystem.__init__(self, **kwargs)
+
+        try:
+            from parmed.amber import AmberParm
```
jchodera: Wait, why does this require `parmed` instead of the `simtk.openmm.app.PrmtopFile` loader?

jchodera:
Also, is the system ready to simulate, or does it require further minimization before use?

openmmtools/testsystems.py
gregoryross: I copied the DHFR test system setup [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/testsystems.py#L3452-L3453). I realize this is not necessary, but I wanted to maintain consistency.

gregoryross:
The system has not been minimized or thermalized. Is that necessary though? Minimization and thermalization should be part of any user's pipeline before any production simulation.

openmmtools/testsystems.py
gregoryross: I'm happy to change that though.

jchodera:
Wow, I don't remember having done that. I'll create an issue to fix that in a separate PR.

jchodera:
> The system has not been minimized or thermalized. Is that necessary though? Minimization and thermalization should be part of any user's pipeline before any production simulation.

Generally yes, but if the system hasn't been minimized at all, it could lead some of the tests (which see if they can simulate the system for a bit) to blow up. Some pre-minimization also leads to a reduction of the time required for every subsequent use to minimize under the new conditions.

openmmtools/testsystems.py
jchodera: The system has not been minimized or thermalized. Is that necessary though? Minimization and thermalization should be part of any user's pipeline before any production simulation.

jchodera: Wow, I don't remember having done that. I'll add an issue to fix the parmed dependencies in a separate PR.

jchodera:
I think the failures are unrelated, @gregoryross, so you should be good to merge.

gregoryross:
@jchodera, I've removed the Parmed dependency and up-dated the DNA forcefield I used for parametrization. I also tried to figure out a way to encapsulate the system after I had minimized and briefly equilibrated it. However,  as openmm doesn't have an up-to-date DNA forcefield, I had to pass the equilibrated structure through tleap again so that I could upload it here. Unfortunately, this (cutting a very long story short) screwed up the solvation of the dodecamer, so I'll stick with the unminimized structure for now. 

With regards to your concern that minimization will be slow, I found that minimization of the structure only took about a minute or so on my desktop CPU. I even didn't observe the structure blowing-up even without minimization, with I think is due to the high quality of the structure. As long as this is okay with you, I'll merge after the latest test finish.

jchodera:
OK, this sounds fine then!

For future reference, it is relatively straightforward to write out a minimized PDB file using `simtk.openmm.app.PDBFile.writeFile()` and just read in this PDB file instead of the original inpcrd file. This is fine for now, though, but later, saving a minute every time the system is used could add up to a lot of time!

gregoryross:
> For future reference, it is relatively straightforward to write out a minimized PDB file using simtk.openmm.app.PDBFile.writeFile() and just read in this PDB file instead of the original inpcrd file.
Ah, good suggestion! Just done this and will merge after tests pass.

PR #232, opened 2017-06-30
Compress serialized state strings

Lnaden:
This PR makes all serialized systems compressed strings under a default encoding to save space when saved on disk, and space in memory. Small amount of overhead added (< 5 seconds) for compression action.

@andrrizzi did I miss anyplace to check for the compression?

Progress towards choderalab/yank#702

### DIFF
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -952,14 +952,14 @@ class ThermodynamicState(object):
         self._pressure = serialization['pressure']
 
         serialized_system = serialization['standard_system']
+        # Decompress system
+        serialized_system = zlib.decompress(serialized_system).decode(self._ENCODING)
         self._standard_system_hash = serialized_system.__hash__()
 
         # Check first if we have already the system in the cache.
         try:
             self._standard_system = self._standard_system_cache[self._standard_system_hash]
         except KeyError:
-            # Decompress system
-            serialized_system = zlib.decompress(serialized_system).decode(self._ENCODING)
             system = openmm.XmlSerializer.deserialize(serialized_system)
             self._standard_system_cache[self._standard_system_hash] = system
             self._standard_system = system
@@ -1142,8 +1142,6 @@ class ThermodynamicState(object):
         """Standardize the system and return its hash."""
         cls._standardize_system(system)
         system_serialization = openmm.XmlSerializer.serialize(system)
-        # Compress
-        system_serialization = zlib.compress(system_serialization.encode(cls._ENCODING))
         return system_serialization.__hash__()
 
     # -------------------------------------------------------------------------
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -952,8 +952,14 @@ class ThermodynamicState(object):
         self._pressure = serialization['pressure']
 
         serialized_system = serialization['standard_system']
-        # Decompress system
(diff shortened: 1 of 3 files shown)


andrrizzi:
> did I miss anyplace to check for the compression?

Nope! I don't think so.

I would make the compression optional through a parameter `__getstate__(..., compressed=False)` though, and check the type in `__setstate__()` to decide whether to decompress or not.

Lnaden:
But if we always assume compression, there is no need to make it optional

andrrizzi:
> But if we always assume compression, there is no need to make it optional

Oh, now I got what you asked me before! Sorry, I misunderstand. I meant that we should always have `compress=True` in `Reporter.write_dictionary()` when we pass a `ThermodynamicState`.

Also, I don't think we need to compress the serialization to generate the standard system hash.

Lnaden:
> in compress=True in Reporter.write_dictionary()...

Ah, that is being handled in the YANK fix in choderalab/yank#707

> I don't think we need to compress the serialization to generate the standard system hash

We do, there are checks to see if the hashes match [here](https://github.com/choderalab/openmmtools/blob/compress/openmmtools/states.py?utf8=%E2%9C%93#L955), we would have to decompress the string first to check its hash in that case. And since we are really only going to handle compressed hashes, it made more sense for me to do this. It prevents divergent code

andrrizzi:
> we would have to decompress the string first to check its hash in that case

I realize that, but that simply slows down `_standardize_and_hash` instead of slowing down `__setstate__`, and I'd prefer that to be inverted. There are multiple functions relying on `_standardize_and_hash()`, and this implementation slows down all of them.

andrrizzi:
Damn. I just realized that if we don't allow the `zlib` compression to be optional, the SMIRNOFF data I generated won't be able to be analyzed by newer versions of yank. Can we make the compression optional through a `compress` argument to `__getstate__` and handle automatically both compressed/raw strings in `__setstate__`?

Lnaden:
Well.. I can, but I'd really prefer not to. What exactly do you need the decompressed versions for anyways?

andrrizzi:
All the smirnoff data I have generated have stored an uncompressed serialized state, so if we only support compressed serialization I won't be able to analyze the data with newer versions of `openmmtools`.

Lnaden:
Would it be so bad to pin to a max version of openmm tools for the older things, or even just add a compression line to your script code for the older things?

I can still put it in, but I'd prefer not since it adds a fair bit of ambiguity in the `__setstate__` calls since we don't know for sure if the string is compressed

andrrizzi:
> Would it be so bad to pin to a max version of openmm tools for the older things ... I'd prefer not since it adds a fair bit of ambiguity in the `__setstate__`

We could do that (the second solution would not work as this is a problem with yank analyze). However, in `__setstate__` the only thing that needs to be done is to check if the serialized string `isinstance` of `bytes` or `str`, and I don't think that trying to avoid this if...else statement is a good enough reason to break backwards compatibility, and pin the maximum version of `openmmtools` in previous `yank` recipes.

Lnaden:
@andrrizzi is this what you wanted for the backwards compatibility?

andrrizzi:
Yeah, this looks great, thanks! Feel free to merge when ready.

Lnaden:
Will do if the tests all pass

Issue #234, opened 2017-06-30
Implement automatic sizing of ContextCache


andrrizzi:
Right now `ContextCache.capacity` needs to be configured by the application to balance performance and avoid segfaults. @jchodera suggests in choderalab/yank#705 that it would be nice to have `ContextCache` figuring out the optimal size by keep adding `Context` until an out of memory error is found.

PR #235, opened 2017-06-30
ContextCache bugfix and optimization

### LINKED ISSUE #231: Context cache failure to access `set*`
Unfortunately, I don't have a minimal example yet, but I did want to note that a few days ago, as I was trying to use the `ContextCache` with a `celery` worker, I ran into a strange bug. Using the `AlchemicalNonequilibriumLangevinIntegrator`, whenever it tried to recover a context and integrator, I got an exception that there was no attribute `set*` (including `setTemperature`, etc.).

 After some debugging, I noticed that the `AlchemicalNonequilibriumLangevinIntegrator` was becoming a regular `CustomIntegrator` somehow, but I'm not quite sure how or why.

This isn't super high priority, since I have a workaround for now, but since the `ContextCache` API is really awesome and convenient, it would be great to figure out what's going on.

### LINKED ISSUE #233: ContextCache should prefer non-default Contexts when integrator is not specified
In some ill cases, `ContextCache` keep alive the default `Context` that is created when no integrator is requested and there are no other `Context` to recycle (see choderalab/yank#705). `ContextCache` should prefer non-default `Context` when it has the choice.

andrrizzi:
Should be ready for review.

In this PR:
- Fix #231: Context cache failure to access `set*`.
  - The problem was in the creation of the integrator hashes. The integrator was copied for standardization, but the interface was not restored, so things like temperature were not set correctly (tagging @pgrinaway ).
- Fix #233: ContextCache should prefer non-default Contexts when integrator is not specified.
  - When no integrator is required and there are multiple `Context`s with compatible thermodynamic state, `ContextCache` now always picks a `Context` with the non-default integrator. This should avoid keeping around an extra `Context` that is not used.
- The default global context cache now has no capacity limit. Until #234 is implemented, the user will have to manually configure it.

### DIFF
--- a/openmmtools/cache.py
+++ b/openmmtools/cache.py
@@ -552,7 +552,7 @@ class DummyContextCache(object):
 # GLOBAL CONTEXT CACHE
 # =============================================================================
 
-global_context_cache = ContextCache(capacity=3, time_to_live=50)
+global_context_cache = ContextCache(capacity=None, time_to_live=None)
 
 
 # =============================================================================
--- a/openmmtools/cache.py
+++ b/openmmtools/cache.py
@@ -114,12 +114,26 @@ class LRUCache(object):
     @time_to_live.setter
     def time_to_live(self, new_time_to_live):
         # Update entries only if we are changing the ttl.
-        ttl_diff = new_time_to_live - self._ttl
-        if ttl_diff == 0:
+        if new_time_to_live == self._ttl:
             return
+
+        # Update expiration of cache entries.
         for entry in self._data.values():
-            entry.expiration += ttl_diff
-        self._remove_expired()
+            # If there was no time to live before, just let entries
+            # expire in new_time_to_live accesses
+            if self._ttl is None:
+                entry.expiration = self._n_access + new_time_to_live
+            # If we don't want expiration anymore, delete the field.
+            # This way we save memory in case there are a lot of entries.
+            elif new_time_to_live is None:
+                del entry.expiration
+            # Otherwise just add/subtract the difference.
+            else:
+                entry.expiration += new_time_to_live - self._ttl
+
+        # Purge cache only if there is a time to live.
+        if new_time_to_live is not None:
+            self._remove_expired()
         self._ttl = new_time_to_live
 
     def empty(self):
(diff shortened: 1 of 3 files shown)


andrrizzi:
Sorry, I just realized that my previous commit doesn't fix completely #231, but I think I understand now what's happening. I'll update the issue and push a fix here.

andrrizzi:
That's probably a test for the `distributed` module when it'll be implemented. I'm not sure how to add it right now. I did made some local tests though, and everything seems to work.

jchodera:
That makes sense!

openmmtools/cache.py
```diff
@@ -459,7 +481,13 @@ def _generate_state_id(thermodynamic_state):
...
         # We take advantage of the cached _standard_system_hash property
         # to generate a compatible hash for the thermodynamic state.
-        return str(thermodynamic_state._standard_system_hash)
+        return thermodynamic_state._standard_system_hash
+
+    @classmethod
+    def _generate_integrator_id(cls, integrator):
+        """Return a unique key for the given Integrator."""
+        standard_integrator = cls._standardize_integrator(integrator)
+        return openmm.XmlSerializer.serialize(standard_integrator).__hash__()
```
Lnaden: I thought the `__hash__` here was non-deterministic. I see later on you use the `adler32` to hash out the integrator, but is it compatible here and/or should you be using the `adler32` hash here as well?

andrrizzi: So I think the situation here is the same as for the states classes. This hash is only kept by the `ContextCache` object, which is unique for each process, and it's not related to the hash computed by `RestorableIntegrator`. So the situation is a little different from `RestorableIntegrator` where the hash becomes part of the state of the integrator (i.e. it's part of its serialization) and can be send around to multiple processes.

If we do end up converting everything to deterministic algorithm, we should use a 64-bit hash. I've used `adler32` in `RestorableIntegrator` only because we need to hash strings of few tens of characters (the names of the integrators).

Lnaden: Great! I figured it was something like that, but I wanted to double check. Looks good then, switching review to approve

andrrizzi: Awesome, thanks!

Issue #238, opened 2017-07-06
Speed up slow tests


Lnaden:
OpenMMTools is getting to be such a large repo that the tests are taking non-trivial lengths of time. Couple that with several tests which are individually slow, it makes the CI painfully slow. I have ID'd some exceptionally slow tests which I would like their respective authors to take a look at and try to speed up. It might not be possible to, but still take a look and see if you can get it to fail/pass faster for the good of everyone.

@maxentile There is the one that is Greg's, but you might be best to look at it

| Test | Wall clock on CI | Author
|:----|:-------:|----:|
|[`test_reset_protocol`](https://github.com/choderalab/openmmtools/blame/b2b528239cef2807fcdb0f1429cad593b905a6ed/openmmtools/tests/test_integrators.py#L379) | 163 s | @maxentile / Greg
|[`test_alchemical_langevin_integrator`](https://github.com/choderalab/openmmtools/blame/b2b528239cef2807fcdb0f1429cad593b905a6ed/openmmtools/tests/test_integrators.py#L726) | 825 s | @jchodera / @pgrinaway 
| [IdealGas `test_mcmc_expectations`](https://github.com/choderalab/openmmtools/blame/3711c4c03cda59f31ccb2f2eb7048f0e4969cd4f/openmmtools/tests/test_mcmc.py#L40) | 70 s | @andrrizzi

andrrizzi:
Thanks @Lnaden! I'll try to fix the ideal gas test, which is the same one on the roadmap for 0.12 (see #226 and #225). I actually imported it as it was from the `openmmmcmc` repo.

maxentile:
Taking a look now! I think the fastest option for the one I'm tagged on is to keep the same logic in `test_reset_protocol`, but replace the `AlchemicalWaterbox` system with something with fewer atoms. Will take a stab at this shortly...

Lnaden:
Those two fixes alone appear to have shaved over 10 min from the tests! I'm going to close this issue for now and we can readdress other slow tests if we have more specific problems. We can tweak individual tests as we make other changes going forward.

PR #240, opened 2017-07-06
Use less expensive testsystem in a slow Travis test

maxentile:
* Adds an `AlchemicalAlanineDipeptide` testsystem
* Replaces `AlchemicalWaterbox` with `AlchemicalAlanineDipeptide` in a slow test

Partially addresses https://github.com/choderalab/openmmtools/issues/238


Lnaden:
I removed the Barostat which was causing the modified test to fail. @maxentile can you confirm that the Barostat was not needed for this test?

PR #241, opened 2017-07-06
Python 2 compatibility fix

andrrizzi:
In Python 2 decoding a `str` object returns a `unicode` object that OpenMM can't deserialize. This fixes the problem and adds a test.

### DIFF
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -15,6 +15,7 @@ Classes that represent a portion of the state of an OpenMM context.
 # =============================================================================
 
 import abc
+import sys
 import copy
 import zlib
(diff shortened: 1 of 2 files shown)


Lnaden:
Why did this not come up in the tests from when I made this change?

andrrizzi:
In the tests, `ThermodynamicState` was finding the system in the cache, so it didn't deserialize it with OpenMM. I already extended the tests to cover also this case.

Lnaden:
Okay, so will we need to cut an emergency 0.11.2?

andrrizzi:
Yep!

PR #242, opened 2017-07-07
Speed up alchemical_langevin_integrator tests

jchodera:
Addresses #238.

This speeds up `run_alchemical_langevin_integrator` from 212s to 96s.


Lnaden:
Thanks!

PR #253, opened 2017-07-12
Add forces module.

### LINKED ISSUE #251: Where to put replace_reaction_field?
Currently it's an instance method of `AbsoluteAlchemicalFactory`. The relative factory will surely want to use it too, so we should probably move it as a function at the `alchemy` module level at least. Are there any other plans for that feature?

andrrizzi:
Fix #251 .

In this PR:
- [x] Add `openmmtools.forces` and `openmmtools.forcefactories` modules.
- [x] Add `UnshiftedReactionFieldForce` class that encapsulate the energy expression of reaction field with `c_rf=0`.
- [x] Move `replace_reaction_field` factory function from `alchemy.AbsoluteAlchemicalFactory` to `forcefactories`.
- [x] Adapt `alchemy` RF tests.

### DIFF
--- a/docs/forces.rst
+++ b/docs/forces.rst
@@ -1,0 +1,22 @@
+.. _forces::
+
+Forces
+======
+
+The module :mod:`openmmtools.forces` implements custom forces that are not natively found in OpenMM.
+
+ - :class:`UnshiftedReactionFieldForce`: A `CustomNonbondedForce <http://docs.openmm.org/7.1.0/api-python/generated/simtk.openmm.openmm.CustomNonbondedForce.html>`_ implementing a reaction field variant with `c_rf` term set to zero and a switching function.
+ - :func:`find_nonbonded_force`: Find the first ``NonbondedForce`` in an OpenMM ``System``.
+ - :func:`iterate_nonbonded_forces`: Iterate over all the ``NonbondedForce``s in an OpenMM ``System``.
+
+Cache objects
+-------------
+
+.. currentmodule:: openmmtools.forces
+.. autosummary::
+    :nosignatures:
+    :toctree: api/generated/
+
+    UnshiftedReactionFieldForce
+    find_nonbonded_force
+    iterate_nonbonded_forces
--- a/docs/forces.rst
+++ b/docs/forces.rst
@@ -5,7 +5,7 @@ Forces
 
 The module :mod:`openmmtools.forces` implements custom forces that are not natively found in OpenMM.
 
- - :class:`UnshiftedReactionFieldForce`: A `CustomNonbondedForce <http://docs.openmm.org/7.1.0/api-python/generated/simtk.openmm.openmm.CustomNonbondedForce.html>`_ implementing a reaction field variant with `c_rf` term set to zero and a switching function.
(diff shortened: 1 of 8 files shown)


jchodera:
Can we separate `openmmtools.forces` (classes that are subclasses of `simtk.openmm.force`) and `openmm.forcefactories` (classes that conveniently create new forces for you)?

jchodera:
Or maybe `openmmtools.force.factories`?

andrrizzi:
Sure, I can do that! When you have time, can you elaborate a little on what you have in mind for the future though? Because I don't think I understand correctly, and I'm afraid that a separate factory module will end up super short. In particular, by
> classes that conveniently create new forces for you

do you mean "classes that create new systems" like `replace_reaction_field()` currently does, or that they create actual force objects?

In both cases, my hunch is that we're going to have very few of these. For example, in Yank we have a single factory function for all restraints (`restraints.create_restraint`), and in the future we could use a single function `replace_reaction_field(system, force_name='UnshiftedReactionField', **kwargs)` for all reaction field variants, so a whole module just for this methods may be an overkill.

jchodera:
Here's what I'm thinking we'll eventually add to `openmm.forces`:
* A bunch of replacements for the electrostatics treatment in `NonbondedForce`, including [Wolf summation](https://doi.org/10.1063%2F1.478738), [generalized reaction field](http://aip.scitation.org/doi/abs/10.1063/1.469273), and other [non-Ewald pairwise methods](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3428531/) (also reviewed [here](http://aip.scitation.org/doi/abs/10.1063/1.2206581))
* Harmonic restraints that use the `Topology` to add weak restraints to MDTraj DSL selected atoms
* Restraints that were previously found in YANK
* New GB models not currently in OpenMM
* Other interesting force types that may be useful in the future but aren't going to be incorporated into OpenMM

> For example, in Yank we have a single factory function for all restraints (restraints.create_restraint), and in the future we could use a single function replace_reaction_field(system, force_name='UnshiftedReactionField', **kwargs) for all reaction field variants, so a whole module just for these methods may be an overkill.

I think it's going to get crowded unless we have some sort of separation between force objects and force factories. I don't like the idea of mixing them in the same namespace. I think convenience factories that generate forces for you or modify systems should go into a separate namespace.

andrrizzi:
Ok! I'll create an extra `forcefactories.py` module.

andrrizzi:
Done! I've also expanded the `alchemy` tests to be able to do test the energy components of RF-replaced systems.

Should be ready for review.

jchodera:
Looks good to me now! I'd just have @Lnaden review and add the minimal `.rst` files so that the API docs show up in the sphinx documentation.

andrrizzi:
I can add the `rst` files, no problem!

andrrizzi:
Done! @Lnaden could you double-check that I didn't messed up something?

docs/forces.rst
```diff
@@ -0,0 +1,22 @@
+.. _forces::
+
+Forces
+======
+
+The module :mod:`openmmtools.forces` implements custom forces that are not natively found in OpenMM.
+
+ - :class:`UnshiftedReactionFieldForce`: A `CustomNonbondedForce <http://docs.openmm.org/7.1.0/api-python/generated/simtk.openmm.openmm.CustomNonbondedForce.html>`_ implementing a reaction field variant with `c_rf` term set to zero and a switching function.
```
Lnaden: Can we add a "why would I need/use this here?"

andrrizzi: Will do!

andrrizzi:
Done!

PR #254, opened 2017-07-23
Add option for switched/shifted reaction field

jchodera:
We're having some trouble with @maxentile with the unshifted reaction field, so I've added an optional argument `shifted` that can be set to `True` to `replace_reaction_field` so @maxentile can use it to simply ensure continuity at the cutoff for standard OpenMM reaction field.


jchodera:
@maxentile : You can use this branch to test, but I'll add some additional tests with the switch on before we merge.

jchodera:
@andrrizzi: I don't think this fixes the hydration free energy issues. I'm still working on that problem!

PR #255, opened 2017-07-25
Add function to apply harmonic restrain to atoms

andrrizzi:
See title! Ready for review.

### DIFF
--- a/openmmtools/forcefactories.py
+++ b/openmmtools/forcefactories.py
@@ -16,6 +16,7 @@ Factories to manipulate OpenMM System forces.
 
 import copy
 
+import numpy as np
 from simtk import openmm, unit
 
 from openmmtools import forces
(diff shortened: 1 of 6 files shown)


openmmtools/forcefactories.py
```diff
@@ -74,6 +75,58 @@ def replace_reaction_field(reference_system, switch_width=1.0*unit.angstrom,
...
+
+    This modifies the ``ThermodynamicState`` object.
+
+    Parameters
+    ----------
+    thermodynamic_state : openmmtools.states.ThermodynamicState
+        The thermodynamic state with the system. This will be modified.
+    sampler_state : openmmtools.states.SamplerState
+        The sampler state with the positions.
+    mdtraj_topology : mdtraj.Topology
```
jchodera: Can you accept either mdtraj or openmm topology objects here?

openmmtools/forcefactories.py
```diff
@@ -74,6 +75,58 @@ def replace_reaction_field(reference_system, switch_width=1.0*unit.angstrom,
...
+        The topology.
+    atoms_dsl : str
+        The MDTraj DSL string for selecting the atoms to restrain.
+    sigma : simtk.unit.Quantity, optional
+        Controls the strength of the restrain. The smaller, the tighter
+        (units of distance, default is 3.0*angstrom).
+
+    """
+    K = thermodynamic_state.kT / sigma  # Spring constant.
```
jchodera: Shouldn't K = kT/sigma**-2?

Exp(-beta K x^2/2) = Exp(-x^2 / (2 sigma^2))

openmmtools/forcefactories.py
andrrizzi: Sure!

openmmtools/forcefactories.py
andrrizzi: Uh oh! Yes it should. I guess I didn't notice it in the original script. Fixing it!

andrrizzi:
Comments addressed! Ok if I merge this?

openmmtools/forcefactories.py
```diff
@@ -74,6 +76,64 @@ def replace_reaction_field(reference_system, switch_width=1.0*unit.angstrom,
...
+    # Create a CustomExternalForce to restrain all atoms.
+    restraint_force = openmm.CustomExternalForce('(K/2)*((x-x0)^2 + (y-y0)^2 + (z-z0)^2)')
+    restrained_atoms = mdtraj_topology.select(atoms_dsl).tolist()
+    # Adding the spring constant as a global parameter allows us to turn it off if desired
+    restraint_force.addGlobalParameter('K', K)
+    restraint_force.addPerParticleParameter('x0')
+    restraint_force.addPerParticleParameter('y0')
+    restraint_force.addPerParticleParameter('z0')
+    for index in restrained_atoms:
+        parameters = sampler_state.positions[index,:].value_in_unit_system(unit.md_unit_system)
```
Lnaden: Quick question, doesn't this restrain the atoms to the current position of the atoms, and not to another atom/location? Or is that the point?

Lnaden: i.e. the atoms will be restrained in absolute space, not in relative space, and only the atoms that are in the dsl.

andrrizzi: Yep! That's done on purpose. For the alchemical path initialization, the idea is to avoid large reorganizations of the structure by fixing in space the receptor's heavy atoms.

andrrizzi:
I'm going to add to this same PR the warning in `AbsoluteAlchemicalFactory` for reaction field.

openmmtools/forcefactories.py
Lnaden: Okay, so this is not really intended to be used for a full production, but more of a specialized tool?

andrrizzi: I guess it's intended to be used every time you want to restrain a molecule's conformation.

Lnaden: Hmm.. Okay, I can live with that, changing review.

andrrizzi:
Ok, thanks! I'll merge as soon as tests pass and proceed with the release.

PR #259, opened 2017-07-28
Virtual sites in alchemical factory and add ions in WaterBox

### LINKED ISSUE #258: AbsoluteAlchemicalFactory does not support virtual sites
We need to add virtual site support to [`AbsoluteAlchemicalFactory`](https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L774-L850), which currently omits them.

cc: https://github.com/choderalab/yank/issues/720#issuecomment-318222114

andrrizzi:
Ready for review!

- Fix #258 : add support for virtual sites in `AbsoluteAlchemicalFactory`.
- Add `positive_ion`, `negative_ion` and `ionic_strength` constructor arguments to `WaterBox`. This was necessary to create a TIP4P test with something that could be alchemically modified (the ions).

### DIFF
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -802,39 +802,31 @@ class AbsoluteAlchemicalFactory(object):
         timer = utils.Timer()
         timer.start('Create alchemically modified system')
 
-        # Build alchemical system to modify.
-        alchemical_system = openmm.System()
+        # Build alchemical system to modify. This copies particles, vsites,
+        # constraints, box vectors and all the forces. We'll later remove
+        # the forces that we remodel to be alchemically modified.
+        alchemical_system = copy.deepcopy(reference_system)
 
-        # Set periodic box vectors.
-        box_vectors = reference_system.getDefaultPeriodicBoxVectors()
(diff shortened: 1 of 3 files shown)


jchodera:
Awesome! This is much more future-proof! A quick read through looks good to me, but might be good to have @Lnaden read through in detail!

Lnaden:
I'll take a dive through it once I am done writing up the part of the website update I'm finishing off

andrrizzi:
> what was wrong about using the getVirtualSite method of the System object to copy over the virtual site?

I tried that first. The `VirtualSite` SWIG object you obtain by `reference_system.getVirtualSite()` belongs to the reference `System` and can't be assigned to others. You'd need to copy it before calling `alchemical_system.setVirtualSite()`, but OpenMM failed when I tried to `deepcopy` it. I don't think it's implemented.

Lnaden:
Ah, that makes sense then. Its probably coded similar to the integrator then once it gets assigned a context, its consumed for the purposes of copying to use elsewhere. Looks good otherwise

PR #263, opened 2017-07-28
Fix default collision_rate in LangevinIntegrator

### LINKED ISSUE #260: Mismatch in default/docstring for collision_rate in LangevinIntegrator
In `openmmtools.integrators.LangevinIntegrator` the actual default `collision_rate` is `1.0 / simtk.unit.picoseconds`, while the docstrings have the default listed as `91.0/simtk.unit.picoseconds`
https://github.com/choderalab/openmmtools/blob/master/openmmtools/integrators.py#L890-L971

sgill2:
The default value for `collision_rate` in `LangevinIntegrator` was different than in the docstring . This changes the default from `1.0 / simtk.unit.picoseconds` to `91.0 / simtk.unit.picoseconds` to match with the rest of the integrators. 
Fixes #260.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -945,7 +945,7 @@ class LangevinIntegrator(ThermostatedIntegrator):
 
     def __init__(self,
                  temperature=298.0 * simtk.unit.kelvin,
-                 collision_rate=91.0 / simtk.unit.picoseconds,
+                 collision_rate=1.0 / simtk.unit.picoseconds,
                  timestep=1.0 * simtk.unit.femtoseconds,
                  splitting="V R O R V",
                  constraint_tolerance=1e-8,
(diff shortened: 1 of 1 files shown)


jchodera:
> LGTM, this is where it should be changed, default 1 /ps is not really a good collision rate.

1/ps is actually the *optimal* collision rate according to [Matthews and Leimkuhler](https://pdfs.semanticscholar.org/7f19/be559b17e349de768527bbdc5263d9d24bda.pdf) and I recommend that be our default.

sgill2:
Hmm, so all child classes of `LangevinIntegrator` should be adjusted to have the 1/ps `collision_rate` as well? I can go change that in this PR if that's the case.

jchodera:
That would be great---thanks!

sgill2:
One of the travis builds timed out with the following message:
```
ThermodynamicState._find_thermostat() method. ... *** glibc detected *** /home/travis/miniconda/envs/test/bin/python: corrupted double-linked list: 0x0000000007015680 ***
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```
The rest of them passed, so maybe that one build could be restarted to see if it ends up passing. Otherwise I think this PR should be about done.

Lnaden:
That particular test has some stochastic failures on Travis, I'll reboot it

PR #264, opened 2017-07-30
Fix ideal gas failures

jchodera:
There have been a number of stochastic test failures for the ideal gas barostat test, so I've tried to update the test to be more reliable.

Instead of fixing the number of equilibration iterations, I've increased the total number of iterations and used `pymbar.timeseries.detectEquilibration()` to identify the production region. Unfortunately, it seems like I'm getting real test failures at this point. Locally, I see things like:
```
FAIL: Testing MCMC expectation for HarmonicOscillator
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/choderalab/openmmtools/openmmtools/openmmtools/tests/test_mcmc.py", line 155, in subtest_mcmc_expectation
    assert nsigma <= NSIGMA_CUTOFF, err_msg.format()
AssertionError: Potential energy expectation
observed    1.81022 +-    0.01639kT | expected    1.50000 | error    0.31022 +-    0.01639 (18.9 sigma) | t0     0 | g   1.0 | Neff    101.0
----------------------------------------------------------------------------
FAIL: Testing MCMC expectation for IdealGas
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/choderalab/openmmtools/openmmtools/openmmtools/tests/test_mcmc.py", line 178, in subtest_mcmc_expectation
    assert nsigma <= NSIGMA_CUTOFF, err_msg.format()
AssertionError: Volume expectation
observed 8664.27240 +-    6.54014kT | expected 8811.36331 | error -147.09091 +-    6.54014 (22.5 sigma) | t0    17 | g   1.1 | Neff     76.8
----------------------------------------------------------------------------
```
It would be useful if @andrrizzi @maxentile @Lnaden might be able to take a look at what could be going on here. We will want to sort these out before the YANK 1.0 release or Langevin paper submission.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -19,7 +19,7 @@ requirements:
     - openmm
     - parmed
     - mdtraj
-
+    
   run:
     - python
     - cython
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -288,7 +288,7 @@ class MCMCSampler(object):
         # Retrieve data.
         self.sampler_state.update_from_context(context)
 
-        timer.report_timing()
+        #timer.report_timing()
 
 
 # =============================================================================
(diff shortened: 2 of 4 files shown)


jchodera:
Looks to be working now!

jchodera:
> For the Timer, in the future, we could switch to a global timer that prints a summary of the collected timings.

Oh, sorry, I had temporarily commented that out, but it's probably best to leave the debug statements off for now until we get something like what you suggest working!

PR #265, opened 2017-08-09
Nose-Hoover chain thermostated integrator.

andysim:
Reworked version of the [pull request to the master OpenMM](https://github.com/pandegroup/openmm/pull/1865) repository, to be more consistent with the existing integrators in this repo, particularly the Andersen thermostat.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -38,8 +38,6 @@ import logging
 import re
 import zlib
 
-import simtk.unit
-
 import simtk.unit as unit
 import simtk.openmm as mm
 
@@ -130,9 +128,9 @@ class RestorableIntegrator(mm.CustomIntegrator,PrettyPrintableIntegrator):
 
     Parameters
     ----------
-    temperature : simtk.unit.Quantity
+    temperature : unit.Quantity
         The temperature of the integrator heat bath (temperature units).
-    timestep : simtk.unit.Quantity
+    timestep : unit.Quantity
         The timestep to pass to the CustomIntegrator constructor (time
         units)
 
@@ -245,9 +243,9 @@ class ThermostatedIntegrator(RestorableIntegrator):
 
     Parameters
     ----------
-    temperature : simtk.unit.Quantity
+    temperature : unit.Quantity
         The temperature of the integrator heat bath (temperature units).
-    timestep : simtk.unit.Quantity
+    timestep : unit.Quantity
         The timestep to pass to the CustomIntegrator constructor (time
         units).
 
@@ -307,7 +305,7 @@ class ThermostatedIntegrator(RestorableIntegrator):
 
         Returns
         -------
-        temperature : simtk.unit.Quantity
+        temperature : unit.Quantity
             The temperature of the heat bath in kelvins.
 
         """
@@ -320,7 +318,7 @@ class ThermostatedIntegrator(RestorableIntegrator):
 
         Parameters
         ----------
-        temperature : simtk.unit.Quantity
+        temperature : unit.Quantity
             The new temperature of the heat bath (temperature units).
 
         """
@@ -439,7 +437,7 @@ class MTSIntegrator(respa.MTSIntegrator):
     of the force group and the frequency (as a fraction of the outermost time step) at
     which to evaluate it.  For example:
 
-    >>> integrator = MTSIntegrator(4*simtk.unit.femtoseconds, [(0,1), (1,2), (2,8)])
+    >>> integrator = MTSIntegrator(4*unit.femtoseconds, [(0,1), (1,2), (2,8)])
 
     This specifies that the outermost time step is 4 fs, so each step of the integrator
     will advance time by that much.  It also says that force group 0 should be evaluated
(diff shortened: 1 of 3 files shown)


jchodera:
Awesome, thanks!

@andrrizzi @Lnaden : Can you review?

andysim:
Has anybody seen [this](https://travis-ci.org/choderalab/openmmtools/jobs/263149741#L2840) problem with py36 before?  It appears to be unrelated to the changes I made, but I can't be absolutely certain.

Lnaden:
Yeah, thats a test which stochastically fails on Travis-CI. I'll restart it

Lnaden:
I should also be able to get through the review for this today

andysim:
Excellent, thanks.  I don't actually have much experience with integrators and implemented this mostly for my own edification; if you have suggestions for default settings I'm more than happy to change them.  Just browsing the travis logs, it looks like the temp set/get is orders of magnitude slower than for the Andersen thermostat, which is strange.  The (probably too short to be significant) simulation timings look consistent between the different thermostats.  Regardless, I think the default collision frequency looks reasonable, but suspect the other parameters are probably overkill.  The `if` statements that codeclimate is complaining about are just to make the algorithm fall back to regular velocity verlet in the case that the chain length is zero.

openmmtools/integrators.py
```diff
@@ -671,6 +672,206 @@ def __init__(self, temperature=298 * simtk.unit.kelvin, collision_rate=91.0 / si
...
+        self.addGlobalVariable("ndf", 0)      # number of degrees of freedom
+        self.addGlobalVariable("bathKE", 0.0) # Thermostat bath kinetic energy
+        self.addGlobalVariable("bathPE", 0.0) # Thermostat bath potential energy
+        self.addGlobalVariable("KE2", 0.0)    # Twice the kinetic energy
+        self.addGlobalVariable("Q", Q)        # Thermostat particle "mass"
+        self.addGlobalVariable("scale", 1.0)
+        self.addGlobalVariable("aa", 0.0)
+        self.addGlobalVariable("wdt", 0.0)
+        for w in range(self.n_ys):
+            self.addGlobalVariable("w%d"%w, self.weights[w])
```
Lnaden: Small consistency thing, can you switch all your string formatters to `.format` over `%`?

openmmtools/integrators.py
```diff
@@ -671,6 +672,206 @@ def __init__(self, temperature=298 * simtk.unit.kelvin, collision_rate=91.0 / si
...
+    """
+
+
+    YSWeights = {
+        1 : [ 1.0000000000000000 ],
+        3 : [ 0.8289815435887510, -0.6579630871775020,  0.8289815435887510 ],
+        5 : [ 0.2967324292201065,  0.2967324292201065, -0.1869297168804260, 0.2967324292201065, 0.2967324292201065 ]
+    }
+
+    def __init__(self, temperature=298*simtk.unit.kelvin, collision_frequency=50/simtk.unit.picoseconds,
```
Lnaden: Can you change all the `simtk.unit` to just `unit`? I noticed at some point there was a double import at the top I can fix in a separate PR

andysim: I just removed the double import and all instances of `simkt.unit` in favor of `unit`, docstrings included.  If you didn't want me to make that many changes, please let me know and I can revert it (squashing the PR at the end is easy).

openmmtools/integrators.py
andysim: Done!

andysim:
Sorry, I forgot to port the test case over in the first version of the PR.  I just added a simple test that checks to make sure the total energy (system+bath) in conserved over 100 steps of dynamics.  I could make the test longer and also make sure the target temperature is achieved but, at ~10 seconds on my laptop, I figured this test is probably time consuming enough.

jchodera:
Thanks, @andysim! 

We're tracking own some sort of segfault issue with the openmm `dev` package we use for testing. Hope to get this sorted out today.

Lnaden:
> I just removed the double import and all instances of simkt.unit 

Thanks for doing that.

The rest of the code looks good though, I'll change to approval.

> tracking own some sort of segfault issue with the openmm 

I tracked down the issue and made changes to how our tests run in #275, this should stop the segfaults and allow the full battery to run. If those pass, we can probably get this merged today

Lnaden:
@andysim Can you merge master into your fork'd branch here? I forgot how git works for a moment so this test will not pass until you merge the fix I did to the test into your PR

andysim:
I'll `pull --rebase` and force a push back up to this branch.  Also, as a heads-up, I noticed that the doctest for ThermostatedIntegrator reports a FAIL in the Travis output.  That alone doesn't seem to constitute a test failure, but it's entirely possible that I caused a problem by removing the `simtk` from the example docstrings.

Lnaden:
Thanks for seeing that. It appears to only happen in the dev version of the build, so I wonder if its related to the other failures. However, the tests in the main channel all pass, including the doctests, so I think its okay. If it crops up again, we'll handle it then.

PR #267, opened 2017-08-11
Add new LangevinSplittingDynamicsMove

Lnaden:
Uses the custom `LangevinDynamics` integrator with optional splitting and metropolization strings.

Could use review by @maxentile (especially the test), and possibly @andrrizzi if possible (I know you are out)

### DIFF
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -982,7 +982,7 @@ class IntegratorMove(BaseIntegratorMove):
 
 
 # =============================================================================
-# LANGEVIN DYNAMICS MOVE
+# LANGEVIN DYNAMICS MOVES
 # =============================================================================
 
 class LangevinDynamicsMove(BaseIntegratorMove):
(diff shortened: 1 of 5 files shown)


maxentile:
Thanks, Levi! I'm taking a look now

openmmtools/mcmc.py
```diff
@@ -1119,6 +1119,158 @@ def _get_integrator(self, thermodynamic_state):
...
+                 measure_shadow_work=False, measure_heat=False, **kwargs):
+        super(LangevinSplittingDynamicsMove, self).__init__(n_steps=n_steps,
+                                                            reassign_velocities=reassign_velocities,
+                                                            timestep=timestep,
+                                                            collision_rate=collision_rate,
+                                                            **kwargs)
+        self.splitting = splitting
+        self.constraint_tolerance = constraint_tolerance
+        self.measure_shadow_work = measure_shadow_work
+        self.measure_heat = measure_heat
```
maxentile: Should the `collision_rate` also be set here (and in `__getstate__` and `__setstate__`)?

Lnaden: Because the class is actually a subclass of `LangevinDynamicsMove` and not of `BaseIntegratorMove`, calling the super class in `__init__`, `__setstate__` and `__getstate__` manipulates the `collision_rate` and `timestep` properties already.

maxentile: Ahh, got it! Thanks for clarifying

openmmtools/tests/test_mcmc.py
```diff
@@ -365,6 +366,20 @@ def test_metropolized_moves():
                                             'move for class {}'.format(move_class.__name__))
 
 
+def test_langevin_splitting_move():
+    """Test that the langevin splitting mcmc move works with different splittings"""
+    splittings = ["V R O R V", "V R R R O R R R V", "V { R O R } V"]
```
maxentile: Might replace the `"V { R O R } V"` with `"O { V R V } O"` or something, so that all shadow-work-generating steps are actually included in the Metropolis test?

Lnaden:
There were two options here: either create the subclass for the `LangevinSplittingDynamicsMove`, or refactor the existing `LangevinDynamicsMove`. I chose to go with the subclass to make a new move so as not to unexpectedly change existing code relying on the existing function. 

I think switching from the default OpenMM Langevin Integrator to the custom one @maxentile wrote, while although allows splitting and metropolization, may confuse users by the sudden change. Maybe in a future major revision this can change, but lets not do it suddenly.

Just wanted to document this on the issue instead of only in our internal slack. Also open to other arguments if people disagree with this one.

setup.py
```diff
@@ -14,7 +14,7 @@
 DOCLINES = __doc__.split("\n")
 
 ########################
-VERSION = "0.13.0"
+VERSION = "0.13.1"
```
andrrizzi: I don't think we have released 0.13.0 yet.

PR #269, opened 2017-08-12
Fix sphinx docs

jchodera:
* Fix Force API documentation, which contained misnamed "Cache objects" and repeated information
* Fix Cache API documentation, which repeated information 
* Fix Force Factories docs, which contained misnamed "Cache objects"


jchodera:
Apologies for the self-merge, but I need to update the online docs for my E-CAM talk on Tuesday!

PR #270, opened 2017-08-14
Attempt to stop segfaulting test  

Lnaden:
Changing how the integrator is referenced so garbage collection may do its thing properly.
This may also not fix anything at all, hard to tell

### DIFF
--- a/openmmtools/tests/test_states.py
+++ b/openmmtools/tests/test_states.py
@@ -549,7 +549,9 @@ class TestThermodynamicState(object):
         test_platforms = [test_platforms[i % len(test_platforms)]
                           for i in range(len(test_integrators))]
 
-        for (is_thermostated, integrator), platform in zip(test_integrators, test_platforms):
+        for platform in test_platforms:
+            # Pop the integrator to allow proper garbage collection so its not referenced anymore when we del it.
+            (is_thermostated, integrator) = test_integrators.pop()
             context = state.create_context(integrator, platform)
             assert platform is None or platform.getName() == context.getPlatform().getName()
             assert isinstance(integrator, context.getIntegrator().__class__)


Lnaden:
Ugh. So its not internal to the single test itself, I think its some other test not appropriately garbage collecting. I can reproduce a different, but I think related error on our local cluster by having the GPU in `exclusive mode`, this particular test `test_method_create_context`, throws a `No compatible CUDA device is available` error, despite doing internal garbage collecting, adding manual sleep timer in to ensure collection is done, and ensuring no outstanding `locals()` point to context or integrator objects. This also only happens in the `nosetest` of the whole test suite, which leads be to believe that its another test being unruly. Digging into it.

jchodera:
Are you sure we can't just avoid this by `del context` to clean up after ourselves at the end of a test segment that creates a `Context`?

Lnaden:
That is what I am going through now, but its not quite so easy. 

For instance, this first attempt popped integrator's out of a list each iteration since before they would be unbound in the list, pointed to with a `integrator` variable, delete the `integrator, context` variables, but would still cause the context and integrator to exist since the (now bound) integrator would still exist in the list, which references a context that it is bound to so not all refs to that context would be gone, blocking garbage collection.

Another example is this line `for c in [context, thermostated_context]` here: https://github.com/choderalab/openmmtools/blob/402fa279d6d7e4ffe8beb65ed5168ccd3a87b7f5/openmmtools/tests/test_states.py#L638

Where the loop creates a context variable `c` which is not `del`'d, so a context still exists at the end of the test.

tl;dr: Its a pain

Lnaden:
I'm going to close this for right now so I can keep pushing changes without triggering tests

Lnaden:
Adding some more info here, I think culprit is the `test_method_apply_to_context` method in `TestThermodynamicState` class. If that function is run before `test_method_create_context`, my local tests fail on the `CUDA` import (see linked OpenMM bug), but if it is run in reverse order, its fine. 

I know that Travis does not have CUDA support, so its not a perfect example, but the fact that there appears to be some kind of borked garbage collection in CUDA leads me to think there might be a related garbage collection problem on Travis which could result in the segfault.

jchodera:
Didn't you say you could see this issue locally? If so, how could it be due to Travis-specific anything?

Lnaden:
What I saw locally is a failure at the same point in the `nosetests`, but it is not identically the same failure. On the local cluster I see a "No Compatible CUDA device available", but not a segfault.

After doing a fair amount of systematic digging, I can replicate the local error outside of nosetests, but cannot create the segfault. Because I can't recreate the segfault, that leads me to think its something on travis-specific hardware (or their image). I opened the OpenMM issue after I drilled down to a minimal fail case, but acknowledge that what we are seeing on travis may not in fact be this issue, so my debugging continues.

I can tell you that even after I `del`'d every single object from the `test_method_apply_to_context` test at the end, the No CUDA devices error is thrown locally, which is then why I went further and found that bug in CUDA garbage collection, which again, can't happen on Travis since no CUDA platform is available and should not be detected.

Lnaden:
More updates:

I was not using the Dev version in my last tests. After switching to openmm-dev from omnia/label/dev channel, I can recreate this test locally every time, without the GPU.

The problem happens on `test_method_create_context` and running the tests manually without nose yields 2 backtraces:

From Python:
```
  File "single.py", line 6, in <module>
    TTS.test_method_create_context()
  File "/lila/home/nadenl/github/openmmtools/openmmtools/tests/test_states.py", line 571, in test_method_create_context
    for _integrator in ThermodynamicState._loop_over_integrators(inconsistent_integrator):
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/openmmtools-0.13.0-py3.5.egg/openmmtools/states.py", line 1167, in _loop_over_integrators
    _integrator = integrator.getIntegrator(integrator_id)
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 14968, in getIntegrator
    return self._integrators[index]
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 14842, in <lambda>
    __getattr__ = lambda self, name: _swig_getattr(self, CompoundIntegrator, name)
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 74, in _swig_getattr
    return _swig_getattr_nondynamic(self, class_type, name, 0)
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 69, in _swig_getattr_nondynamic
    return object.__getattr__(self, name)
AttributeError: type object 'object' has no attribute '__getattr__'
Segmentation fault (core dumped)
```
And from `GDB` to `backtrace` the segfault
```
(gdb) backtrace
#0  0x0000000000000041 in ?? ()
#1  0x00007fff9534b977 in _wrap_delete_VerletIntegrator (args=<optimized out>)
   from /home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/_openmm.cpython-35m-x86_64-linux-gnu.so
#2  0x00007ffff79770f9 in PyCFunction_Call (func=0x7fff957b0bd0, args=0x7ffff7ef8080, kwds=<optimized out>) at Objects/methodobject.c:109
#3  0x00007ffff792233a in PyObject_Call (func=func@entry=0x7fff957b0bd0, arg=arg@entry=0x7ffff7ef8080, kw=kw@entry=0x0) at Objects/abstract.c:2166
#4  0x00007ffff7923429 in PyObject_CallFunctionObjArgs (callable=0x7fff957b0bd0) at Objects/abstract.c:2446
#5  0x00007fff95348ef8 in SwigPyObject_dealloc (v=0x7fff7bcdec60)
   from /home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/_openmm.cpython-35m-x86_64-linux-gnu.so
#6  0x00007ffff79691b5 in dict_dealloc (mp=0x7fff82b00f88) at Objects/dictobject.c:1644
#7  0x00007ffff798a171 in subtype_dealloc (self=0x7fff7cc689e8) at Objects/typeobject.c:1194
#8  0x00007ffff7955d17 in list_dealloc (op=0x7fff82b00188) at Objects/listobject.c:326
#9  0x00007ffff79691b5 in dict_dealloc (mp=0x7fff7cc66cc8) at Objects/dictobject.c:1644
#10 0x00007ffff798a171 in subtype_dealloc (self=0x7fff7cc684e0) at Objects/typeobject.c:1194
#11 0x00007ffff7988186 in tupledealloc (op=0x7fff7c96d288) at Objects/tupleobject.c:236
#12 0x00007ffff7955d17 in list_dealloc (op=0x7fff82b000c8) at Objects/listobject.c:326
#13 0x00007ffff7969317 in free_keys_object (keys=0x1dd9df0) at Objects/dictobject.c:351
#14 dict_dealloc (mp=0x7fff82b00288) at Objects/dictobject.c:1652
#15 0x00007ffff79528df in frame_dealloc (f=0x2065f08) at Objects/frameobject.c:442
#16 0x00007ffff7a4f037 in tb_dealloc (tb=0x7fff7c8e6088) at Python/traceback.c:55
#17 0x00007ffff7a4f047 in tb_dealloc (tb=0x7fff7c8e60c8) at Python/traceback.c:54
#18 0x00007ffff7969e77 in insertdict (value=0x7ffff7d63130 <_Py_NoneStruct>, hash=-6852834400899615361, key=0x7ffff0484830, mp=0x7ffff7f66748) at Objects/dictobject.c:807
#19 PyDict_SetItem (op=0x7ffff7f66748, key=0x7ffff0484830, value=0x7ffff7d63130 <_Py_NoneStruct>) at Objects/dictobject.c:1232
#20 0x00007ffff796c528 in PyDict_SetItemString (v=0x7ffff7f66748, key=<optimized out>, item=0x7ffff7d63130 <_Py_NoneStruct>) at Objects/dictobject.c:2898
#21 0x00007ffff7a2d958 in PyImport_Cleanup () at Python/import.c:370
#22 0x00007ffff7a3fcd5 in Py_Finalize () at Python/pylifecycle.c:576
#23 0x00007ffff7a613c3 in Py_Main (argc=argc@entry=2, argv=argv@entry=0x602010) at Modules/main.c:798
#24 0x0000000000400b54 in main (argc=2, argv=<optimized out>) at ./Programs/python.c:65
```
The failures started happening after [This OpenMM PR](https://github.com/pandegroup/openmm/pull/1870) was merged.

jchodera:
Make sure to add that to https://github.com/pandegroup/openmm/issues/1875 since that's the only way it's likely to get fixed.

Lnaden:
I really don't think these are related bugs is the problem. I think I happened to stumble into a second bug trying to track down this one. I'm working on a minimal test case for this as well, but I need to do some more digging, the Python back trace helps, but still incomplete.

Lnaden:
In fact, I can almost assuredly say these are not the same bug since the one on that issue fails on a separate line than this issues based on the Python backtrace

Lnaden:
Okay, I have finally isolated this down to **two separate errors** being thrown, one is the instigating error which signals python to stop, *then* the segfault happens on final garbage collection.

I've stepped through the failure and here is what I have come to:
1. Deepcopy of a `CompoundIntegrator` destroys some SWIG methods (still isolating exact causes) which then causes the initial failure of the test
2. Generic garbage collection of CompoundIntegrator throws a segfault no mater what operations where performed on it (isolating conditions).

I will finish coming up with min test cases and post an issue to OpenMM shortly

Lnaden:
Okay, finally isolated and minimized the problems and posted to OpenMM issue tracker.

jchodera:
Thanks!

jchodera:
In the meantime, I'm testing if removing the 'dev' channel so we pick up production openmm allows Travis to pass. If so, I suggest we create separate Travis branches with and without dev OpenMM.

Lnaden:
That is probably a good idea. I can work on that this morning.

PR #284, opened 2017-09-06
Removed dynamic typing mechanism in CompoundState to fix pickling.

andrrizzi:
Should be ready for review as soon as tests pass.

I had to convert the `ThermodynamicState._standardize_system()` method from class to instance method to make the change. All state classes are now tested for pickling.


pgrinaway:
Hey, FYI It looks like this change fixed my problem--thanks!

andrrizzi:
Awesome, I'll merge this in then!

The only thing failing is codeclimate that is outrageously accusing me of duplicating code.

PR #290, opened 2017-09-12
Generalize restrain atoms to non-protein receptors.

### LINKED ISSUE #285: Pass to restrain_atoms a list of atom indices
I came across a case where creating a DSL selection string for the atoms I need to restrain is quite a pain. I'm thinking of making [`forcefactories.restrain_atoms()`](https://github.com/choderalab/openmmtools/blob/master/openmmtools/forcefactories.py#L89) a little more flexible by changing its signature from
```python
def restrain_atoms(..., topology, atoms_dsl,...):
```
to
```python
def restrain_atoms(..., restrained_atoms, ...):
```
where `restrained_atoms` is a simple list of atom indices. Any objections to this?

andrrizzi:
- Fix #285 : Allow passing a list of atom indices to `restrain_atoms()`.
- `forcefactories.restrain_atoms()` now computes the center of geometry of the atoms given by  `Context.getMolecules()` that match the restrained atoms.
- The center of geometry is moved to the origin only if the system is in NPT ensemble.
- An error is raised if the restrained atoms belong to multiple molecules.

### DIFF
--- a/openmmtools/forcefactories.py
+++ b/openmmtools/forcefactories.py
@@ -86,8 +86,19 @@ def replace_reaction_field(reference_system, switch_width=1.0*unit.angstrom,
 # RESTRAIN ATOMS
 # =============================================================================
 
-def restrain_atoms(thermodynamic_state, sampler_state, topology,
-                   atoms_dsl, sigma=3.0*unit.angstroms):
+def restrain_atoms_by_dsl(thermodynamic_state, sampler_state, topology, atoms_dsl, **kwargs):
+    # Make sure the topology is an MDTraj topology.
+    if isinstance(topology, mdtraj.Topology):
+        mdtraj_topology = topology
+    else:
+        mdtraj_topology = mdtraj.Topology.from_openmm(topology)
+
+    # Determine indices of the atoms to restrain.
+    restrained_atoms = mdtraj_topology.select(atoms_dsl).tolist()
+    restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, **kwargs)
+
+
+def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3.0*unit.angstroms):
     """Apply a soft harmonic restraint to the given atoms.
 
     This modifies the ``ThermodynamicState`` object.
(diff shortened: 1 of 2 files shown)


openmmtools/forcefactories.py
```diff
@@ -110,22 +121,41 @@ def restrain_atoms(thermodynamic_state, sampler_state, topology,
...
+        raise ValueError('No atoms to restrain.')
+
+    # We need to translate the restrained molecule to the origin
+    # to avoid MonteCarloBarostat rejections (see openmm#1854).
+    if thermodynamic_state.pressure is not None:
+        # First, determine all the molecule atoms. Reference platform is the cheapest to allocate?
+        reference_platform = openmm.Platform.getPlatformByName('Reference')
+        integrator = openmm.VerletIntegrator(1.0*unit.femtosecond)
+        context = openmm.Context(system, integrator, reference_platform)
+        molecules_atoms = context.getMolecules()
```
Lnaden: Do we have any custom/alchemical forces relying on a `CustomCompoundBondForce` or a `CustomBond` force to replicate a force which is NOT a molecule? e.g. if we have multiple restraints such as in the future, this could cause problems on the second pass of restraints

openmmtools/forcefactories.py
```diff
@@ -86,8 +86,19 @@ def replace_reaction_field(reference_system, switch_width=1.0*unit.angstrom,
 # RESTRAIN ATOMS
 # =============================================================================
 
-def restrain_atoms(thermodynamic_state, sampler_state, topology,
-                   atoms_dsl, sigma=3.0*unit.angstroms):
+def restrain_atoms_by_dsl(thermodynamic_state, sampler_state, topology, atoms_dsl, **kwargs):
```
Lnaden: We will need to make a change to YANK to use this correctly in the `pipeline` correct? If that issue is not already open, this is just to serve as a reminder.

openmmtools/forcefactories.py
```diff
@@ -110,22 +121,41 @@ def restrain_atoms(thermodynamic_state, sampler_state, topology,
...
+        # Make sure the atoms to restrain belong only to a single molecule.
+        molecules_atoms = [set(molecule_atoms) for molecule_atoms in molecules_atoms]
+        restrained_atoms_set = set(restrained_atoms)
+        restrained_molecule_atoms = None
+        for molecule_atoms in molecules_atoms:
+            if restrained_atoms_set.issubset(molecule_atoms):
+                # Convert set to list to use it as numpy array indices.
+                restrained_molecule_atoms = list(molecule_atoms)
+                break
+        if restrained_molecule_atoms is None:
```
Lnaden: Question: Doesn't this mean that ALL restraints in YANK will not work in explicit solvent? Since the ligand is restrained relative to the receptor, then its two separate molecules and so will always fail this criteria?

Lnaden:
After discussion with @andrrizzi, most of my comments are moot as this functionality is not being put in YANK. Looks fine otherwise

andrrizzi:
Thanks!

PR #301, opened 2017-10-12
Add logical and bitwise operators for math_eval

Lnaden:
Adds the ability to provide additional operators to the math_eval function. Only adds the `and`, `&`, `or`, and `|` operators.

This is a feature needed in an upcoming version of YANK


andrrizzi:
Can you maybe add a couple of test cases in `test_utils.py:test_math_eval` before merging?

Lnaden:
> I wonder if we do want to support also and and or because

They are both supported right now but both are set operations (so bitwise) because this is only suppose to be a set operator, not logical operator. But I can make that even more clear now if you want

Lnaden:
I liked your solution better, I had tried the recursion but only parsed the values, so it threw an error the first time I tried it. Your's seems to work

PR #304, opened 2017-11-14
Make libyaml optional

dwhswenson:
Not all of us have pyyaml installed with libyaml support. For those who don't, this uses the old pure python yaml as fallback. The (much faster) `CLoader` and `CDumper` are still preferred.

Uses the approach recommended at: https://pyyaml.org/wiki/PyYAMLDocumentation

### DIFF
--- a/openmmtools/storage/iodrivers.py
+++ b/openmmtools/storage/iodrivers.py
@@ -26,6 +26,11 @@ import numpy as np
 import netCDF4 as nc
 from sys import getsizeof
 
+try:
+    from yaml import CLoader as Loader, CDumper as Dumper
+except ImportError:
+    from yaml import Loader, Dumper
+
 from simtk import unit
 
 from ..utils import typename, quantity_from_string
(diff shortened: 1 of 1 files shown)


Lnaden:
How are you installing OpenMMTools such that `libyaml` does not get installed? I'm pretty sure the `conda` installation of `pyyaml` ships complied against `libyaml` and should have `CLoader/Dumper`

dwhswenson:
> How are you installing OpenMMTools such that `libyaml` does not get installed?

`pip install -e .` I never said I was a *normal* user. ;-) And before you ask, `pyyaml` was installed via MacPorts, because I'm a fossil and haven't moved to `brew` yet. I remain the guinea pig who doesn't use `conda` on my personal machines (for a little while longer, at least) -- specifically to catch edge cases like this.

Anyway, this is just intended to be a little fix for the oddballs like me. It won't affect most users.

The AppVeyor failure is weird -- this PR doesn't touch anything in the conda recipe or in the appveyor.yml, so I assume that was just bad luck.

Lnaden:
> And before you ask, pyyaml was installed via MacPorts 

Ah, that explains a fair amount! Good to see someone is still looking for non-standard corner cases (and fixing it!)

> The AppVeyor failure is weird

Yeah, AppVeyor is.... touchy. You are right that this PR should not touch AppVeyor. I'll double check to make sure its just an AppVeyor glitch and not something which needs addressed. Otherwise I think this PR is fine and I'll merge it in if I can't find anything odd

PR #311, opened 2017-11-25
Adapt CustomExternalForce expression to whether periodic boundary conditions are in use or not

### LINKED ISSUE #310: CustomExternalForce should use periodicdistance to avoid major artifacts when heavy atoms are restrained to fixed points in space
cc: https://github.com/pandegroup/openmm/issues/1854#issuecomment-346202857

In order to avoid major artifacts when using restraints to fixed points in space via `CustomExternalForce` (in [`restrain_atoms`](https://github.com/choderalab/openmmtools/blob/f50f0844b28106978973e6374613bc8d24a3c4b9/openmmtools/forcefactories.py#L101)), for periodic systems, we should replace
```python
restraint_force = mm.CustomExternalForce('(K/2)*((x-x0)^2 + (y-y0)^2 + (z-z0)^2)')
```
with
```python
restraint_force = mm.CustomExternalForce('(K/2)*periodicdistance(x, y, z, x0, y0, z0)^2')
```
as suggested in [this comment](https://github.com/pandegroup/openmm/issues/1854#issuecomment-346493353).

The relevant code is [here](https://github.com/choderalab/openmmtools/blob/f50f0844b28106978973e6374613bc8d24a3c4b9/openmmtools/forcefactories.py#L158).

jchodera:
Fixes #310

### DIFF
--- a/openmmtools/forcefactories.py
+++ b/openmmtools/forcefactories.py
@@ -155,7 +155,11 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
         sampler_state.positions -= centroid * distance_unit
 
     # Create a CustomExternalForce to restrain all atoms.
-    restraint_force = openmm.CustomExternalForce('(K/2)*((x-x0)^2 + (y-y0)^2 + (z-z0)^2)')
+    if thermodynamic_state.system.usesPeriodicBoundaryConditions():
+        energy_expression = '(K/2)*periodicdistance(x, y, z, x0, y0, z0)^2' # periodic distance
+    else:
+        energy_expression = '(K/2)*((x-x0)^2 + (y-y0)^2 + (z-z0)^2)' # non-periodic distance
+    restraint_force = openmm.CustomExternalForce(energy_expression)
     # Adding the spring constant as a global parameter allows us to turn it off if desired
     restraint_force.addGlobalParameter('K', K)
     restraint_force.addPerParticleParameter('x0')
--- a/openmmtools/forcefactories.py
+++ b/openmmtools/forcefactories.py
@@ -155,7 +155,7 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
         sampler_state.positions -= centroid * distance_unit
 
     # Create a CustomExternalForce to restrain all atoms.
-    if thermodynamic_state.system.usesPeriodicBoundaryConditions():
+    if system.usesPeriodicBoundaryConditions():
         energy_expression = '(K/2)*periodicdistance(x, y, z, x0, y0, z0)^2' # periodic distance
     else:
         energy_expression = '(K/2)*((x-x0)^2 + (y-y0)^2 + (z-z0)^2)' # non-periodic distance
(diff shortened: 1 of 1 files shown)


openmmtools/forcefactories.py
```diff
@@ -155,7 +155,11 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
         sampler_state.positions -= centroid * distance_unit
 
     # Create a CustomExternalForce to restrain all atoms.
-    restraint_force = openmm.CustomExternalForce('(K/2)*((x-x0)^2 + (y-y0)^2 + (z-z0)^2)')
+    if thermodynamic_state.system.usesPeriodicBoundaryConditions():
```
andrrizzi: You can directly use `thermodynamic_state.is_periodic`. This avoids making a copy of the internal standard `System`.

jchodera: Does the `system` property of `thermodynamic_state` really create a *copy* of the internal standard system? If so, this seems *super* confusing, since this would be correcft
```python
system = thermodynamic_state.system
system.addForce(force)
thermodynamic_state.system = system
```
while this would be garbage
```python
thermodynamic_state.system.addForce(force)
```
Do we have to have that behavior?

andrrizzi: There are two reasons for this
1. If the user modifies directly the system, it could get out of sync with the internal state of the `ThermodynamicState` object (for example by changing a barostat temperature), so only a copy is returned.
2. Internally the `ThermodynamicState` stores only a _standard_ state `System` that is shared across compatible thermodynamic states, not the actual system. If you remember, I had to implement this because creating 30-40 copies of the system object for all the alchemical states was a big performance burden.

The documentation [states it explicitly](https://github.com/choderalab/openmmtools/blob/master/openmmtools/states.py#L290).

jchodera: Oof, OK. I suppose the only thing that will make this better is pure Python versions of System.

andrrizzi: Yeah. A solution to this could be to add a `ThermodynamicState.add_force()` method to bypass having to make a copy. In the end, the `ThermodynamicState` class is a very fancy `System`. This is the same reason why I implemented the `is_periodic` property, for example.

jchodera: OK, I'll use `._is_periodic`.

jchodera: I'm just very worried about things like `ThermodynamicState` violating the "principle of minimum surprise" in API design. There's no way for someone to know that
```python
thermodynamic_state.system.addForce(force)
```
won't actually work the way it seems it should, for example, without having read and comprehended all the API documentation, and the mention that this is a copy is [really buried](http://openmmtools.readthedocs.io/en/latest/api/generated/openmmtools.states.ThermodynamicState.html#openmmtools.states.ThermodynamicState.system) in there.

andrrizzi: I agree. It can be confusing, although it's quite standard to return a copy of a mutable object. Keep in mind that the `System` object here is exposed only because there are occasions in which we need it, but I think that in the ideal design the user should not need to access it at all, and `ThermodynamicState` should provide the methods for everything (including `add_force()`).

The only alternative I can see is to create some sort of proxy object that callbacks the `ThermodynamicState` every time the system is modified, but at this point, we might as well make `ThermodynamicState` itself the proxy object.

PR #320, opened 2018-01-15
Exact PME electrostatics treatment and fast computation of reduced potentials

### LINKED ISSUE #319: Alchemically modify charges with updateParametersInContext
This is how I plan to add exact treatment of PME electrostatics in `alchemy`.
1. Allow the value `"exact"` for the argument `alchemical_pme_treatment` in the `AbsoluteAlchemicalFactory` constructor.
2. An exception is raised if `achemical_pme_treatment` is set to `"exact"`, but charges are decoupled and/or if `(softcore_beta, softcore_d, softcore_e, softcore_f) != (0, 1, 1, 1)`.
3. `AbsoluteAlchemicalFactory` creates a `CustomNonbondedForce` with empty interaction groups that permanently store the fully interacting charges of alchemical atoms.
4. `AlchemicalState` reads the original charges only once and cache them internally (charges are associated to the standard system hash).

The advantage of this approach is that `AlchemicalState` completely masks how the `lambda_electrostatics` parameter is implemented in the OpenMM `System`. Better ideas for how to keep in memory the original charges are welcomed.

andrrizzi:
Fix #319.

I've run some `timeit` tests, and reading the charges from the `CustomNonbondedForce` is actually pretty fast so I don't think implementing the caching system for the charges is worth it. With
```python
import timeit

def read_charges(system):
    original_charges_force = None
    for force in system.getForces():
        if (isinstance(force, openmm.CustomNonbondedForce) and
                    force.getEnergyFunction() == '0.0;'):
            original_charges_force = force

    # Set alchemical atoms charges.
    _, alchemical_atoms = original_charges_force.getInteractionGroupParameters(0)
    for atom_idx in alchemical_atoms:
        original_charge = original_charges_force.getParticleParameters(atom_idx)[0]

test_system = testsystems.SrcExplicit()
factory = AbsoluteAlchemicalFactory(alchemical_pme_treatment='exact')

# Alchemically modify first 150 atoms of Src.
alchemical_region = AlchemicalRegion(alchemical_atoms=range(150))
alchemical_system = factory.create_alchemical_system(test_system.system, alchemical_region)
globals = {
    'read_charges': read_charges,
    'alchemical_system': alchemical_system,
}
number = 1000
tot_time = timeit.timeit('read_charges(alchemical_system)', globals=globals, number=number)
print('read_charges: time for {} executions: {} s (time per execution {} s)'.format(number, tot_time, tot_time / number))
```
I got on the cluster (CUDA platform)
```
read_charges: time for 1000 executions: 0.13602294865995646 s (time per execution 0.00013602294865995646 s)
```

### DIFF
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -419,6 +419,9 @@ class AlchemicalState(object):
                 err_msg = 'Could not find parameter {} in the system'
                 raise AlchemicalStateError(err_msg.format(parameter_name))
 
+        # Write NonbondedForce charges if PME is treated exactly.
+        self._set_exact_pme_charges(system)
+
     def check_system_consistency(self, system):
         """Check if the system is consistent with the alchemical state.
 
@@ -475,6 +478,11 @@ class AlchemicalState(object):
             except Exception:
                 err_msg = 'Could not find parameter {} in context'
                 raise AlchemicalStateError(err_msg.format(parameter_name))
+
+        # Write NonbondedForce charges if PME is treated exactly.
+        updated_nonbonded_force = self._set_exact_pme_charges(context.getSystem())
+        if updated_nonbonded_force is not None:
+            updated_nonbonded_force.updateParametersInContext(context)
 
     @classmethod
     def _standardize_system(cls, system):
@@ -559,6 +567,43 @@ class AlchemicalState(object):
                 parameter_name = force.getGlobalParameterName(parameter_id)
                 if parameter_name in supported_parameters:
                     yield force, parameter_name, parameter_id
+
+    @staticmethod
+    def _find_exact_pme_forces(system):
+        """Return the NonbondedForce and the CustomNonbondedForce with the original charges."""
+        original_charges_force = None
+        nonbonded_force = None
+        for force in system.getForces():
+            if (isinstance(force, openmm.CustomNonbondedForce) and
+                        force.getEnergyFunction() == '0.0;'):
+                original_charges_force = force
+            elif isinstance(force, openmm.NonbondedForce):
+                nonbonded_force = force
+        return original_charges_force, nonbonded_force
+
+    def _set_exact_pme_charges(self, system):
+        """Write NonbondedForce charges if PME is treated exactly.
+
+        Return the updated NonbondedForce if the charges were set,
+        or None otherwise.
+        """
+        # Find CustomNonbondedForce storing the original charges
+        # and the NonbondedForce with the charges to set.
+        original_charges_force, nonbonded_force = self._find_exact_pme_forces(system)
+
+        # If we don't treat PME exactly, we don't need to set the charges.
+        if original_charges_force is None:
+            return None
+
+        # Set alchemical atoms charges.
+        lambda_electrostatics = self.lambda_electrostatics
+        _, alchemical_atoms = original_charges_force.getInteractionGroupParameters(0)
+        for atom_idx in alchemical_atoms:
+            charge, sigma, epsilon = nonbonded_force.getParticleParameters(atom_idx)
+            original_charge = original_charges_force.getParticleParameters(atom_idx)[0]
+            charge = lambda_electrostatics * original_charge
+            nonbonded_force.setParticleParameters(atom_idx, charge, sigma, epsilon)
+        return nonbonded_force
 
 
 # =============================================================================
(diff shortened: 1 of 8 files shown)


andrrizzi:
Tests pass finally.

andrrizzi:
I'm going to push here also the implementation of #321 while this is waiting for review because I need everything in one place.

andrrizzi:
I've added another couple of features/optimizations.
- There's a new option `AlchemicalState.update_alchemical_charges` that makes changes in electrostatics with exact PME treatment incompatible. This will allow us to simulate electrostatics on multiple contexts and avoid calling `updateParametersInContext`.
- `SamplerState` now strips the unit only once to positions and velocities on `apply_to_context` and cache them internally. The cache is invalidated if positions and velocities are changed. I've added the utility class `TrackedQuantity` for this purpose.

openmmtools/alchemy.py
```diff
@@ -471,13 +483,39 @@ def apply_to_context(self, context):
...
             except Exception:
                 err_msg = 'Could not find parameter {} in context'
                 raise AlchemicalStateError(err_msg.format(parameter_name))
 
-    @classmethod
-    def _standardize_system(cls, system):
+        # Handle lambda_electrostatics changes with exact PME electrostatic treatment.
+        # If the context doesn't use exact PME electrostatics, or if lambda_electrostatics
+        # hasn't changed, we don't need to do anything.
+        if (_UPDATE_ALCHEMICAL_CHARGES_PARAMETER not in context_parameters or
```
Lnaden: What happens here if `_UPDATE_ALCHEMICAL_CHARGES_PARAMETER` is False?

openmmtools/alchemy.py
```diff
@@ -1143,6 +1426,39 @@ def _build_alchemical_bond_list(alchemical_atoms, reference_forces, system):
     # Internal usage: Alchemical forces
     # -------------------------------------------------------------------------
 
+    def _add_alchemical_forces(self, alchemical_system, alchemical_forces_by_lambda):
```
Lnaden: Is it possible to fall back on not using force groups if there are not enough available?

openmmtools/alchemy.py
```diff
@@ -1167,10 +1483,11 @@ def _alchemically_modify_PeriodicTorsionForce(reference_force, alchemical_region
         """
         # Don't create a force if there are no alchemical torsions.
         if len(alchemical_region.alchemical_torsions) == 0:
-            return [copy.deepcopy(reference_force)]
+            return {'': [copy.deepcopy(reference_force)]}
```
Lnaden: I don't understand the need for the change do key-less dict here?

openmmtools/alchemy.py
```diff
@@ -1287,7 +1606,140 @@ def _alchemically_modify_HarmonicBondForce(reference_force, alchemical_region):
                 # Standard bond.
                 force.addBond(particle1, particle2, theta0, K)
 
-        return [force, custom_force]
+        return {'': [force], 'lambda_bonds': [custom_force]}
+
+    def _get_electrostatics_energy_expressions(self, reference_force):
```
Lnaden: What happens if someone specifies an exact force and not PME?

openmmtools/states.py
```diff
@@ -1478,10 +1617,8 @@ class SamplerState(object):
     velocities
     box_vectors : 3x3 simtk.unit.Quantity.
         Current box vectors (length units).
-    potential_energy : simtk.unit.Quantity or None
-        Potential energy of this configuration.
-    kinetic_energy : simtk.unit.Quantity
-        Kinetic energy of this configuration.
+    potential_energy
```
Lnaden: What happened to the docstring here?

openmmtools/alchemy.py
andrrizzi: You can set `split_alchemical_forces` to `False` if there are not enough force groups available. Otherwise, `RuntimeError` is raised.

openmmtools/alchemy.py
andrrizzi: Each `alchemically_modify_X()` method now has a different signature. The key of the returned `dict` is the parameter that controls that force so that we know upstream in which force group to put it.

openmmtools/alchemy.py
andrrizzi: As long as `reference_force` is a `NonbondedForce`, the method should behave correctly. We call this only inside `_alchemically_modify_NonbondedForce`.

openmmtools/states.py
andrrizzi: `potential_energy` and `kinetic_energy` are properties now so I moved their docstrings in the property definitions.

andrrizzi:
Thanks for the review!

I just noticed that in `AlchemicalState.__setstate__` we read the `update_alchemical_charges` attribute from the `dict` containing the serialization. I'll make that optional (that defaults to `True`) or it'll break backwards compatibility of our YANK datasets.

openmmtools/alchemy.py
Lnaden: Right, but I mean can we just default to fall back to non-`split_alchemical_forces` if there are too many forces to split into the 32 groups?

andrrizzi: Ah! I see, sorry. I went for the exception because in general, I would prefer to know if there's a performance penalty instead of proceeding silently, but I can change this behavior if you think we should automatically fall back to `split_alchemical_forces=False`.

I think it'd be very easy for an application to go with something like this anyway:
```python
try:
    factory.create_alchemical_system(...)
except RuntimeError:
    factory.split_alchemical_forces = False
    factory.create_alchemical_system(...)
```

Issue #330, opened 2018-02-08
Allowing subset of atoms to define geometric center for rotation 


kamran-haider:
I was writing code to sample water orientations with oxygen atoms fixed and found `MCRotationMove` from the `mcmc` module super-convenient. I had to make one modification though, which allowed using coordinates of the oxygen atom as the center of geometry for atoms to move. See the modified code below and also the images of water configurations before and after using the modified code. I am not sure if this would be a common scenario in the desired usage of openmmtools. If you guys think this modification may be useful, let me know and I will generate a PR. I like the idea of using openmmtools as a general toolbox for putting together components of MC simulations of biomolecules so thanks for creating it!

Thanks,
Kamran    

```python
    def rotate_positions(cls, positions, indices=None):
        """Return the positions after applying a random rotation to them.

        Parameters
        ----------
        positions : nx3 numpy.ndarray simtk.unit.Quantity
            The positions to rotate.
        indices : list
            Indices of a subset of atoms to center the rotation around, None by default.

        Returns
        -------
        rotated_positions : nx3 numpy.ndarray simtk.unit.Quantity
            The rotated positions.

        """
        positions_unit = positions.unit
        x_initial = positions / positions_unit

        # Define coordinates for the center of rotation.
        x_rot_centers = x_initial

        # Update coordinates for the center of rotation from the subset defined by indices
        if indices is not None:
            if len(indices) < 1 or len(indices) > x_initial.shape[0]:
                raise IndexError("The length of indicies must be >= % d and <= %d" % (1, x_initial.shape[0]))
            x_rot_centers = positions[indices, :]

        x_initial_mean = x_rot_centers.mean(0)

        # Generate a random rotation matrix.
        rotation_matrix = cls.generate_random_rotation_matrix()

        # Apply rotation.
        x_proposed = (rotation_matrix * np.matrix(x_initial - x_initial_mean).T).T + x_initial_mean
        return unit.Quantity(x_proposed, positions_unit)
```

Before:
![image](https://user-images.githubusercontent.com/7668971/35963925-43438e70-0c84-11e8-8b4a-04fa56a61479.png)

After:
![image](https://user-images.githubusercontent.com/7668971/35963964-58235352-0c84-11e8-9fe4-6defdc8891c4.png)

andrrizzi:
[...]

PR #331, opened 2018-02-08
Modified MC rotation moves

kamran-haider:
Potential enhancement: Following up on discussion in #330, I modified the `rotate_positions` method of `MCRotationMove` to allow a subset of atom positions define the geometric center of rotation.  The unit test for metropolized moves wouldn't test for this modification so when running tests locally, everything goes smoothly. I checked the output configurations, as shown in issue thread. Let me know if I should write a unit test for this.

### DIFF
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -1782,14 +1782,14 @@ class MCRotationMove(MetropolizedMove):
         super(MCRotationMove, self).__init__(**kwargs)
 
     @classmethod
-    def rotate_positions(cls, positions, indices=None):
+    def rotate_positions(cls, positions, rotation_center_indices=None):
         """Return the positions after applying a random rotation to them.
 
         Parameters
         ----------
         positions : nx3 numpy.ndarray simtk.unit.Quantity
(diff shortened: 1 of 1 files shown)


openmmtools/mcmc.py
```diff
@@ -1782,13 +1782,15 @@ def __init__(self, **kwargs):
...
     @classmethod
-    def rotate_positions(cls, positions):
+    def rotate_positions(cls, positions, rotation_center_indices=None):
         """Return the positions after applying a random rotation to them.
 
         Parameters
         ----------
         positions : nx3 numpy.ndarray simtk.unit.Quantity
             The positions to rotate.
+        rotation_center_indices : list
```
andrrizzi: Could you add a `, optional` here? Also, it would be fantastic to document that the center of rotation is chosen to be the center of geometry of this set of atoms, and that the default behavior uses all the atoms in `positions`.

kamran-haider: No worries at all and thanks so much for the comments.  I agree and will get back with the unit test and updated docstring.

PR #332, opened 2018-02-11
Fix typo in ndof attribute

maxentile:
In `WaterBox` and `WaterCluster`, the `ndof` attribute was computed incorrectly when `constrained=True` -- only 3 d.o.f. were subtracted from the system, regardless of the number of waters in the system.
```python
>>> WaterBox(constrained=False).ndof
4509
>>> WaterBox(constrained=True).ndof
4506
```

Fix: subtract 3 d.o.f. per water.


jchodera:
Actually, I can merge!

PR #333, opened 2018-02-16
Add option to restrain only oxygens

maxentile:
Addresses JDC's request
(https://choderalab.slack.com/archives/C0LD6795X/p1518393073000045) to
provide an option that allows the origin-centered harmonic restraining
potential to apply only to oxygens, rather than all atoms


jchodera:
Thanks!

Issue #328, opened 2018-01-30
Move some logic from factories to CustomForce objects


andrrizzi:
In our factories (alchemical and restraints in particular), we create a lot of different custom forces that serve different purposes, but to recover and manipulate those forces we currently have to rely on many implementation details to distinguish them from other forces (see for example `AbsoluteAlchemicalFactory._find_force_components` or the restraint unbiasing implementation in choderalab/yank#887).

I was thinking of expanding the `forces` module and encapsulate some of this logic into actual `RestorableForce`s objects, similarly to what we do with integrators, to easily differentiate them by type and equip them with a useful interface. I think this will simplify a lot of the code inside our factories.

PR #336, opened 2018-03-13
Restraint forces

### LINKED ISSUE #292: Correctly copy our CustomIntegrator
I think we can make all our `CustomIntegrator` deepcopiable without losing the correct class and internal attributes by overriding the `openmm.Integrator.__copy__()` method with something like this.
```python
def __copy__(self):
    """Overwrite openmm.Integrator implementation to copy class and attributes."""
    integrator_copied = super(RestorableIntegrator, self).__copy__()

    # Assign correct class instead of CustomIntegrator.
    integrator_copied.__class__ = self.__class__

    # Copy attributes. CustomIntegrator has only 1 attribute (this),
    # everything else is part of the implementation.
    attributes_self = {k: v for k, v in self.__dict__.items() if k != 'this'}
    integrator_copied.__dict__.update(copy.deepcopy(attributes_self))

    return integrator_copied
```
I can add this very easily, but I'll **have to make the `LangevinIntegrator._registered_step_types` a class variable instead of an instance variable** because Python goes nuts when you try to deepcopy a `dict` containing a bound method and you get a `RecursionError`. Does this sound ok?

Note that the recently implemented pandegroup/openmm#1870 does not solve this particular problem.

### LINKED ISSUE #335: Travis failures
Not sure how we missed this, but see the failures [here](https://travis-ci.org/choderalab/openmmtools/jobs/347651589#L3360-L3380).
```
FAIL: Doctest: openmmtools.integrators.ThermostatedIntegrator
  File "/home/travis/miniconda/envs/test/lib/python2.7/doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for openmmtools.integrators.ThermostatedIntegrator
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/openmmtools-0.14.1-py2.7.egg/openmmtools/integrators.py", line 227, in ThermostatedIntegrator
File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/openmmtools-0.14.1-py2.7.egg/openmmtools/integrators.py", line 283, in openmmtools.integrators.ThermostatedIntegrator
Failed example:
    integrator.getTemperature()
Expected:
    ...
    AttributeError: type object 'object' has no attribute '__getattr__'
Got:
    Quantity(value=380.0, unit=kelvin)
```

andrrizzi:
The main thing in this PR is the addition in the `forces.py` module of restraint forces as anticipated in #328. These forces provide a common interface that will allow us in YANK's analysis module to reweight radially-symmetric restraints without implementing non-extensible hacks to figure out the restraints properties (cc choderalab/yank#882). The most important feature introduced is the ability to compute the standard state correction with an energy/radius cutoff and reweighting to a square-well potential.

Other smaller things in this PR:
- Fix #335: doctest
- Fix #292: copy Python attributes of custom integrators and forces on `deepcopy()`
- I added the utility function `forces.find_forces()` that should be handy to extract forces from an OpenMM system by class/name.

Do I have to manually update the list of classes in the docs or are they automatically generated?


jchodera:
@andrrizzi : Is this PR ready for review?

andrrizzi:
Yep!

jchodera:
> Do I have to manually update the list of classes in the docs or are they automatically generated?

This is a good question for @Lnaden, but I think you have to manually update the list of classes.

@Lnaden : Can you review when you have a chance?

Lnaden:
> Do I have to manually update the list of classes in the docs or are they automatically generated?

If you have the `autodoc` extension included, it should compile if you also specify in the docs where you want them built. But if you allow it to just do its thing with minimal settings, it will be a bit of a cudgel and document everything it finds where you target it.

andrrizzi:
> This is great, will this wind up replacing some of the Restraint classes in yank eventually?

I don't think so. I think the classes in YANK will keep serving the role of restraint factories that create these force objects, add automatic determination of the parameters, and parse DSL/SMIRKS strings (through the `Topography` object).

> If you have the autodoc extension included, it should compile if you also specify in the docs where you want them built. But if you allow it to just do its thing with minimal settings, it will be a bit of a cudgel and document everything it finds where you target it.

Ok, thanks! I'll merge this, see what gets generated automatically, and finish up the docs update in a separate PR.

dwhswenson:
This PR seems to have removed the generic `RestorableIntegrator` class introduced in #140 ... should we now only refer to `ThermostatedIntegrator`? Will it work the same? (Especially the methods `is_restorable` and `restore_interface`). The most recent release of openmmtools is causing test failures for OpenPathSampling.

dwhswenson:
I think that the extent of our use is:

https://github.com/openpathsampling/openpathsampling/blob/master/openpathsampling/engines/openmm/engine.py#L14-L40

If there's now a better way to accomplish that, based on pure OpenMM, it would probably be better for us to do that (at least, long term). Will `simtk.openmm.XmlSerializer.deserialize(integrator_xml)` now give us the full integrator interface? Or is there something else we should use?

In any case, we might keep a variant of the function linked above to support older versions of OpenMM. But even that looks like it might be a 1-line fix, so if we're the only ones suffering the problem, it'll be as easy to fix it on our end as for you to push a bugfix.

> Apologies for the lack of advance warning!

No worries -- API changes are expected when we're using a version `0.x`!

andrrizzi:
Apologies for the change. The `RestorableIntegrator` functionalities are still there, but I've moved them into `openmmtools.utils.RestorableOpenMMObject` which is capable of doing the same thing also with `Force`s. You can also use `ThermostateIntegrator` if you prefer since it inherits from `RestorableOpenMMObject`.

PR #338, opened 2018-03-15
Optimization of CompoundThermodynamicState deserialization

andrrizzi:
As the title says.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,6 +8,10 @@ Development snapshot
 0.14.0 - Exact treatment of alchemical PME electrostatics, water cluster test system, optimizations
 ===================================================================================================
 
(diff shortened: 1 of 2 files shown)


andrrizzi:
The failing doctest is being already fixed in #336 so this should be good to go!

jchodera:
Do you need @Lnaden to review?

jchodera:
Also, we probably want to cut a bugfix release when this is merged.

andrrizzi:
I think he wanted to test this branch with YANK before reviewing/merging.

> Also, we probably want to cut a bugfix release when this is merged.

After this and #336 are merged, I'll cut a feature release.

Lnaden:
Oh yes, this is MUCH faster Merging

PR #343, opened 2018-03-21
Fix argument order for GeodesicBAOABIntegrator 

jchodera:
This fixes a problem with the API of `GeodesicBAOABIntegrator` so it can be used interchangeably with `BAOABIntegrator` and `LangevinIntegrator`.

At some point, we should add a test that makes sure all of our `LangevinIntegrators` can be invoked like:
```python
integrator = integrator_class(temperature, collision_rate, timestep)
```
to catch this issue in the future.

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -2073,7 +2073,7 @@ class BAOABIntegrator(LangevinIntegrator):
 class GeodesicBAOABIntegrator(LangevinIntegrator):
     """Create a geodesic-BAOAB integrator."""
 
-    def __init__(self, K_r=2, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


andrrizzi:
If I remember correctly, that syntax is supported only in Python 3. Until we drop Python 2 we're stuck with parsing `kwargs`.

jchodera:
> If I remember correctly, that syntax is supported only in Python 3. Until we drop Python 2 we're stuck with parsing kwargs.

So I should rework this to inspect/update `kwargs` instead of specifying a default this way?

andrrizzi:
I'm afraid so. I wonder if we should drop Python 2 support soon here as well.

andrrizzi:
Tests pass! I'll merge this and proceed with the release.

PR #354, opened 2018-06-05
[WIP] Extra-chance GHMC

maxentile:

This PR adapts the Extra-Chance GHMC integrator @kyleabeauchamp had implemented in https://github.com/kyleabeauchamp/openmmtools/tree/mjhmc2/openmmtools/hmc_integrators . When (G)HMC would normally encounter a rejection, this integrator will try again up to `n_extra_chances` times to extend the trajectory subject to a suitably modified accept-reject test. This can reduce the influence of "momentum-flipping" and appears to be much more efficient than standard GHMC in some cases.

I copy-pasted Kyle's code and modified it slightly, but it occurs to me now I could probably open a PR directly from Kyle's fork to preserve the early commit history if preferable. If so, let me know and I'll close this PR and attempt that.

(Also [fixes a small bug](https://github.com/choderalab/openmmtools/commit/db2a22f66645adf22436585c3c4908b24050d262) I noticed in the tests.)

In addition to updating the documentation and changelog, my remaining to-do's are:
- [ ] Add warnings that this is experimental
- [ ] Resolve question about correctness if `n_steps_hmc != steps_per_extra_hmc`
- [ ] Make tests pass

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -42,6 +42,7 @@ import simtk.openmm as mm
 
 from openmmtools.constants import kB
 from openmmtools import respa, utils
+import pandas as pd
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
I'm OK with whatever you feel most happy with.  My only question is whether you anticipate the need for RESPA support, which the current implementation does not support but the mixin-based approach enables.

maxentile:
Thanks Kyle! That's a good point. I don't anticipate needing RESPA support here, but it would be better to include it. I'll close this PR and open a PR from your fork instead.

PR #362, opened 2018-07-19
Add SamplerState interface to OpenMM CustomCVForce

Lnaden:
This PR gives the `SamplerState` a regular way to access the CustomCVForce variables and values which may appear in a `Context`.

### DIFF
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -19,6 +19,7 @@ import sys
 import copy
 import zlib
 import weakref
+import collections
 
 import numpy as np
 from simtk import openmm, unit
@@ -1618,6 +1619,7 @@ class SamplerState(object):
     total_energy
     volume
     n_particles
+    collective_variables
 
     Examples
     --------
@@ -1676,6 +1678,13 @@ class SamplerState(object):
     # -------------------------------------------------------------------------
 
     def __init__(self, positions, velocities=None, box_vectors=None):
+        # Allocate variables, they get set in _initialize
+        self._positions = None
+        self._velocities = None
+        self._box_vectors = None
+        self._collective_variables = None
+        self._kinetic_energy = None
+        self._potential_energy = None
         self._initialize(copy.deepcopy(positions), copy.deepcopy(velocities), copy.deepcopy(box_vectors))
 
     @classmethod
@@ -1760,16 +1769,18 @@ class SamplerState(object):
             value = unit.Quantity(value)
         self._box_vectors = value
 
+    # Derived properties
+
     @property
     def potential_energy(self):
         """simtk.unit.Quantity or None: Potential energy of this configuration."""
-        if self.positions is None or self.positions.has_changed:
+        if self._test_positions_valid:
             return None
         return self._potential_energy
 
     @potential_energy.setter
     def potential_energy(self, new_value):
-        self._potential_energy = new_value
+        raise AttributeError("Cannot set potential energy as it is a function of Context")
 
     @property
     def kinetic_energy(self):
@@ -1780,7 +1791,18 @@ class SamplerState(object):
 
     @kinetic_energy.setter
     def kinetic_energy(self, new_value):
-        self._kinetic_energy = new_value
+        raise AttributeError("Cannot set kinetic energy as it is a function of Context")
+
+    @property
+    def collective_variables(self):
+        """dict or None: Collective variables for this configuration if present in Context"""
+        if self._test_positions_valid:
+            return None
+        return self._collective_variables
+
+    @collective_variables.setter
+    def collective_variables(self, new_value):
+        raise AttributeError("Cannot set collective variables as it is a function of Context")
 
     @property
     def total_energy(self):
@@ -1833,7 +1855,9 @@ class SamplerState(object):
         ----------
         context_state : simtk.openmm.Context or simtk.openmm.State
             The object to read. If a State, it must contain information
-            on positions, velocities and energies.
+            on positions, velocities and energies. Collective
+            variables can only be updated from a Context, NOT a State
+            at the moment. If a State is provided, collective variables are nullified.
 
         Raises
         ------
@@ -1843,7 +1867,7 @@ class SamplerState(object):
         """
         self._read_context_state(context_state, check_consistency=True)
 
-    def apply_to_context(self, context, ignore_velocities=False):
+    def apply_to_context(self, context, ignore_velocities=False, ignore_collective_variables=False):
         """Set the context state.
 
         If velocities and box vectors have not been specified in the
@@ -1857,6 +1881,10 @@ class SamplerState(object):
             If True, velocities are not set in the Context even if they
             are defined. This can be useful if you only need to use the
             Context only to compute energies.
+        ignore_collective_variables : bool, optional
+            If True, the collective variables are not updated from the
+            Context, so the SamplerState's internal tracking
+            of their values remains unchanged
 
         """
         # NOTE: Box vectors MUST be updated before positions are set.
@@ -1865,6 +1893,8 @@ class SamplerState(object):
         context.setPositions(self._unitless_positions)
         if self._velocities is not None and not ignore_velocities:
             context.setVelocities(self._unitless_velocities)
+        if not ignore_collective_variables:
+            self._set_collective_variables(context)
 
     def has_nan(self):
         """Check that energies and positions are finite.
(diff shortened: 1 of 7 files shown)


Lnaden:
This does not make OpenMMTools *require* OpenMM 7.3 yet, it should gracefully trap the existence, or lack thereof, of OpenMM 7.3

Lnaden:
Are there any objections to me dropping these two lines from the `setup.py` file? They are causing build clashes between `conda build` and `setuptools`/`pip` trying to install some ancient version of `certifi` on top of the one `conda` downloads.

https://github.com/choderalab/openmmtools/blob/master/setup.py#L174-L175

andrrizzi:
No problems here!

openmmtools/states.py
```diff
@@ -2027,8 +2063,41 @@ def _read_context_state(self, context_state, check_consistency):
...
+        """
+        Update the collective variables from the context object
+
+        Parameters
+        ----------
+        context_state : simtk.openmm.Context or simtk.openmm.State
+            The object to read. This only works with Context's for now,
+            but in the future, this may support OpenMM State objects as well.
+        """
+        if isinstance(context_state, openmm.State):
```
andrrizzi: I think we should raise an exception if the user passes an `openmm.State` object and `get_collective_variables=True`. Otherwise, this will fail silently and end up with a value for `SamplerState.collective_variables` indistinguishable from having a `System` with no collective variables.

openmmtools/tests/test_states.py
```diff
@@ -854,12 +854,11 @@ def test_unitless_cache(self):
             sampler_state.positions = copy.deepcopy(positions)
             assert sampler_state._unitless_positions_cache is None
 
-            # TODO reactivate this test once OpenMM 7.2 is released with bugfix for #1940
-            # if isinstance(sampler_state._positions._value, np.ndarray):
-            #     old_unitless_positions = copy.deepcopy(sampler_state._unitless_positions)
-            #     sampler_state.positions[5:8] = [[2.0, 2.0, 2.0], [2.0, 2.0, 2.0], [2.0, 2.0, 2.0]] * pos_unit
-            #     assert sampler_state.positions.has_changed
-            #     assert np.all(old_unitless_positions[5:8] != sampler_state._unitless_positions[5:8])
+            if isinstance(sampler_state._positions._value, np.ndarray):
```
andrrizzi: Thanks for re-activating these tests! I forgot about this.

openmmtools/states.py
Lnaden: I don't really like this idea. This can lead the user to be mandated to use the `get_collective_variables=False` in every call to this function when the collective variables are less common. 

> SamplerState.collective_variables indistinguishable from having a System with no collective variables.

Yes, though I thought that was partially the point? If a user never supplies the information about a given state, then we can't use that. I don't like the idea of the users being required to now track collective variables, which are an optional construct in of themselves.

andrrizzi: The user wouldn't have to set `get_collective_variable=False` in every call as long as he pass an actual `Context` object. Passing the `State` object is there exclusively for optimization (i.e. if I already have one around I don't have to read the info off the `Context` again).

In any case, I think avoiding silent failing would be worth forcing the users to add the extra option, and the default behavior should be the safest. With this implementation, for example, all our metropolized `MCMCMove`s are currently failing silently and not retrieving CV variables that should actually be available after `apply` has been performed.

Lnaden:
Okay! Changes made, let me know what you think.

andrrizzi:
Looks great, thanks for the changes!

With the current implementation now, the integrator `MCMCMove` tests should fail because of this line: https://github.com/choderalab/openmmtools/blob/master/openmmtools/mcmc.py#L718 .

We'll have to change the implementation of `BaseIntegratorMove.apply()` to pass a `Context` to make it work, or better, if we want to maintain the current performance, we'll have to add `ignore_X=False` arguments to `SamplerState.update_from_context()`, and substitute that line with
```python
sampler_state.update_from_context(context_state, ignore_collective_variables=True)
sampler_state.update_from_context(context, ignore_positions=True, ignore_velocities=True, ignore_energy=True)
```
With repex, I think this would save us `n_states` positions+velocities read ops from the GPU w.r.t. simply passing the context.

andrrizzi:
To put that into context, with the `BaseIntegratorMove`s we already have an OpenMM `State` around because we need to check that the potential energy is not `NaN` for the nan recovery feature.

Lnaden:
I'm not sure I follow what you would prefer I do here, nor where I need to make the change

andrrizzi:
Sorry, that was unclear. I'd prefer adding the rest of the `ignore_X` arguments to `SamplerState.update_from_context()` and use
```python
sampler_state.update_from_context(context_state, ignore_collective_variables=True)
sampler_state.update_from_context(context, ignore_positions=True, ignore_velocities=True, ignore_energy=True)
```
to avoid degrading the performance of the integrator moves.

Lnaden:
Okay, but would you want all those to default to True?

andrrizzi:
I think the safer default is probably `False`, similarly to `ignore_collective_variables`, so that users will be forced to think about what they are not keeping up-to-date with the latest changes in the `Context`.

Lnaden:
Having a thought. Would it be more clear to use inclusive keywords instead of exclusive? e.g. `get_collective_variables=True` instead of `ignore_collective_variables=False`

jchodera:
That sounds clearer to me!

andrrizzi:
Me and Levi just finished talking about this. I thought of the `get_X=True` variables at the beginning, but I changed my comments to suggests `ignore_X=False` because 1) it's consistent with `apply_to_context`, which uses `ignore_velocities` instead of `set_velocities` 2) it sounds clearer to me that the old values of the ignored attributes will be left untouched instead of being set to `None`.

openmmtools/states.py
```diff
@@ -2012,23 +2089,69 @@ def _read_context_state(self, context_state, check_consistency):
...
+        context_state : simtk.openmm.Context or simtk.openmm.State
+            The object to read. This only works with Context's for now,
+            but in the future, this may support OpenMM State objects as well.
+        ignore_collective_variables : bool
+            Choose to just ignore this value
+        """
+        if ignore_collective_variables:
+            if isinstance(context_state, openmm.State):
+                # Invalidate if State is passed in
+                self._collective_variables = None
```
andrrizzi: I'm not sure we want to do this. I'm worried about the fact that these two snippets:
```python
sampler_state.update_from_context(context_state, ignore_collective_variables=True)
sampler_state.update_from_context(context, ignore_positions=True, ignore_velocities=True)
```
and
```python
sampler_state.update_from_context(context, ignore_positions=True, ignore_velocities=True)
sampler_state.update_from_context(context_state, ignore_collective_variables=True)
```
will result in two different states. I think this mistake would be hard to find and it's easy to miss this behavior in the docs. I think it's safer to have always `ignore_collective_variables` leave the CV as it is because if the user is using the `ignore_X` attributes it means that he's updating everything by steps.

Hmm. I understand now that you were probably worried about the fact that the user won't be able to invalidate a particular attribute during the manual update. If this is the case, I'd rather have the properties setters allow to set (exclusively) the `None` value.

Lnaden: Because we support both `State` and `Context` objects, which now cause a loss of information, we have to do something to handle it. I don't want the user to by default be able to put the `SamplerState` into something inconsistent.

>  I understand now that you were probably worried about the fact that the user won't be able to invalidate a particular attribute during the manual update.

When would this be?

> I'd rather have the properties setters allow to set (exclusively) the None value.

I can change that, but lets figure out all the use cases first.

andrrizzi: > I don't want the user to by default be able to put the SamplerState into something inconsistent.

This is already possible if the user does
```python
sampler_state.update_from_context(context, ignore_collective_variables)
```
and `context` is an actual `Context` object. We must assume that users know what they're doing when they are requesting only partial information to optimize the communication with the GPU, but I'm worried an inconsistent behavior of the input arguments would trick also more advanced users (me included).

Lnaden:
The collective variables should now be consistent with the other `ignore_X` flags. `State` objects are still the exception.

andrrizzi:
Thanks! Feel free to merge when tests pass.

Lnaden:
We keep having SegFaults during the `_find_thermostat` tests. Did we ever figure out why that happens on Travis only?

andrrizzi:
Nope T.T I gave up after a couple of hours of trying different things. Last time that happened the errors magically disappeared with a new OpenMM version. I'll try to figure out what's going on separately anyway. Locally, all those tests pass here.

Lnaden:
All the tests pass locally for me as well

PR #363, opened 2018-07-21
GlobalParameterState and changes for multi-restraint calculations

andrrizzi:
- Deprecated the signature of `IComposableState._on_setattr` to fix a bug where the objects were temporarily left in an inconsistent state when an exception was raised and caught.
- Deprecated `update_alchemical_charges` in `AlchemicalState` in anticipation of the new implementation of the exact PME that will be based on the `NonbondedForce` offsets rather than `updateParametersInContext()`.
- Allow restraint force classes to be controlled by a parameter other than `'lambda_restraints'`. This will enable multi-restraints simulations.
- Implemented `states.GlobalParametersState`. This more or less generalize the logic of `AlchemicalState` and implement the "suffix" mechanism that we'll be able to use in multi-restraint calculations to control parameters such as `lambda_restraints_restraint1`, `lambda_restraints_restraint2` and, in the future, in multiple alchemical region calculations. With this, the new `RestraintState` class in YANK can be simplified to:
 ```python
class RestraintState(GlobalParameterState):
    lambda_restraints = GlobalParameterState.GlobalParameter('lambda_restraints', standard_value=1.0)

    @lambda_restraints.validator
    def lambda_restraints(self, instance, new_value):
        if not (0.0 <= new_value <= 1.0):
            raise ValueError('lambda_restraints must be between 0.0 and 1.0')
        return new_value

    # A constructor just to give parameters_name_suffix a more meaningful name.
    def __init__(restraint_name=None, **kwargs):
        super().__init__(parameters_name_suffix=restraint_name, **kwargs)
```
`AlchemicalState` will be similarly simplified, but we'll have to wait to do it until the new exact PME implementation will be in.

### DIFF
--- a/openmmtools/forces.py
+++ b/openmmtools/forces.py
@@ -700,7 +700,7 @@ class RadiallySymmetricCentroidRestraintForce(RadiallySymmetricRestraintForce,
                  restrained_atom_indices1, restrained_atom_indices2,
                  controlling_parameter_name='lambda_restraints'):
         # Initialize CustomCentroidBondForce.
-        energy_function = controlling_parameter_name + ' * ' + energy_function
+        energy_function = controlling_parameter_name + ' * (' + energy_function + ')'
         custom_centroid_bond_force_args = [2, energy_function]
         super(RadiallySymmetricCentroidRestraintForce, self).__init__(
             restraint_parameters, restrained_atom_indices1, restrained_atom_indices2,
@@ -749,7 +749,7 @@ class RadiallySymmetricBondRestraintForce(RadiallySymmetricRestraintForce,
                  controlling_parameter_name='lambda_restraints'):
         # Initialize CustomBondForce.
         energy_function = energy_function.replace('distance(g1,g2)', 'r')
-        energy_function = controlling_parameter_name + ' * ' + energy_function
+        energy_function = controlling_parameter_name + ' * (' + energy_function + ')'
         super(RadiallySymmetricBondRestraintForce, self).__init__(
             restraint_parameters, [restrained_atom_index1], [restrained_atom_index2],
             controlling_parameter_name, energy_function)
--- a/openmmtools/forces.py
+++ b/openmmtools/forces.py
@@ -237,11 +237,15 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject):
     and :func:`restrained_atom_indices2` (with their setters) that return
     the indices of the restrained atoms.
 
-    You will also have to implement :func:`_create_bond`, which should add the
-    bond using the correct function/signature.
+    You will also have to implement :func:`_create_bond`, which should add
+    the bond using the correct function/signature.
 
     Optionally, you can implement :func:`distance_at_energy` if an
     analytical expression for distance(potential_energy) exists.
+
+    If you subclass this, and plan on adding additional global parameters,
+    you need to invoke this class ``super().__init__`` first as the
+    ``controlling_parameter_name`` must be the first global variable.
 
     Parameters
     ----------
@@ -269,7 +273,6 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject):
                  restrained_atom_indices2, controlling_parameter_name,
                  *args, **kwargs):
         super(RadiallySymmetricRestraintForce, self).__init__(*args, **kwargs)
-        self._controlling_parameter_name = controlling_parameter_name
 
         # Unzip bond parameters names and values from dict.
         assert len(restraint_parameters) == 1 or isinstance(restraint_parameters, collections.OrderedDict)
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by andrrizzi** [comment]
The two failing tests (see #358) have been already fixed in #362 so we can ignore them here.

**Comment by Lnaden** [comment]
Is this ready for merging? If so, I am going to pull master into here to see if it fixes the tests. I  want to make sure this is working as intended. Do you also want to try to fix some of the CodeClimate things its complaining about? I can take a pass through that if you have other things you are working o.

**Comment by andrrizzi** [comment]
Whops! I didn't notice the failing doctests, I can fix it. I've already merged `master` to fix a conflict with the release history.

**Comment by andrrizzi** [comment]
Tests finally pass! Hopefully, this is the last time I need to make special code for Python 2.

File: openmmtools/forces.py [hunk 1]
```diff
@@ -254,14 +254,22 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject):
...
 
+    Attributes
+    ----------
+    controlling_parameter_name
+
     """
 
     def __init__(self, restraint_parameters, restrained_atom_indices1,
-                 restrained_atom_indices2, *args, **kwargs):
+                 restrained_atom_indices2, controlling_parameter_name,
```
Lnaden: Can we make `controlling_parameter_name` have a default? E.g. `controlling_parameter_name = 'lambda_restraints'`? Or would we instead just want `controlling_parameter_name = 'restraints'` and the `lambda_` prefix is assumed prepended?

openmmtools/forces.py
```diff
@@ -669,23 +683,28 @@ class RadiallySymmetricCentroidRestraintForce(RadiallySymmetricRestraintForce,
...
     restrained_atom_indices1
     restrained_atom_indices2
+    controlling_parameter_name
 
     """
 
     def __init__(self, energy_function, restraint_parameters,
-                 restrained_atom_indices1, restrained_atom_indices2):
+                 restrained_atom_indices1, restrained_atom_indices2,
+                 controlling_parameter_name='lambda_restraints'):
```
Lnaden: Ah, see, here in this subclass, `controlling_parameter_name` an optional keyword arg, which changes the call from the parent function. We should make it consistent in the base class as well

openmmtools/forces.py
```diff
@@ -726,12 +745,14 @@ class RadiallySymmetricBondRestraintForce(RadiallySymmetricRestraintForce,
...
     def __init__(self, energy_function, restraint_parameters,
-                 restrained_atom_index1, restrained_atom_index2):
+                 restrained_atom_index1, restrained_atom_index2,
+                 controlling_parameter_name='lambda_restraints'):
         # Initialize CustomBondForce.
         energy_function = energy_function.replace('distance(g1,g2)', 'r')
-        energy_function = 'lambda_restraints * ' + energy_function
-        super(RadiallySymmetricBondRestraintForce, self).__init__(restraint_parameters,
-                 [restrained_atom_index1], [restrained_atom_index2], energy_function)
+        energy_function = controlling_parameter_name + ' * ' + energy_function
```
Lnaden: So I just noticed a potential bug here. If the user passes in an `energy_function` which has additive terms and not wrapped in `( )`, this could lead to unintended behavior, which would NOT throw an error. Off the top of my head, the only practical thing I can think of is some kind of external field or shifted function, but this is also an easy enough fix it may be worth fixing now.

e.g.
```python
>>> energy_function = '(K/2)*r^2 + 5*r'
>>> controlling_parameter_name = 'X`
>>> controlling_parameter_name + ' * ' + energy_function
X * (K/2)*r^2 + 5*r
```
the `5*r` term won't be scaled.

Easy fix is just `controlling_parameter_name + ' * (' + energy_function + ')'`

openmmtools/forces.py
andrrizzi: > We should make it consistent in the base class as well

Sounds good! I'll move the default to the base class and use kwargs here. I'll have to keep it buried in `kwargs` in the base class though because Python 2 doesn't allow having both kwargs and keyword arguments.

openmmtools/forces.py
andrrizzi: Good catch! Will change it.

openmmtools/forces.py
Lnaden: Oh yeah... I forgot about that. Might be best to leave it then and it can be removed in 0.17 when we drop Py2 officially

andrrizzi: Ah, I just rememberd why I did this. It's because I need to use `controlling_parameter_name` in the subclass so I need the default to be specified there. I wouldn't write the default in two places as it's harder to keep track of (i.e., if a dev changes only one of the two the code may become inconsistent and unintuitive).

Lnaden: Also a good point

Lnaden:
Awesome, I think this is ready. Do we want to cut a release after this?

andrrizzi:
I'd wait to release until we're sure these changes will work for the multi-restriants feature in YANK. I don't think we have an `openmmtools-dev` version uploaded, but we could run the tests locally after we've implemented the new features in YANK before releasing this.

Lnaden:
Okay we'll hold off on the release. I'll use this branch for developing the new restraints in YANK. I'm on the fence about whether we should merge this or not. If we plan on making changes to this, then I would be hesitant to do so.

andrrizzi:
That's fine I think! Let's remember to merge it before making other changes to `openmmtools` though.

Lnaden:
HA! HAHAHAHAHAHA! Finally got the windows tests to pass!

andrrizzi:
I've reverted the commits implementing the CV restraint conversion, which are blocking this PR. Once I merge this, I'll open a separate PR reverting the reverted commits that we can merge once we've figure out a way to solve the performance issues.

Issue #352, opened 2018-05-02
Ensure Context cache resets global context parameters and integrator parameters 


jchodera:
When using something like the [`FIREMinimizationIntegrator`](https://github.com/choderalab/yank/blob/master/Yank/fire.py#L11) or integrators that depend on global context or integrator parameters to track progress through a cycle, it is essential that the `Context` cache resets these parameters to their corresponding settings in the passed integrator when the context is retrieved. I don't believe this is currently done.

PR #364, opened 2018-07-30
Extend default compatibility of integrators

andrrizzi:
This adds the optimization discussed in #352. The state of the integrator was always consistent even before this change, but now the compatibility of the integrator has been extended by default to all global variables, so new `Context` objects are created less often.

This also adds a blacklist `ContextCache.INCOMPATIBLE_INTEGRATOR_ATTRIBUTES` that can be used to exclude some integrator parameters from the copy when requesting the cached context.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -1,13 +1,20 @@
 Release History
 ===============
 
-0.15.1 - Something New This Way Comes [WIP]
-===========================================
+0.15.1 - Development
+====================
 
 New features
 ------------
-- Add ability for ``SamplerState`` to access new `OpenMM Custom CV Force Variables <http://docs.openmm.org/development/api-python/generated/simtk.openmm.openmm.CustomCVForce.html#simtk.openmm.openmm.CustomCVForce>`_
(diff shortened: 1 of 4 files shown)


andrrizzi:
I'll change the release history separately to avoid conflicts with #363 .

andrrizzi:
I ended up updating the release history here since we don't know when/if we'll merge #363 . This is ready for review.

andrrizzi:
This is ready. Is it ok if I merge this?

andrrizzi:
@jchodera I updated the default number of restart attempts after a NaN from 0 to 4 as we discussed today.

jchodera:
Did we want to do 4 or 6? What is a reasonable default here?

andrrizzi:
We've been using 6 with context reinitialization at the last step, but that was a made up number as well so I don't know which one is better. If context reinitialization is required for some reason, with 4 we get there faster (and keep eventual parallel chains blocked for a smaller time). Otherwise, I can set it to 6 if we want to be more conservative. I have no particular preference here as we can set the argument to any number anyway if necessary.

jchodera:
Let's stick with 4 for now then!

PR #365, opened 2018-07-31
Convert RadiallySymmetricRestraint Forces to CV Forces

Lnaden:
This is my implementation of the CV Force conversion of the `RadiallySymmetricRestraintForce` class and subclasses. The call to the parent class is slightly different, the parent class is a `CustomCVForce`, and subclasses are no longer `CustomBond-` or `CustomCentroidBondForce`s.

These tests pass locally but the implementation should still be reviewed. 

While waiting for a review, I will work on the YANK side of things to take advantage of these changes.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -27,6 +27,18 @@ Deprecated
 - Deprecated ``update_alchemical_charges`` in ``AlchemicalState`` in anticipation of the new implementation of the
   exact PME that will be based on the ``NonbondedForce`` offsets rather than ``updateParametersInContext()``.
 
+Compatibility Warnings
+----------------------
+- ``RadiallySymmetricRestraintForce``'s conversion to ``CustomCVForce``'s break the backwards compatibility with older
+  codes which implement subclasses of it. The largest change is how the Energy Function is implemented, and subclass
+  hierarchy. E.g. All these classes are now ``CustomCVForce``'s instead of ``CustomBondForce`` and
+  ``CustomCentroidBondForce`` objects.
+
+     - This affects ``RadiallySymmetricRestraintForce``,
+       ``RadiallySymmetricCentroidRestraintForce``, ``RadiallySymmetricBondRestraintForce``,
+       ``HarmonicRestraintForceMixIn``, ``HarmonicRestraintForce``, ``HarmonicRestraintBondForce``,
+       ``FlatBottomRestraintForceMixIn``, ``FlatBottomRestraintForce``, and ``FlatBottomRestraintBondForce``.
+
 
 0.15.0 - Restraint forces
 =========================
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -15,6 +15,9 @@ New features
   control global variables (`#363 <https://github.com/choderalab/openmmtools/pull/363>`_).
 - Allow restraint force classes to be controlled by a parameter other than ``lambda_restraints``. This will enable
   multi-restraints simulations (`#363 <https://github.com/choderalab/openmmtools/pull/363>`_).
+- ``RadiallySymmetricRestraintForce`` and all subclasses are now a ``CustomCVForce`` subclass wrapping their
+  respective ``CustomBondForce`` and ``CustomCentroidBondForce`` objects. This breaks backwards compatibility, but
+  enables tracking the restraint distances through the new ``SamplerSate`` collective variable features.
 
 Deprecated
 ----------
--- a/openmmtools/forces.py
+++ b/openmmtools/forces.py
@@ -21,6 +21,7 @@ import inspect
 import logging
 import math
 import re
+import yaml
 
 import scipy
 import numpy as np
@@ -253,25 +254,43 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject,
     Optionally, you can implement :func:`distance_at_energy` if an
     analytical expression for distance(potential_energy) exists.
 
+    If you want to access any of the restraint parameters, concrete class
+    implementations must infer them from the energy function, including
+    if they should bear units or not. A helper function exists in
+    this parent class called :func:``_get_parameter_in_energy_function``
+    which accepts a single arg of a string of the parameter name
+    (case sensitive) and returns the int or float value that is the
+    parameter.
+
+    If you subclass this, and plan on adding additional
+    global parameters, you need to invoke this class super().__init__ first as the
+    ``controlling_parameter_name`` must be the first global variable.
+
     Parameters
     ----------
     restraint_parameters : OrderedDict
         An ordered dictionary containing the bond parameters in the form
         parameter_name: parameter_value. The order is important to make
         sure that parameters can be retrieved from the bond force with
-        the correct force index.
+        the correct force index. If these are unit bearing quantities,
+        these will be converted to the OpenMM MD Unit system before
+        being stripped and assigned to the Force
     restrained_atom_indices1 : iterable of int
         The indices of the first group of atoms to restrain.
     restrained_atom_indices2 : iterable of int
         The indices of the second group of atoms to restrain.
     controlling_parameter_name : str
         The name of the global parameter controlling the energy function.
-    energy_string : str
+    energy_function : str
         Energy Expression for the CustomCVForce object. Use the variable
-        "Distance" to reference the internal distance force CV.
+        "symmetric_restraint_distance" to reference the internal distance force CV.
         You should NOT add the value of the parameters in the
         string, the ``restraint_parameters`` dict will automatically
-        augment this string, overwriting any setting.
+        augment this string.
+
+        If you were to accidentally include the values, then
+        the restraint_parameters key is effectively ignored, and
+        could lead to unintended behavior
     *args, **kwargs
         Additional parameters to pass to the super constructor.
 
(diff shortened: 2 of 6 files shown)


docs/releasehistory.rst
```diff
@@ -15,6 +15,9 @@ New features
   control global variables (`#363 <https://github.com/choderalab/openmmtools/pull/363>`_).
 - Allow restraint force classes to be controlled by a parameter other than ``lambda_restraints``. This will enable
   multi-restraints simulations (`#363 <https://github.com/choderalab/openmmtools/pull/363>`_).
+- ``RadiallySymmetricRestraintForce`` and all subclasses are now a ``CustomCVForce`` subclass wrapping their
```
jchodera: We should warn that this may be an API-breaking change in the changelog!

openmmtools/forces.py
```diff
@@ -267,35 +283,65 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject):
...
+        # Construct
         super(RadiallySymmetricRestraintForce, self).__init__(*args, **kwargs)
         self._controlling_parameter_name = controlling_parameter_name
 
-        # Unzip bond parameters names and values from dict.
-        assert len(restraint_parameters) == 1 or isinstance(restraint_parameters, collections.OrderedDict)
-        parameter_names, parameter_values = zip(*restraint_parameters.items())
+        # Create the distance force used for the CV force wraps
+        distance_force = self._create_distance_force()
+        self.addCollectiveVariable("Distance", distance_force)
```
jchodera: What should our naming convention be for these CVs? 
* Do we want to capitalize them (`Distance`) or make them lowercase (`distance`)? 
* Be more explicit in the names (e.g. `receptor_ligand_distance`) or use math symbols (`r`, `theta1`, etc.)?

docs/releasehistory.rst
Lnaden: I did?

``RadiallySymmetricRestraintForce`` and all subclasses are now a ``CustomCVForce`` subclass wrapping their  respective ``CustomBondForce`` and ``CustomCentroidBondForce`` objects. **This breaks backwards compatibility**, but enables tracking the restraint distances through the new ``SamplerSate`` collective variable features.

Emphasis mine

openmmtools/forces.py
Lnaden: I prefer the capital "Distance` because it clearly indicates this is a unique variable the user must think about. I don't want to use `r` or `distance(g1,g2)` here because this parent class is agnostic of the underlying CV implementation.

docs/releasehistory.rst
jchodera: Just meant that there might be a "Warnings" section rather than burying this in the "Features" (since it may break your program, which may not be a "feature").

openmmtools/forces.py
jchodera: Have we thought through what names we will use for CVs for all the other restraints?

docs/releasehistory.rst
Lnaden: Fair point. I'll make this more visible

openmmtools/forces.py
Lnaden: Right now the only other restraint classes I have are the Boresch and RMSD ones. 

The RMSD one's have the CV `RMSD`
Boresch has CVs `Distance`, `AngleA`, `AngleB`, `TorsionA`/`B`/`C.

There is overlap there, yes, but the way I implemented the `SamplerState` CV checking is supports multiple forces having overlapping CV Names. I see the overlap here. I'm open to new name ideas, its very easy to change at this point since I have not implemented any of the analysis to take advantage of this conversion yet.

Lnaden: [Here](https://github.com/choderalab/yank/blob/96249bcd7b26a779df880052de3813afa1bad3c1/Yank/restraints.py#L1639-L1682) is the reference implementation of the Boresch restraints are in YANK at the moment.

openmmtools/forces.py
```diff
@@ -256,8 +266,14 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject):
         The indices of the second group of atoms to restrain.
     controlling_parameter_name : str
         The name of the global parameter controlling the energy function.
+    energy_string : str
```
andrrizzi: Should we maybe call this `energy_function` for consistency with the other classes?

openmmtools/forces.py
```diff
@@ -267,35 +283,67 @@ class RadiallySymmetricRestraintForce(utils.RestorableOpenMMObject):
...
+    def _create_distance_force(self):
+        """
+        Create and return the underlying force object
+
+        Returns
+        -------
+        force : OpenMM.Force
+            The force measuring distance to be used as the CV
+        """
+
```
andrrizzi: I didn't know that creating a function without at least `pass` was valid syntax!

Lnaden: Neither did I. That was an oversight on my part. I think having a `pass` here is better anyways so its clear that this is an empty function and the end of the function.

openmmtools/forces.py
```diff
@@ -321,10 +369,7 @@ def restrained_atom_indices2(self):
     @property
     def restraint_parameters(self):
         """OrderedDict: The restraint parameters in dictionary form."""
-        parameter_values = self.getBondParameters(0)[-1]
-        restraint_parameters = [(self.getPerBondParameterName(parameter_idx), parameter_value)
-                                for parameter_idx, parameter_value in enumerate(parameter_values)]
-        return collections.OrderedDict(restraint_parameters)
+        return collections.OrderedDict(self._restraint_parameters)
```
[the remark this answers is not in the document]
Lnaden: What if we did not rely on the `OpenMM.XmlSerializer` alone to serialize and de-serialize the force? Could we add a `__getstate__` and `__setstate__` for the base force objects and use the OpenMMTools `serialize` and `deserialize` functions?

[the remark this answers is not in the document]
andrrizzi: I doubt it is possible unfortunately because I think the serialization of the `System` is performed at the C/C++ level and it automatically serializes the forces in the system in XML format.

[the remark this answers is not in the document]
andrrizzi: > The other option would be to infer it from the `energy_function` as you said then?

Yes, for `_controlling_parameter_name`, although the `controlling_parameter_name + ' * (' + energy_function + ')'` is added in the subclass right now, so we may need to push that line up one level in the hierarchy to keep things clean. I believe that line is shared in all subclasses anyway.

> but I am worried that will only lead to confusing namespace overwrites

I agree. I'm also worried about the fact that a serialized system with this method won't be able to be read as a normal OpenMM `System` with `XmlSerializer.deserializeSystem()` without our code.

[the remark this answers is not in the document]
andrrizzi: Whops! Sorry I was overthinking it. We'll just have to return `self.getGlobalParameterName(0)` inside the `controlling_parameter_name` property. Maybe also placing an `assert` statement on `__init__()` to check that at that point there are no other global parameters to make sure the index is correct.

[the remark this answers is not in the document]
Lnaden: @andrrizzi on what we talked about, there actually is a `restraint_parameters` property which already exists in the released version. Without some way to generically parse it, I don't know how to preserve this property.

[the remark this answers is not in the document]
andrrizzi: Ah, you're right, I forgot about that! Let's just push the implementation of that property in the subclasses then since they know which one to retrieve.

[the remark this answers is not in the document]
Lnaden: Although I don't think anything actually uses that property in OpenMMTools or YANK

openmmtools/forces.py
```diff
@@ -227,37 +228,71 @@ def _compute_harmonic_radius(spring_constant, potential_energy):
...
+        being stripped and assigned to the Force
     restrained_atom_indices1 : iterable of int
         The indices of the first group of atoms to restrain.
     restrained_atom_indices2 : iterable of int
         The indices of the second group of atoms to restrain.
     controlling_parameter_name : str
         The name of the global parameter controlling the energy function.
+    energy_function : str
+        Energy Expression for the CustomCVForce object. Use the variable
+        "symmetric_restraint_distance" to reference the internal distance force CV.
```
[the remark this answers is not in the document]
andrrizzi: Sounds good to me!

openmmtools/forces.py
```diff
@@ -21,6 +21,7 @@
 import logging
 import math
 import re
+import yaml
```
andrrizzi: Are we using this module here?

openmmtools/forces.py
```diff
@@ -656,9 +724,48 @@ def _determine_integral_limits(self, thermodynamic_state, radius_cutoff,
...
+        """
+        energy_function = self.getEnergyFunction()
+        # This compile will find the *first* entry, and therefore the
+        # one OpenMM uses as its value in implementation.
+        # There are some ambiguous energy function corner cases users could write,
+        # but that is then their error.
+        # Should catch int, float, +-, sci. notation
+        re_search = re.compile(r';{} = ([-+\d\.eE]+);'.format(parameter))
+        try:
+            value = re_search.search(energy_function).group(1)
```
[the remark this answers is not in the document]
andrrizzi: I think this case is unlikely enough that we could just mention it in the docs and leave it as undefined behavior of the function, but if we want to catch it, I'd probably go for an `assert` in `__init__`.

openmmtools/forces.py
Lnaden: Oh, no. I had that in as part of other serialization attempts. I'll fix it.

Lnaden:
The windows tests are failing because netcdf is having DLL issues, even when I pin `libnetcdf` to the exact version the last stable conda-forge build version on AppVeyor, it still fails. Since that does not bear on this PR, I'm merging to keep moving on more important things and we will have to figure that out later

PR #366, opened 2018-08-15
Mulitiple Alchemical Regions

adw62:
This adds multiple alchemical regions for Nonbonded, Angle, Bond and Torsion forces where each region is controlled by a different `AlchemicalState`. 

Although this is working in my testing `examples/MultiRegion_example/` I think it is still a WIP so would be interested to hear how it can be changed and improved.

Thanks,
Alex


jchodera:
Thanks so much for this!

@andrrizzi is on vacation this week, but can review early next week!

andrrizzi:
Whoops! Sorry I missed that this PR was open. I'll try review during the weekend.

openmmtools/alchemy.py
```diff
@@ -1482,37 +1593,36 @@ def _alchemically_modify_PeriodicTorsionForce(reference_force, alchemical_region
...
 
-        # Create PeriodicTorsionForce to handle unmodified torsions.
-        force = openmm.PeriodicTorsionForce()
-        force.setForceGroup(reference_force.getForceGroup())
+        if len(alchemical_region.alchemical_torsions) == 0:
+            return {'': [copy.deepcopy(force)]}
 
         # Create CustomTorsionForce to handle alchemically modified torsions.
-        energy_function = "lambda_torsions*k*(1+cos(periodicity*theta-phase))"
+        energy_function = "lambda_torsions_{}*k*(1+cos(periodicity*theta-phase))".format(alchemical_region.name)
```
andrrizzi: I think this will result in the variable being called `lambda_torsions_None` if `alchemical_region.region_name` is not set. It should be `lambda_torsions` (without the trailing `_` character) in this case.

The same comment applies for all the lambda variable names in the energy expressions below.

openmmtools/alchemy.py
```diff
@@ -1582,43 +1691,59 @@ def _alchemically_modify_HarmonicBondForce(reference_force, alchemical_region):
...
+        return sterics_mixing_rules, exceptions_sterics_energy_expression
+
+    def _get_electrostatics_energy_expressions(self, reference_force, alchemical_region_idx):
         """Return the energy expressions for electrostatics."""
 
         # The final expression will be prefix + method + suffix.
-        electrostatics_prefix = ('U_electrostatics;'
-                                 'U_electrostatics=(lambda_electrostatics^softcore_d)*ONE_4PI_EPS0*chargeprod')
+        electrostatics_prefix = ('U_{0}_electrostatics;'
+                                 'U_{0}_electrostatics=(lambda_electrostatics_{0}^softcore_d)*ONE_4PI_EPS0*chargeprod').format(alchemical_region_idx)
```
andrrizzi: I believe you can avoid adding the name of the region to `U_electrostatics`. It'll be a local variable in the kernel anyway so it's not a problem if there are duplicate variables in different forces.

openmmtools/alchemy.py
```diff
@@ -1788,28 +1913,27 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
-        exceptions_sterics_energy_expression = ('U_sterics;'
-                                                'U_sterics = (lambda_sterics^softcore_a)*4*epsilon*x*(x-1.0);'
-                                                'x = (sigma/reff_sterics)^6;'
-                                                # Effective softcore distance for sterics.
-                                                'reff_sterics = sigma*((softcore_alpha*(1.0-lambda_sterics)^softcore_b + (r/sigma)^softcore_c))^(1/softcore_c);')
-
-        # Define energy expression for sterics.
+        RegionName = alchemical_region.name
+        #if RegionName == None:
+        #    RegionName = 'zero'
```
andrrizzi: Fix PEP8 conventions.

andrrizzi:
I'm also thinking that exact treatment of PME will be a nightmare to implement with the current system so it's probably a good idea in this PR to raise an error if `alchemical_pme_treatment == 'exact' and len(alchemical_regions) > 1`. We can fix it much more easily once OpenMM 7.3 is released.

jchodera:
>  We can fix it much more easily once OpenMM 7.3 is released.

We've just built the 7.3 beta, so release is imminent.

adw62:
Thanks for the feedback, really appreciate it! I can start making these changes to work with just direct space electrostatics for now.

Can I use  `GlobalParameterState` with `AlchemicalState` currently or is this something to come in the future?

andrrizzi:
The `GlobalParameterState` class is implemented (you can see the docstring example [here](https://github.com/choderalab/openmmtools/blob/multi-restraints/openmmtools/states.py#L2672-L2820)), but unfortunately, it's not on the `master` branch yet, it's in `multi-restraint`. We haven't merged that yet because we may have to revert some other features in that PR. You can still take it as a model though! The `GlobalParameterState` class should be pretty stable, and the docstring has an example on how you can implement your own `ComposableState` using `GlobalParameterState` as a parent class.

PR #367, opened 2018-09-09
Tutorials

andrrizzi:
This adds the getting started tutorial and the developer's guide. I've added the test of the code snippets using the `sphinx.ext.doctest` plugin, but I don't expect the tests to pass until #363 is merged, which contains `GlobalParameterState` and is blocked by the `CVRestraint`s performance problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
   # TODO: remove the IGNORE_EXCEPTION_DETAIL when we drop Python2 support
   - cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=2 --with-timer --with-doctest --doctest-options="+IGNORE_EXCEPTION_DETAIL" -a '!slow' && cd ..
   # Test the tutorials
-  - conda install --yes --quiet sphinx
+  - conda install --yes --quiet sphinx numpydoc
   - cd docs && sphinx-build -b doctest . _build/ gettingstarted.rst devtutorial.rst && cd ..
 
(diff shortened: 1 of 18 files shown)


jchodera:
Thanks! Do you want any feedback on this?

Let's chat on Monday about whether we want to merge the `CustomCVForce` refactoring. I think we want to go ahead with it and take the 20% speed hit for now. The performance impact can likely be reduced later. The major performance issue we should worry about for now is the `offset` changes to `AbsoluteAlchemicalFactory` since this is ~50% speed hit right now.

andrrizzi:
> Thanks! Do you want any feedback on this?

No need thank you! It's pretty much what I presented on Thu, and we can correct it while we find errors. I'm trying to get the sphinx tests to work now.

> Let's chat on Monday about whether we want to merge the `CustomCVForce` refactoring.

Sounds good! Any time works for me.

jchodera:
Great! Feel free to merge whenever!

andrrizzi:
Hmm. Three failures and three completely different errors for Py2.7, 3.5, and 3.6. The sphinx tests pass locally so I'm not sure about what's going on. I might give up on using Travis to test the tutorial for now.

jchodera:
Sounds fine! I'd go ahead and merge and see if we can refine from there?

andrrizzi:
Sounds good! I'll remove the modifications to the `.travis.yaml` file before merging.

PR #369, opened 2018-09-17
[WIP] Convert restraint forces to inherit from CVCustomForce

andrrizzi:
These features were implemented originally in #365 and merged in #363, but we had to move them to a separate PR because we haven't resolved the performance issues that come with the `CVCustomForce` yet.

If the OpenMM performance does not increase, we may need to expose both versions: `CVCustomForce` and `CustomCentroidBondForce`.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -1,6 +1,29 @@
 Release History
 ***************
 
+0.17.0 - Development
+====================
+
+New features
+------------
+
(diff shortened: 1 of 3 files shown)


jchodera:
I think we had decided to expose both `CustomCVForce` and non-`CustomCVForce` versions of these. Perhaps it would be simplest to retain the original force objects in `openmmtools.forces` and instead put the new `CustomCVForce` forces in `openmmtools.cvforces`, so we can simply import from one or the other to get the desired functionality?

andrrizzi:
I agree, but I'll keep this PR open until we solve the problem in any case as a reference.

PR #372, opened 2018-09-18
[WIP] Multiple Alchemical Regions

adw62:
This should allow for multiple alchemical regions when using direct-space electrostatics.

`_alchemically_modify_NonbondedForce()` will now loop over `alchemical_regions` extending `forces_by_lambda` to build up the system. 

Previously I neglected the forces between pairs of `alchemical_regions` these are now included and added to `forces_by_lambda` under the parameter associated with the first region in the pair.  

Cheers,
Alex


andrrizzi:
Great, thank you! I'll review as soon as I have time.

Actually, could you edit this PR to be merged to the `alchemical-regions` branch that I just created instead of `master`? This way I'll be able to also implement exact PME with OpenMM 7.3 and convert `AlchemicalState` to use the new `GlobalParameterState` before we deploy the new code.

adw62:
Thanks for the feedback! I can think of quite a few changes I should make to `_alchemically_modify_NonbondedForce()` based on these comments. Might be worth waiting until I make these changes to continue reviewing. Should be able to make them today.

openmmtools/alchemy.py
```diff
@@ -1787,56 +1977,101 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
-        use_exact_pme_treatment = is_ewald_method and self.alchemical_pme_treatment == 'exact'
+            is_periodic_method = is_ewald_method or nonbonded_method == openmm.NonbondedForce.CutoffPeriodic
+            use_exact_pme_treatment = is_ewald_method and self.alchemical_pme_treatment == 'exact'
 
-        # Warn about reaction field.
-        if is_rf_method:
-            logger.warning('Reaction field support is still experimental. For free energy '
+            # Warn about reaction field.
+            if is_rf_method:
+                logger.warning('Reaction field support is still experimental. For free energy '
```
andrrizzi: These flags (from here up to `# Define energy expression for electrostatics based on nonbonded method.`) depend only on the reference `NonbondedForce` and can be moved outside the loop.

openmmtools/alchemy.py
```diff
@@ -1869,261 +2104,311 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
+                    aa_electrostatics_custom_nonbonded_force.addInteractionGroup(alchemical_atomset_0, alchemical_atomset_0)
+
+            # ---------------------------------------------------------------
+            # Distribute exceptions contributions in appropriate bond forces
+            # ---------------------------------------------------------------
+
+            all_custom_nonbonded_forces = all_sterics_custom_nonbonded_forces + all_electrostatics_custom_nonbonded_forces
+            # Move all NonbondedForce exception terms for alchemically-modified particles to CustomBondForces.
+            if len(region_names) > 1:
+                for exception_index in range(reference_force.getNumExceptions()):
```
andrrizzi: I think the two for loops (for `len(region_names) > 1` and `else`) are identical beside for the definition of `both_alchemical` and `only_one_alchemical`, with the latter always `False` in the `len(region_names) > 1` so you could have a single for loop with an internal if-else rather than the opposite.

openmmtools/alchemy.py
```diff
@@ -1787,56 +1977,101 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
         """
         # TODO Change softcore_beta to a dimensionless scalar to multiply some intrinsic length-scale, like Lennard-Jones alpha.
         # TODO Try using a single, common "reff" effective softcore distance for both Lennard-Jones and Coulomb.
-
+        forces_by_lambda = {}
```
andrrizzi: It may be good for readability to initialize the `forces_by_lambda` variable right before you use it at the bottom of the function.

openmmtools/alchemy.py
```diff
@@ -1787,56 +1977,101 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
 
-        # --------------------------------------------------
-        # Determine energy expression for all custom forces
-        # --------------------------------------------------
+        region_permus = non_pairs
+        if len(populated_region_ids) > 1:
+            for alchemical_region in alchemical_regions:
+                if alchemical_region.region_name == None:
+                    raise ValueError('For a alchemical calculation with multiple regions interacting via Nonbonded Forces '
+                                     'all regions populated with atoms should be named')
```
andrrizzi: I think we can make this requirement stricter (i.e. the exception is raised always when there are multiple unnamed alchemical regions) and move this check in `create_alchemical_system()`.

openmmtools/alchemy.py
```diff
@@ -1787,56 +1977,101 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
+            #region_permus is all permutations of alchemical regions so we can deal with interactions between
+            # alchemical regions (aa)(pairs). And the standard interactions of the individual alchemical regions (non_pairs)
+            #with themselves (aa)(non_pair) and the non-alcemical regions (na)(non_pair)
+            pairs = (list(itertools.combinations(populated_region_ids, 2)))
+            region_permus += pairs
```
andrrizzi: If I understand correctly the code, you may be able to substitute the block of code above with
```python
region_permutations = itertools.combinations_with_replacement(range(len(alchemical_regions)), 2)
```
and maybe move the definition right before the for loop for readability.

PR #377, opened 2018-10-01
Fix readthedocs

jchodera:
This PR fixes the `docs/environment.yml` to ensure the readthedocs build works.

Here is a build of the [fixed branch](https://openmmtools.readthedocs.io/en/fix-docs/).


andrrizzi:
Thank you!

PR #380, opened 2018-10-22
Reimplement exact PME to use parameter offset in OpenMM 7.3

andrrizzi:
This removes the old implementation of exact PME that used `updateParametersInContext` in favor of using the parameter offset feature offered in OpenMM 7.3.

`AlchemicalState` now is a very short extension of `GlobalParameterState` that provides a general way to implement composable states controlling global parameters. I've also added `GlobalParameterFunction`, which is the general version of `AlchemicalFunction`.

I've also removed support for Python 2 here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,6 @@ script:
 
 env:
   matrix:
-    - python=2.7 CONDA_PY=27
     - python=3.5 CONDA_PY=35
     - python=3.5 CONDA_PY=35 DEVOMNIA=true
     - python=3.6 CONDA_PY=36
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -13,14 +13,14 @@ requirements:
   build:
     - python
     - setuptools
-    - openmm >=7.2
+    - openmm >=7.3
 
   run:
     - python
     - numpy
     - scipy
     - six
-    - openmm >=7.2
+    - openmm >=7.3
     - parmed
     - mdtraj
     - netcdf4
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -228,7 +228,7 @@ as a generic variable called ``lambda`` goes from ``1.0`` to ``0.0``.
 .. testcode::
 
     # Enslave lambda_sterics and lambda_electrostatics to a generic lambda variable.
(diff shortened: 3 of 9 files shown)


jchodera:
After this is merged, we should cut 0.17.0!

jchodera:
Quick timing update for c-Met with 133 states (29 different electrostatic discharging states):

Old code:
```
Computing energy matrix took    2.485s
```
New code:
```
Computing energy matrix took    0.413s
```
Quite impressive!

andrrizzi:
> Do we have to add the new classes (such as state.GlobalParameterFunction) to docs/states.rst for the API docs to be rendered?

I think you're right. I'll add them before merging.

> Does the `test_overlap()` test pass locally?

I'll run all the `alchemy.py` tests on the cluster to make sure.

> After this is merged, we should cut 0.17.0!

I'd wait for OpenMM 7.3 release before releasing this feature and test the dev openmmtools in the meantime. I'm back on debugging the NaNs today.

Also, I still would like to consider keeping the order of lambda consistent between different models with the addition of a `sqrt_lambda_electrostatics` parameter. I think the changes to the code would be minimal, just changing the name of the offset parameter and an addition to `AlchemicalState` like

```python
class AlchemicalState:
    ...
    _sqrt_lambda_electrostatics = GlobalParameter('sqrt_lambda_electrostatics', expression='sqrt(lambda_electrostatics)')
```

andrrizzi:
I'd work on it on a separate PR in any case.

jchodera:
@andrrizzi : Can this be merged and a new release cut?

andrrizzi:
I still have to run the full `test_alchemy.py` suite before merging. Also, I'd wait for NaN issue on OpenMM 7.3 to be fixed before making a release since we'll have to pin the minimum openmm version. That's my first priority coding-wise.

jchodera:
Thanks for clarifying!

andrrizzi:
I've run all the tests in `test_alchemy.py` and they passed. AppVeyor is failing because of problems with the win build of openmm7.3rc, but I think we can merge this for now.

jchodera:
Hooray!

jchodera:
@andrrizzi : Maybe we should change the YANK default for `alchemical_pme_treatment:` to `exact` now?

PR #382, opened 2018-10-27
Change name of step variable to allow reset to work

pgrinaway:

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1868,7 +1868,7 @@ class ExternalPerturbationLangevinIntegrator(NonequilibriumLangevinIntegrator):
         self.addComputeGlobal("perturbed_pe", "energy")
         # Assumes no perturbation is done before doing the initial MD step.
         self.beginIfBlock("step < 1")
-        self.addComputeGlobal("step", "1")
+        self.addComputeGlobal("step", "step + 1")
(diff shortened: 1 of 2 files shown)


pgrinaway:
Looks like some [weirdness](https://travis-ci.org/choderalab/openmmtools/jobs/447164641#L3431) in the build here....

jchodera:
@andrrizzi likely has insight into that segfault. It was previously claimed to be stochastic, but it seems to happen quite often.

jchodera:
I've restarted the build just in case.

openmmtools/integrators.py
```diff
@@ -1856,19 +1856,19 @@ def __init__(self, *args, **kwargs):
...
-        self.addGlobalVariable("first_step", 0)
+        self.addGlobalVariable("step", 0)
 
     def _add_integrator_steps(self):
         self.addComputeGlobal("perturbed_pe", "energy")
         # Assumes no perturbation is done before doing the initial MD step.
-        self.beginIfBlock("first_step < 1")
-        self.addComputeGlobal("first_step", "1")
+        self.beginIfBlock("step < 1")
+        self.addComputeGlobal("step", "1")
```
jchodera: To match the other integrators, why not increment `step` each step?

pgrinaway: ok

PR #384, opened 2018-11-06
Bugfix in Nose-Hoover chain integrator

Olllom:
@andysim and I realized that constraints had not been taken into account when calculating the number of degrees of freedom in the `NoseHooverChainVelocityVerletIntegrator`. 
As a result, the system temperature did not converge to the target value for systems with constraints.

I added a test to make sure that the target temperature is now matched.

The fix changes the API of the constructor: It requires the system to be passed as an argument (only in order to extract the correct number of constraints; the integrator then forgets about the system).

Cheers,
Andreas Krämer

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -585,6 +585,7 @@ class NoseHooverChainVelocityVerletIntegrator(ThermostatedIntegrator):
 
     Create a velocity Verlet integrator with Nosé-Hoover chain thermostat.
 
+    >>> from openmmtools import testsystems
     >>> waterbox = testsystems.WaterBox()
     >>> system = waterbox.system
     >>> timestep = 1.0 * unit.femtoseconds
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -585,7 +585,8 @@ class NoseHooverChainVelocityVerletIntegrator(ThermostatedIntegrator):
 
     Create a velocity Verlet integrator with Nosé-Hoover chain thermostat.
 
-    >>> system = testsystems.WaterBox()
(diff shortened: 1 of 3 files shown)


openmmtools/integrators.py
```diff
@@ -585,14 +585,15 @@ class NoseHooverChainVelocityVerletIntegrator(ThermostatedIntegrator):
 
     Create a velocity Verlet integrator with Nosé-Hoover chain thermostat.
 
+    >>> system = testsystems.WaterBox()
```
andysim: I think this should be called something like `waterbox` or `testsystem` and then the quantity passed into the c'tor below should be `waterbox.system`.  I could be wrong, but that's my memory of it.

andrrizzi:
Looks good to me! Could you fix the doctests and the regular tests that got broken with the API?

andrrizzi:
In particular, some tests in [test_integrators_and_testsystems.py](https://github.com/choderalab/openmmtools/blob/master/openmmtools/tests/test_integrators_and_testsystems.py#L94) and many tests in `test_integrators.py` (search for `integrator_class()`) assume that you can instantiate an Integrator without arguments, so if we don't make `system` an optional argument we need to create an exception instantiating it in the test code.

Olllom:
@andysim Thanks for the catch.

@andrrizzi 
Instead of adding an exception to all the tests, I made `system=None` a keyword argument of the constructor.
The integrator will fall back to the old behavior, which is correct for unconstrained systems. A severe warning will be printed, if the system is not specified.
- This avoids changing all the tests in test_integrator.py
- In contrast, an exception was added to test_integrator_and_testssystem.py - it actually makes of sense to test if the integrator plays nicely with all test systems.

PR #385, opened 2018-11-09
Fix bug when computing reduced potential in multiple states

andrrizzi:
This fixes a bug affecting ``reduced_potential_at_states()`` when computing the reduced potential of systems in different ``GlobalParameterState``s when the same alchemical parameter appeared in force objects split in different force groups.


jchodera:
Thanks! Can you comment what impact this might have had on existing calculations?

andrrizzi:
This has no impact on our calculations because the current implementation of `AbsoluteAlchemicalFactory` keeps the forces associated with the same alchemical parameter into the same force group, but I thought it would be wise to fix this before it bites.

PR #386, opened 2018-11-10
Set exact PME as default

andrrizzi:
- Set exact PME as the default value for `alchemical_pme_treatment` in the `AbsoluteAlchemicalFactory` constructor.
- I've added an optional `ignore_velocities` argument in the `__setstate__` and `__getstate__` methods of `SamplerState`. This will allow us to generalize in YANK the way MPI process send back the info about the `SamplerState` to the root process (including collective variables) without wasting bandwidth.
- I'm also checking if using the CPU platform to create `Context`s fixes the random segfaults in the Travis tests.


andrrizzi:
Thanks! Too bad that changing platform didn't work for the segfaults.

Issue #387, opened 2018-11-12
Integrators docs are broken


jchodera:
The [integrators docs page is broken](https://openmmtools.readthedocs.io/en/latest/integrators.html).

jchodera:
The [readthedocs build logs](https://readthedocs.org/projects/openmmtools/builds/8079152/) are filled with warnings like
```
WARNING: [autosummary] failed to import u'openmmtools.integrators.VVVRIntegrator': no module named openmmtools.integrators.VVVRIntegrator
```

PR #388, opened 2018-11-12
Fix integrators docs

jchodera:
Addresses #387.


jchodera:
Looks like this is fixed: [preview](https://openmmtools.readthedocs.io/en/fix-integrators-docs/integrators.html)

I wonder if there's a way we can test for this automatically...

PR #392, opened 2019-01-17
SamplerStates accepts lists of Vec3 as input

ajsilveira:

### DIFF
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -1750,11 +1750,11 @@ class SamplerState(object):
 
     Parameters
     ----------
-    positions : Nx3 simtk.unit.Quantity
+    positions : Nx3 simtk.unit.Quantity or list(simtk.openmm.Vec3)
         Position vectors for N particles (length units).
-    velocities : Nx3 simtk.unit.Quantity, optional
+    velocities : Nx3 simtk.unit.Quantity or list(simtk.openmm.Vec3), optional
         Velocity vectors for N particles (velocity units).
-    box_vectors : 3x3 simtk.unit.Quantity
+    box_vectors : 3x3 simtk.unit.Quantity or list(simtk.openmm.Vec3) 
         Current box vectors (length units).
 
     Attributes
@@ -1834,7 +1834,13 @@ class SamplerState(object):
         self._collective_variables = None
         self._kinetic_energy = None
         self._potential_energy = None
-        self._initialize(copy.deepcopy(positions), copy.deepcopy(velocities), copy.deepcopy(box_vectors))
+        args = []
+        for input in [positions, velocities, box_vectors]:
+            if isinstance(input, unit.Quantity) and isinstance(input._value, list):
+                args.append(np.array(input/input.unit)*input.unit)
+            else:
+               args.append(copy.deepcopy(input))    
+        self._initialize(*args)
 
     @classmethod
     def from_context(cls, context_state, ignore_collective_variables=False):
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -1750,11 +1750,11 @@ class SamplerState(object):
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


openmmtools/states.py
```diff
@@ -1750,11 +1750,11 @@ class SamplerState(object):
 
     Parameters
     ----------
-    positions : Nx3 simtk.unit.Quantity
+    positions : Nx3 simtk.unit.Quantity or list(simtk.openmm.Vec3)
```
andrrizzi: I'd revert changes this as it seems that we accept simple lists without units. You can maybe add in the description below that `Quantities` that use `List[Vec3]` values are supported, or actually any value that `np.array(input)` supports.

openmmtools/states.py
```diff
@@ -1834,7 +1834,13 @@ def __init__(self, positions, velocities=None, box_vectors=None):
         self._collective_variables = None
         self._kinetic_energy = None
         self._potential_energy = None
-        self._initialize(copy.deepcopy(positions), copy.deepcopy(velocities), copy.deepcopy(box_vectors))
+        args = []
+        for input in [positions, velocities, box_vectors]:
+            if isinstance(input, unit.Quantity) and isinstance(input._value, list):
```
andrrizzi: We can expand this and just convert on initialization everything to a numpy array, not just lists. I think Patrick also mentioned this would be useful. You can remove entirely the if-else in this case as doing `np.array()` basically creates a copy.

andrrizzi: > You can remove entirely the if-else

Or maybe substitute it with
```python
if not isinstance(input, np.ndarray):
    # Convert to numpy array
```
to avoid creating a copy of a big array when you don't need it.

ajsilveira: The problem is that  applying np.array() to a list of Vec3 with units, it gives an array of quantities:
>>> q = [(2, 4, 5), (8, 9, 5)]*simtk.unit.nanometer
>>> q
Quantity(value=[(2, 4, 5), (8, 9, 5)], unit=nanometer)
>>> print(np.array(q))
[[Quantity(value=2, unit=nanometer) Quantity(value=4, unit=nanometer)
  Quantity(value=5, unit=nanometer)]
 [Quantity(value=8, unit=nanometer) Quantity(value=9, unit=nanometer)
  Quantity(value=5, unit=nanometer)]]

ajsilveira: Yes, this was causing the problem with the reporter.

andrrizzi: Sorry! I actually meant this
```python
for input in [positions, velocities, box_vectors]:
    if not isinstance(input._value, np.ndarray):
        args.append(np.array(input/input.unit)*input.unit)
    else:
        args.append(copy.deepcopy(input))
```

ajsilveira: if input were not a quantity (None, for instance), the "if not" statement would give an attribute error, am I wrong?

ajsilveira: Yes. The problem was caused by using the mdtraj output of positions (via the function openmm_positions https://github.com/mdtraj/mdtraj/blob/34f77aaf767f27d486e8f4762c08fbfd5d3ea03e/mdtraj/core/trajectory.py#L1182 ) as the input of the SamplerState. That output is a list of Vec3s with units.

openmmtools/states.py
ajsilveira: I don't know what would be a good description.
positions : simtk.unit.Quantity of Nx3 or a lists of Vec3s. 
What do you think?

andrrizzi: The outer class of what is returned by `openmm_positions` is still a `Quantity` technically. It should be `Quantity(list(Vec3))` so I'd leave just `Quantity`. There are other types of `_value` that work (e.g., list of lists, tuple of lists, list of ndarray, etc.) so it'd be difficult to enumerate them all anyway.

PR #401, opened 2019-02-03
Update badges on README

jchodera:
This is a minor cleanup of the badges in the [README](https://github.com/choderalab/openmmtools/blob/cleanup-readme/README.md).


jchodera:
Failure is still the barostat segfault.

PR #406, opened 2019-03-15
Add property accessors for retrieving HostGuest* molecules as OEMols

### LINKED ISSUE #405: multistate API warning is printed even if multistate module is not imported`
```
>>> import openmmtools
/home/chodera/miniconda/lib/python3.6/site-packages/openmmtools-0.18.0-py3.6-linux-x86_64.egg/openmmtools/multistate/__init__.py:75: UserWarning: Warning: openmmtools.multistate API is experimental
  warnings.warn('Warning: openmmtools.multistate API is experimental')
```
Presumably we should only issue this warning if you are using any of the `MultiStateSampler` classes or subclasses, rather than when you simply import `openmmtools`.

jchodera:
This PR allows retrieval of the `HostGuest*` host and guest molecules as `OEMol`s if the OpenEye toolkit is installed. If it isn't, an exception will only be raised if the properties are accessed.

Examples:
```python
>>> from openmmtools import testsystems
/Users/choderaj/github/choderalab/openmmtools/openmmtools/openmmtools/multistate/__init__.py:75: UserWarning: Warning: openmmtools.multistate API is experimental
  warnings.warn('Warning: openmmtools.multistate API is experimental')
>>> t = testsystems.HostGuestVacuum()
>>> t.host_oemol
<oechem.OEGraphMol; proxy of <Swig Object of type 'OEGraphMolWrapper *' at 0x1a21edf6c0> >
>>> t.guest_oemol
<oechem.OEGraphMol; proxy of <Swig Object of type 'OEGraphMolWrapper *' at 0x1a21fc3b70> >
>>> t.oemols
[<oechem.OEGraphMol; proxy of <Swig Object of type 'OEGraphMolWrapper *' at 0x1a21edf6c0> >, <oechem.OEGraphMol; proxy of <Swig Object of type 'OEGraphMolWrapper *' at 0x1a21fc3b10> >]
```

This also include some other cleanup:
- [x] Update to unreleased 0.19.0
- [x] Remove after_success script in travis since we no longer push `openmmtools-dev`
- [x] Fixes #405 
- [x] Clean up `HostGuest*` API to use `handle_kwargs` for more flexible override support


jchodera:
@andrrizzi : I've added this capability to help us debug some issues with `perses`. Do you think it's worth adding the OpenEye encrypted license and testing infrastructure to test these accessors on travis?

openmmtools/__init__.py
```diff
@@ -8,6 +8,3 @@
 # Define global version.
 from openmmtools import version
 __version__ = version.version
-
-# Import modules.
-from openmmtools import testsystems, integrators, alchemy, mcmc, states, cache, utils, constants, forces, forcefactories, storage, multistate
```
andrrizzi: I still get confused by Python 3 import system, but I'm not sure this is harmless as it may cause this to fail in some cases
```python
import openmmtools
openmmtools.integrators.LangevinIntegrator  # raises exception
```
which, if it actually happens, would break a bunch of other packages. Alternatively, if the goal is to fix #406, we can move the warning inside the `MultiStateSampler`, `MultiStateReporter`, and `MultiStateAnalyzer` class as
```python
class MultiStateSampler:
    warning.warn(message, stacklevel=2)
```
I _think_ the warning would be raised also if you try to use a class that inherit from those, although I didn't try it.

openmmtools/multistate/multistatesampler.py
```diff
@@ -49,6 +49,12 @@
 
 logger = logging.getLogger(__name__)
 
+# ==============================================================================
+# Warn that API is experimental
+# ==============================================================================
+
+import warnings
+warnings.warn('Warning: openmmtools.multistate API is experimental')
```
andrrizzi: We can probably remove this if we have it in `multistate/__init__.py`

andrrizzi:
> Do you think it's worth adding the OpenEye encrypted license and testing infrastructure to test these accessors on travis?

Would be nice, but I don't think it's critical since OE usage is so restricted in this package. An alternative could be to just add a test and mark it as `slow` so that it will be run together with the comprehensive tests we manually start before releasing.

openmmtools/__init__.py
andrrizzi: Or maybe it would make sense to print that `warn` in the `__new__` magic method to make it support inheritance.

openmmtools/multistate/multistatesampler.py
jchodera: Whoops! I had meant to do that!

openmmtools/__init__.py
andrrizzi: Or also, we could selectively not import only `multistate`, or remove [these imports instead](https://github.com/choderalab/openmmtools/blob/master/openmmtools/multistate/__init__.py#L64-L70) after moving the warning inside the `mulstistate` submodules.

jchodera: @andrrizzi : This seems to work just fine:
```
>>> import openmmtools
/Users/choderaj/github/choderalab/openmmtools/openmmtools/openmmtools/multistate/__init__.py:77: UserWarning: Warning: openmmtools.multistate API is experimental
  warnings.warn('Warning: openmmtools.multistate API is experimental')
>>> openmmtools.integrators.LangevinIntegrator()
<openmmtools.integrators.LangevinIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x109bc4900> >
```
Does it really raise an exception for you?

jchodera: Ah, I see!

jchodera: > Or maybe it would make sense to raise that warn in the __new__

I tried that, but couldn't make it work.

> Or also, we could selectively not import only multistate

I think that would just make things even more confusing.

> remove these imports instead after moving the warning inside the mulstistate submodules.

We definitely don't want to do that, since the subpackage organization into files is arbitrary and we want to mask that.

andrrizzi: Let's try it out then! It could be possible that it changed sometime in one of the new Python 3 versions.

andrrizzi: Ah, sorry, just seen that you added it to `__init__`. +1

jchodera: Do we use the idiom 
```
import openmmtools
openmmtools.subpackage.class()
```
a lot? It seems like we shouldn't _expect_ this to work.

andrrizzi: I don't think so, but I think we do use `from openmmtools.subpackage import class` a lot instead of `from openmmtools.subpackage.module import class`.

PR #407, opened 2019-03-19
Add test for repex mixing with MPI

andrrizzi:
I'm opening the PR right now just to share the minimal test case that reproduces the error described in choderalab/yank#1130 (the code is [here](https://github.com/choderalab/openmmtools/blob/fix-mpi-mixing/openmmtools/tests/test_sampling.py#L1478-L1583)).

To run it with MPI, simply run `nosetests` with `mpiexec` as usual
```
build_mpirun_configfile "nosetests openmmtools/tests/test_sampling.py:TestReplicaExchange.test_uniform_mixing -s"
mpiexec.hydra -f hostfile -configfile configfile
```
I still have to figure out the reason, but at least this is a more manageable test case.


openmmtools/multistate/multistatesampler.py
```diff
@@ -1195,9 +1195,7 @@ def _propagate_replicas(self):
         # Update all sampler states. For non-0 nodes, this will update only the
         # sampler states associated to the replicas propagated by this node.
         for replica_id, propagated_state in zip(replica_ids, propagated_states):
-            propagated_positions, propagated_box_vectors = propagated_state  # Unpack.
-            self._sampler_states[replica_id].positions = propagated_positions
-            self._sampler_states[replica_id].box_vectors = propagated_box_vectors
+            self._sampler_states[replica_id].__setstate__(propagated_state, ignore_velocities=True)
```
jchodera: Why do you need to use `__setstate__` here?

jchodera: It would be important to document a change in internal data structures like this, since we can confuse `state` as in `SamplerState` as well as serialized state!

andrrizzi: There's no change in internal data structure. This just serializes the entire sampler state instead of just positions and box vectors (i.e. it serializes also the information on the collective variables so that they are transmitted to MPI node 0).

jchodera: Why can't we just do 
```python
self._sampler_states[replica_id] = propagated_state
```
?

andrrizzi: We can! But the `__get/setstate()__` methods have a `ignore_velocities` argument which allows us to save bandwidth and MPI would just call those methods anyway to serialize them before sending them over the network.

jchodera: I'd prefer the cleaner code option---if it happens under the hood automatically, we don't need to do it!

Also, we don't always randomize the velocities each iteration, so we only should be ignoring velocities _if_ we are going to be randomizing the velocities each iteration. Otherwise, we're introducing a significant error.

andrrizzi: To be clear, we don't send velocities only because we don't store velocities on disk. The `SamplerState` in the MPI process remains with velocities whether or not they are randomized.

The `__get/setstate__` methods are used to serialize things throughout the codebase, and they are effectively part of the `States`' API so they don't look weird to me. Besides being more efficient, I also think the code is more transparent about what is really happening here (i.e. we're serializing things and sending them, not just returning them), but I'm happy to change it if we think the loss in performance will be bearable.

andrrizzi: Just to make sure you caught it, I need `__setstate__` there because [this function](https://github.com/choderalab/openmmtools/blob/fix-mpi-mixing/openmmtools/multistate/multistatesampler.py#L1233) returns the `__getstate__`.

andrrizzi:
The issue still persist, but I'm going to merge this test in `master` so that it'll be easier to debug on the most up-to-date code.

I've opened separate issues about the serialization API (#448) and the still persisting mixing problem with MPI (#449).

PR #417, opened 2019-04-23
[WIP] Stage three sams

hannahbrucemacdonald:

Adding option for third-stage sams, where the third and final stage runs with fixed weights.
The stage update protocol involves both a criteria (logz/histogram/visits) and a threshold for 2nd stage.

If the third stage has the same criteria as second stage, a check is made during set up that the third-stage is a tighter (more stringent threshold criteria) during set up. If the third stage criteria differs from the second stage criteria then the thresholds cannot be compared, and instead, if during simulation, the third-stage criteria is already met when the second-stage criteria is met then the simulation will revert to a two-stage update scheme.


jchodera:
This looks like a good choice for API and YAML support for now! We can always generalize this later once we get things working robustly.

PR #420, opened 2019-04-29
Fixing FIREMinimizerIntegrator

hannahbrucemacdonald:

Previously NaN check prevented the integrator from taking any uphill steps, when the correct protocol should allow for n_min consecutive uphill steps.

Relaxing test criteria (tolerance and max_iterations) as test was erroneously passing before as ^ was causing delta_t to decrease rapidly


jaimergp:
I just realized that `jax-md` provides a [FIRE minimizer](https://github.com/google/jax-md/blob/master/jax_md/minimize.py#L101-L168) as well, so maybe we can also check against their implementation.

PR #421, opened 2019-05-10
Expanded ensemble

ajsilveira:
- Include the option weight_update to perform expanded ensemble simulations
- minimum_visits and beta_factor are defined by the user
- set histogram as the default for flatness criteria

### DIFF
--- a/openmmtools/multistate/sams.py
+++ b/openmmtools/multistate/sams.py
@@ -171,7 +171,7 @@ class SAMSSampler(multistate.MultiStateSampler):
                  log_target_probabilities=None,
                  state_update_scheme='global-jump',
                  locality=5,
-                 weights_update=True,
+                 weight_update=True,
                  update_stages='two-stage',
                  flatness_criteria='histogram-flatness',
                  flatness_threshold=0.5,
@@ -198,7 +198,7 @@ class SAMSSampler(multistate.MultiStateSampler):
             ``restricted-range`` will compute the probabilities for each of the states in the local neighborhood, increasing jump probability
         locality : int, optional, default=1
             Number of neighboring states on either side to consider for local update schemes.
-        weights_update : bool, optional, default=True
+        weight_update : bool, optional, default=True
              If False, the weight updating is disabled. This allows to perform expanded ensemble simulations.
         update_stages : str, optional, default='two-stage'
             One of ['one-stage', 'two-stage']
@@ -231,7 +231,7 @@ class SAMSSampler(multistate.MultiStateSampler):
         self.log_target_probabilities = log_target_probabilities
         self.state_update_scheme = state_update_scheme
         self.locality = locality
-        self.weights_update = weights_update
+        self.weight_update = weight_update
         self.update_stages = update_stages
         self.flatness_criteria = flatness_criteria
         self.flatness_threshold = flatness_threshold
@@ -293,7 +293,7 @@ class SAMSSampler(multistate.MultiStateSampler):
     log_target_probabilities = _StoredProperty('log_target_probabilities', validate_function=None)
     state_update_scheme = _StoredProperty('state_update_scheme', validate_function=_StoredProperty._state_update_scheme_validator)
     locality = _StoredProperty('locality', validate_function=None)
-    weights_update = _StoredProperty('weights_update', validate_function=None)
+    weight_update = _StoredProperty('weight_update', validate_function=None)
     update_stages = _StoredProperty('update_stages', validate_function=_StoredProperty._update_stages_validator)
     flatness_criteria = _StoredProperty('flatness_criteria', validate_function=_StoredProperty._flatness_criteria_validator)
     flatness_threshold = _StoredProperty('flatness_threshold', validate_function=None)
@@ -383,16 +383,16 @@ class SAMSSampler(multistate.MultiStateSampler):
         super()._restore_sampler_from_reporter(reporter)
         self._cached_state_histogram = self._compute_state_histogram(reporter=reporter)
         logger.debug('Restored state histogram: {}'.format(self._cached_state_histogram))
-        data = reporter.read_online_analysis_data(self._iteration, 'logZ', 'weights_update', 'stage', 't0', 'beta_factor')
+        data = reporter.read_online_analysis_data(self._iteration, 'logZ', 'weight_update', 'stage', 't0', 'beta_factor')
         self._logZ = data['logZ']
-        self.weights_update = str(data['weights_update'][0])
+        self.weight_update = str(data['weight_update'][0])
         self._stage = int(data['stage'][0])
         self._t0 = int(data['t0'][0])
         self.beta_factor= int(data['beta_factor'][0])
 
         # Compute log weights from log target probability and logZ estimate
         self._update_log_weights()
-        if (self.weights_update):
+        if (self.weight_update):
             # Determine t0
             self._update_stage()
 
@@ -401,7 +401,7 @@ class SAMSSampler(multistate.MultiStateSampler):
     def _report_iteration_items(self):
         super(SAMSSampler, self)._report_iteration_items()
 
-        self._reporter.write_online_data_dynamic_and_static(self._iteration, logZ=self._logZ, weights_update=self.weights_update,
+        self._reporter.write_online_data_dynamic_and_static(self._iteration, logZ=self._logZ, weight_update=self.weight_update,
                                                             stage=self._stage, t0=self._t0, beta_factor=self.beta_factor)
         # Split into which states and how many samplers are in each state
         # Trying to do histogram[replica_thermo_states] += 1 does not correctly handle multiple
@@ -444,7 +444,7 @@ class SAMSSampler(multistate.MultiStateSampler):
             swap_fraction_accepted = float(n_swaps_accepted) / n_swaps_proposed
         logger.debug("Accepted {}/{} attempted swaps ({:.1f}%)".format(n_swaps_accepted, n_swaps_proposed,
                                                                        swap_fraction_accepted * 100.0))
-        if (self.weights_update):
+        if (self.weight_update):
             # Update logZ estimates
             self._update_logZ_estimates(replicas_log_P_k)
 
--- a/openmmtools/multistate/sams.py
+++ b/openmmtools/multistate/sams.py
@@ -175,7 +175,7 @@ class SAMSSampler(multistate.MultiStateSampler):
                  update_stages='two-stage',
                  flatness_criteria='histogram-flatness',
                  flatness_threshold=0.5,
-                 mininum_visits=100,
+                 minimum_visits=100,
                  weight_update_method='rao-blackwellized',
                  beta_factor=0.8,
                  adapt_target_probabilities=False,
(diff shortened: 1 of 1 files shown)


openmmtools/multistate/sams.py
```diff
@@ -628,7 +648,6 @@ def _update_logZ_estimates(self, replicas_log_P_k):
         for (replica_index, state_index) in enumerate(self._replica_thermodynamic_states):
             logger.debug(' Replica %d state %d' % (replica_index, state_index))
             # Compute attenuation factor gamma
-            beta_factor = 0.8
```
andrrizzi: This may be the main cause of test errors. The local variable `beta_factor` is deleted here, but I don't see `self.beta_factor` used anywhere.

andrrizzi:
Thank you! I think this is ready to be merged. Can you just update the `releasehistory.rst` in `docs/` before?

ajsilveira:
This would be an "enhancement" in 0.19.0 version ?

andrrizzi:
Yes, that works.

ajsilveira:
@andrrizzi the last modifications are not working: I do not know how to write `validate_function` in `_StoreProperty` for the dict `criteria`.

openmmtools/multistate/sams.py
```diff
@@ -222,17 +234,23 @@ def __init__(self,
...
         self.gamma0 = gamma0
         self.logZ_guess = logZ_guess
         # Private variables
         # self._replica_neighbors[replica_index] is a list of states that form the neighborhood of ``replica_index``
         self._replica_neighbors = None
         self._cached_state_histogram = None
+        self.criteria = { criteria: value for criteria, value in (('flatness', flatness),  ('round_trips', round_trips),
+                                                                 ('minimum_logZ', minimum_logZ),  ('minimum_visits', minimum_visits))
+                                                                 if value is not None}
+        self.advance = {criteria: False for criteria, value in self.criteria.items() if value is not None}
```
andrrizzi: I don't think we need to expose this in the public API so I'd call this `_advance` instead. If we want to allow the users to force the next stage, we can instead expose the `self._stage` attribute.

openmmtools/multistate/sams.py
```diff
@@ -222,17 +234,23 @@ def __init__(self,
...
         self.weight_update_method = weight_update_method
+        self.beta_factor = beta_factor
         self.adapt_target_probabilities = adapt_target_probabilities
         self.gamma0 = gamma0
         self.logZ_guess = logZ_guess
         # Private variables
         # self._replica_neighbors[replica_index] is a list of states that form the neighborhood of ``replica_index``
         self._replica_neighbors = None
         self._cached_state_histogram = None
+        self.criteria = { criteria: value for criteria, value in (('flatness', flatness),  ('round_trips', round_trips),
```
andrrizzi: I'd do things a little differently here to simplify the code.

1. Save the `minimum_logZ`, `minimum_visits`, etc. independently as individual `_StoredProperty`. You don't really need a validator for these. The old validator was necessary to check the value of the `flatness_criteria` string, but this is no longer necessary as here you're hardcoding the possible criterias in the contructor arguments.
2. I'd implement `self._criteria` as private property so that it gets updated automatically when one of the criterias is changed by the user (i.e. something like this)
```python
@property
def _criteria(self):
    return { criteria: value for criteria, value in (('logZ_flatness', self.logZ_flatness), 
                                                     ('minimum_round_trips', self.minimum_round_trips),
                                                     ('minimum_logZ', self.minimum_logZ), 
                                                     ('minimum_visits', self.minimum_visits))
                                                     if value is not None}
```
This way if a user modifies `sams_sampler.minimum_visits`, then `sams_sampler._criteria` is automatically updated.

openmmtools/multistate/sams.py
```diff
@@ -254,12 +272,13 @@ def _update_stages_validator(instance, scheme):
             return scheme
 
         @staticmethod
-        def _flatness_criteria_validator(instance, scheme):
-            supported_schemes = ['minimum-visits', 'logZ-flatness', 'histogram-flatness']
-            if scheme not in supported_schemes:
-                raise ValueError("Unknown update scheme '{}'. Supported values "
-                                 "are {}.".format(scheme, supported_schemes))
-            return scheme
+        def _criteria_validator(instance, **kwargs):
```
andrrizzi: I think you can get rid of this with the new implementation.

openmmtools/multistate/sams.py
```diff
@@ -280,15 +299,19 @@ def _adapt_target_probabilities_validator(instance, scheme):
...
+    if self.update_stages == 'two-stage':
+        criteria = _StoredProperty('criteria', validate_function=_StoredProperty._criteria_validator)
     weight_update_method = _StoredProperty('weight_update_method', validate_function=_StoredProperty._weight_update_method_validator)
+    beta_factor = _StoredProperty('beta_factor', validate_function=None)
     adapt_target_probabilities = _StoredProperty('adapt_target_probabilities', validate_function=_StoredProperty._adapt_target_probabilities_validator)
     gamma0 = _StoredProperty('gamma0', validate_function=None)
     logZ_guess = _StoredProperty('logZ_guess', validate_function=None)
 
     def _initialize_stage(self):
+        self.round_trips = 0
```
andrrizzi: If you want to expose this in the public API, you'll have to move this initialization in `__init__`. Otherwise, I think this will crash
```python
sampler = SAMS()
sampler.round_trips  # Raise AttributeError
```
Alternatively, you can make it private, of course.

openmmtools/multistate/sams.py
```diff
@@ -280,15 +299,19 @@ def _adapt_target_probabilities_validator(instance, scheme):
...
+        self._downhill = False
```
andrrizzi: Could you add a comment about what this `_downhill` variable means?

openmmtools/multistate/sams.py
```diff
@@ -416,6 +441,15 @@ def _mix_replicas(self):
             else:
                 raise Exception('Programming error: Unreachable code')
 
+        n_replica, n_states = self.n_replicas, self.n_states
+        for replica_index, current_state_index in enumerate(self._replica_thermodynamic_states):
+            if self._downhill:
```
andrrizzi: Maybe a short comment about what's happening here would be helpful too.

openmmtools/multistate/sams.py
ajsilveira: What about :
```
        # The variable _downhill keeps track of which end state the walker has visited more recently.
        # This is used for computing the number of round trips.
        # _downhill = False: the walker is going uphill because it has visited the initial state more recently than the final state
```

andrrizzi: Sounds great, thanks!

openmmtools/multistate/sams.py
```diff
@@ -171,10 +172,14 @@ def __init__(self,
                  log_target_probabilities=None,
                  state_update_scheme='global-jump',
                  locality=5,
-                 update_stages='two-stage',
-                 flatness_criteria='logZ-flatness',
-                 flatness_threshold=0.2,
+                 weight_update=True,
                  weight_update_method='rao-blackwellized',
+                 update_stages='two-stage',
+                 flatness = None,
```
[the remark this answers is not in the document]
ajsilveira: `flatness` refers to the histogram so I replaced it by `histogram_flatness`. I also replaced `logZ_flatness` by `minimum_logZ` because the corresponding criteria is satisfied if a minimum logZ value is attained and this is not related to the flatness in logZ.

andrrizzi:
That Travis error is very weird, but we've switched our building system to the Cookiecutter scheme and your fork got out of date so I'd try a [sync](https://help.github.com/en/articles/syncing-a-fork).

openmmtools/multistate/sams.py
```diff
@@ -171,10 +172,14 @@ def __init__(self,
...
                  locality=5,
-                 update_stages='two-stage',
-                 flatness_criteria='logZ-flatness',
-                 flatness_threshold=0.2,
+                 weight_update=True,
                  weight_update_method='rao-blackwellized',
+                 update_stages='two-stage',
+                 histogram_flatness = None,
+                 minimum_visits = None,
+                 minimum_round_trips = None,
```
andrrizzi: I think the errors are due to a bunch of tests that fail because right now by default no condition for switching stage is set. Maybe we can change the defaults to use our "best practice" criteria?

andrrizzi: You can ignore the MacOS failures btw. Only linux is working right now.

ajsilveira: Ok. Could I set the histogram flatness as the default?

andrrizzi: Yeah, that sounds great.

ajsilveira: When running pytest, I am getting the following error (test_sampling.py:1060: AssertionError) for all samplers:
```
E                 Full diff:
E                 {'_adapt_target_probabilities': False,
E                 '_beta_factor': 0.8,
E                 '_downhill': False,
E                 '_gamma0': 1.0,
E                 '_have_displayed_citations_before': False,
E                 '_histogram_flatness': 0.3,
E                 '_iteration': 3,
E                 -  '_last_err_free_energy': inf,
E                 ?                           ^^^
E                 +  '_last_err_free_energy': masked,
E                 ?                           ^^^^^^
```

andrrizzi: We're actually still using `nose` with openmmtools. We've not converted all the tests to `pytest` yet. Not sure if this is going to help, but it's worth the shot.

docs/releasehistory.rst
```diff
@@ -1,15 +1,23 @@
...
-0.19.0
-======
+0.18.2 - Bugfix release
+=======================
 
-New features
+Enhancements
 ------------
+- ``SAMSSampler`` supports expanded ensemble simulations
+- ``SAMSSampler`` accepts multiple criteria for switching the stage
```
andrrizzi: I think something went wrong when resolving conflicts and these got moved in 0.18.2, which has been already released. Can you move these back to a new 0.19.0 section?

openmmtools/multistate/sams.py
```diff
@@ -234,6 +250,17 @@ def __init__(self,
...
+    @property
+    def _criteria(self):
+        return { criteria: value for criteria, value in (('histogram_flatness', self.histogram_flatness),
+                                                         ('minimum_round_trips', self.minimum_round_trips),
+                                                         ('minimum_logZ', self.minimum_logZ),
+                                                         ('minimum_visits', self.minimum_visits))
+                                                         if value is not None}
+    @property
+    def _advance(self):
+        return {criteria: False for criteria, value in self._criteria.items() if value is not None}
```
andrrizzi: I'm not sure this is going to work. Here you're regenerating an all-false dictionary every time you call `self._advance` so doing `self._advance[criteria] = True` won't actually change it.

Actually, if I understand the code correctly, you can even remove this as an object variable/property and simply initialize an `advance` local variable inside `_update_stage()` before checking all criteria.

ajsilveira: After merging master, the version that it is being installed is 0.16, is this correct?

andrrizzi: Hmm. Are you having those problems locally? On Travis, it looks like versioneer is installing 0.18.2 as expected: https://travis-ci.org/choderalab/openmmtools/jobs/542376730#L884

ajsilveira: Yes, on lilac.

andrrizzi: I see, if you're installing from source, it shouldn't matter which version is printed as long as the installed code is updated. Unless I'm misunderstanding the problem.

PR #424, opened 2019-05-23
Update everything to match MolSSI cookiecutter

jchodera:
This PR overhauls the testing infrastructure to match the current [MolSSI cookiecutter](https://github.com/MolSSI/cookiecutter-cms).

`logsumexp` has been moved form `scipy.misc` to `scipy.special`.

cc: https://github.com/choderalab/yank/pull/1162


jchodera:
The one deviation (I know of) is that I'm still using `nosetests` here since it would take considerable effort to convert all the test generators over to `pytest` right now.

jaimergp:
I haven't tried, but `pytest` [should be able to run](https://docs.pytest.org/en/latest/nose.html) `nosetests` just fine.

andrrizzi:
The main problem for the conversions will be the many nose generators using `yield`. As far as I understand, `pytest`'s support for those is shaky.

devtools/conda-envs/test_env.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - libnetcdf >=4.6.2 # workaround for libssl issues
+  - pyyaml
+  - cython
+  - sphinxcontrib-bibtex
+  - mpiplus
+  - pymbar
+  - pyyaml
+
+    # Testing
+  - nose
```
andrrizzi: I think we'll have to add also `nose-timer` here to get the `--with-timer` option working.

jchodera:
@andrrizzi : Interestingly, we're now seeing segfaults in a different place!
```
Test method AlchemicalState.apply_to_context. ... python(2864,0x7fff8a436340) malloc: *** error for object 0x7fb943705c38: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

PR #426, opened 2019-05-29
Versioneer and improve netcdf opening robustness

andrrizzi:
I've updated `setup.py` to use `versioneer` using the cookiecutter system. If I understand correctly we won't need to bump manually the version number twice every time we do a release after this. Let's see if tests pass after this.


andrrizzi:
It looks like there's a systematic segfaulting that pops up with the MacOS tests. I'll check if this can be reproduced locally.

andrrizzi:
Looks like this is happening only on Travis. Any objection if I merge this and carry on with the bugfix release? I don't think I'll have a lot of time to dig into this in the next few days.

andrrizzi:
I'll try to run the full test suite locally on my Mac in any case before releasing.

andrrizzi:
All tests pass locally. I've also modified our "robust" opening of the netcdf file to set the environment variable `HDF5_USE_FILE_LOCKING` to `'FALSE'` in the last attempt to open the file. Let's see if this improve things (see choderalab/yank#1165).

I'll merge when tests pass and make the release.

PR #430, opened 2019-07-19
FIx module path

andrrizzi:
A simple bugfix for a function that was incorrectly imported from `openmmtools.utils` instead of `openmmtools.multistate.utils`.


andrrizzi:
Thank you!

PR #431, opened 2019-07-20
WIP Alchemical Custom Forces + Custom Lennard-Jones Switching

Olllom:
## Description
This pull request contains two features that are largely independent:
1) An extension of the alchemical facility to work with custom forces, especially NBFIXes coming from OpenMM's PSF parser.
2) Custom van-der-Waals switching functions, in particular, CHARMM's VSWITCH and VFSWITCH.

If you would like to have one or both in openmmtools, feel free to take it. I am happy to amend it where needed.

### 1) Alchemical Functions
- added methods `_alchemically_modify_CustomNonbondedForce` and `_alchemically_modify_CustomBondForce` to `AlchemicalFactory`. These methods identify Lennard-Jones forces based on their parameters (`sigma`, `epsilon`) and tabulated functions (`acoeff`, `bcoeff`) names. If LJ forces are identified, they are automatically augmented by a softcore potential. (If you find that check too loose, I could also check for specific energy functions.) By this change, the `AlchemicalFactory` will handle all LJ forces coming from `openmm.app.CharmmPsfFile.createSystem` as well as ones that have been augmented by switching functions (see 2) ).
- Added functionality to facilitate alchemical manipulation of other custom forces: a function `modify_custom_force` that can add lambda variables and softcore potentials to an existing energy string, or simply replace the energy string by a user-defined alchemical transformation. `AlchemicalState` has been modified to pick up user-defined lambda variables `lambda_...`
- Added a testsystem from CHARMM files (small molecule in salt water).

### 2) Van-der-Waals Switching Functions
Important, for example, for membrane simulations using the C36 force field. 
- CHARMM-GUI provides a VFSWITCH for OpenMM, but it may not be bad to have a similar implementation in openmmtools, too.
- Support for custom, user-defined switching functions.

## Todos
- [ ] Profiling of VSWITCH and VFSWITCH.
- [ ] Compare VSWITCH and VFSWITCH energies with CHARMM output. 
- [ ] Add documentation and example of modifying custom forces to AlchemicalFactory, AlchemicalState.

## Status
- [ ] Ready to go


jchodera:
@Olllom : Thanks so much for this contribution! At first glance, it looks very much like how I'd implement this scheme.

@andrrizzi : Can you take a look on Monday and give some feedback?

Relatedly, I've been thinking about how we might modularize the choice of softcore functional forms in the future, especially given new efficient forms that have been proposed in addition to the generalized form from [Michael Shirts](https://aip.scitation.org/doi/abs/10.1063/1.3607597) we currently support:
* This [linear version](https://www.mpibpc.mpg.de/277406/Buelens_2011_JCC_33_25-33.pdf) from Helmut Grubmüller
* An exciting improvement from [Bert de Groot](https://pubs.acs.org/doi/abs/10.1021/ct300220p)

We can always perform this generalization later, after this PR is merged.

andrrizzi:
Thank you! Will take a look in the next few days.

> I've been thinking about how we might modularize the choice of softcore functional forms

I've also worked on incapsulating some logic into `Force` objects to simplify the factory. Right now I have the nonbonded forces implemented, but I need to clean up docstrings and tests before pushing it.

Olllom:
Andrea, Jaime, and I talked about this PR briefly earlier this week. We were wondering if somebody other than @andrrizzi could offer a (preliminary) review. I could also break this into smaller chunks if that helps. The new `CharmmSolvated` testsystem is straightforward.  The switching functions should be easy, too.

The latest commits include 

1. a test with respect to CHARMM energies for the VSWITCH and VFSWITCH LJ switching functions
2. a test with respect to CHARMM energies for the alchemical endstates
3. a rebase to the current master

The CHARMM reference calculations are hosted [here](https://github.com/Olllom/charmm-vs-openmm-energies/tree/master/switching).

Since we are now mentioning this PR in a paper it would be good if we could move towards merging it into master.

Best,
Andreas

andrrizzi:
Hey Andreas! Thanks for updating the PR with `master`. If nobody gets to this, I'll spare a couple of hours during the weekend to review it.

openmmtools/alchemy.py
```diff
@@ -231,6 +233,9 @@ def from_system(cls, system, *args, **kwargs):
             If specified, the state will search for a modified
             version of the alchemical parameters with the name
             ``parameter_name + '_' + parameters_name_suffix``.
+        custom_lambdas : list of str, optional
```
andrrizzi: I'd prefer not to expose this argument. The reason is simply that dynamically-created classes do not behave nicely in many cases (serialization above all), which we'll have to support if we expose this feature. The encouraged way to extend this is through (static) class extension
```python
class MyAlchemicalState(AlchemicalState):
   custom_param1 = AlchemicalState._LambdaParameter(...)
   custom_param2 = AlchemicalState._LambdaParameter(...)
```
If your application requires the dynamic instantiation, could you wrap these lines in a function in your code? I don't think that a `MultiStateSampler` would currently be able to resume correctly if you'll pass one of these classes to it though.

openmmtools/alchemy.py
```diff
@@ -2038,6 +2060,112 @@ def _alchemically_modify_AmoebaMultipoleForce(self, reference_force, alchemical_
             alchemical_region = alchemical_region[0]
         raise Exception("Not implemented; needs CustomMultipleForce")
 
+    def _alchemically_modify_CustomBondForce(self, reference_force, alchemical_regions, _):
+        energy_function = reference_force.getEnergyFunction()
+
+        # this is a quick fix.
+        # TODO: extend to multiple alchemical regions
+        assert len(alchemical_regions) == 1
```
andrrizzi: Could you raise a `NotImplementedError` here rather than an assert explaining this is not currently supported?

openmmtools/alchemy.py
```diff
@@ -2038,6 +2060,112 @@ def _alchemically_modify_AmoebaMultipoleForce(self, reference_force, alchemical_
...
+        # Alchemically modify
+        modified_force = copy.deepcopy(reference_force)
+
+        if is_custom_lj_force(reference_force):
+            softcore_energy_string = modify_vdw_energy_string_softcore(
+                energy_string=energy_function, do_annihilate=alchemical_region.annihilate_sterics,
+                sigma_string="sigma", bonded_force=True
+            )
+            terms = softcore_energy_string.split(";")
+            softcore_energy_string = ";".join(terms)
```
andrrizzi: Does this have any effect since `terms = softcore_energy_string.split(";")`?

openmmtools/alchemy.py
```diff
@@ -2038,6 +2060,112 @@ def _alchemically_modify_AmoebaMultipoleForce(self, reference_force, alchemical_
...
+        else:
+            return {'': [copy.deepcopy(reference_force)]}
+
+    def _alchemically_modify_CustomNonbondedForce(self, reference_force, alchemical_regions, _):
+
+        energy_function = reference_force.getEnergyFunction()
+
+        # this is a quick fix
+        # TODO: extend to multiple alchemical regions
+        assert len(alchemical_regions) == 1
```
andrrizzi: Same here.

openmmtools/alchemy.py
```diff
@@ -2038,6 +2060,112 @@ def _alchemically_modify_AmoebaMultipoleForce(self, reference_force, alchemical_
...
+                energy_string=energy_function, do_annihilate=alchemical_region.annihilate_sterics,
+                sigma_string="0.5*(sigma1+sigma2)"
+                # we just recalculate sigma since this is certainly not the bottleneck
+            )
+            modified_force.setEnergyFunction(softcore_energy_string)
+        else:
+            # No fitting modification found
+            return {'': [copy.deepcopy(reference_force)]}
+
+        if is_custom_lj_force(reference_force) or is_nbfix_force(reference_force):
```
andrrizzi: I think this if-clause is always `True`, so you could remove one level of indentation here.

openmmtools/alchemy.py
```diff
@@ -2502,6 +2630,283 @@ def check_energy_expression(custom_force, parameter):
...
+        The scale factor that will be multiplied with the original energy function. This scale will usually include
+        lambda parameters.
+    effective_radius_string : str
+        The energy expression that is used for the effective radius in the softcore potential.
+    full_energy_string : str
+        When redefining the complete energy string, pass full_energy_string. This overwrites any energy expressions
+        passed as scale_string or effective_radius_string.
+    kwargs:
+        Any additional global parameters that the force depends on. Parameters that begin with "lambda_" will
+        be added to the alchemical state upon creation.
```
andrrizzi: I don't think this happens automatically currently. Maybe add that "will have to be added"?

openmmtools/alchemy.py
```diff
@@ -2502,6 +2630,283 @@ def check_energy_expression(custom_force, parameter):
...
+            bonded_force=isinstance(force, openmm.CustomBondForce))
+    force.setEnergyFunction(full_energy_string)
+
+    # add global parameters
+    for arg in kwargs:
+        if is_variable_in_energy_string(full_energy_string, arg):
+            force.addGlobalParameter(arg, kwargs[arg])
+            if not arg.startswith("lambda_"):
+                if hasattr(alchemical_region, arg) and getattr(alchemical_region, arg) != kwargs[arg]:
+                        warnings.warn("Argument {} in alchemical region was () but is getting overwritten "
```
andrrizzi: I think the `()` should be `{}`.

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,366 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+            return True
+    for i in range(nonbonded_force.getNumExceptions()):
+        particle1, particle2, qq, sig, eps = nonbonded_force.getExceptionParameters(i)
+        vdw = (eps.value_in_unit(unit.kilojoule_per_mole) * sig.value_in_unit(unit.nanometer))
+        if vdw > 1e-5:
+            return True
+    return False
+
+
+def get_forces_that_define_vdw(system):
```
andrrizzi: Maybe rename this `find_vdw_forces` for consistency with `forces.find_forces`?

andrrizzi: It looks like this function encapsulates some logic that is also exposed in `alchemy.is_nbfix_force` and `is_custom_lj_force`. Does it make sense to remove the redundant logic?

openmmtools/tests/test_alchemy.py
```diff
@@ -2325,9 +2325,314 @@ def test_serialization(self):
...
+
+def test_is_nbfix_force():
+    charmm_solvated = testsystems.CharmmSolvated()
+    nbfix_forces = [charmm_solvated.system.getForce(i)
+                    for i in range(charmm_solvated.system.getNumForces())
+                    if is_nbfix_force(charmm_solvated.system.getForce(i))]
+    assert len(nbfix_forces) == 1
+
+
+def endstate_test(use_vswitch=False, pme_tol=1e-5, platform=openmm.Platform.getPlatformByName("Reference")):
```
andrrizzi: We usually use `check_endstates` for this kind of non-test functions.

openmmtools/tests/test_alchemy.py
```diff
@@ -2325,9 +2325,314 @@ def test_serialization(self):
...
+    Test that the energies in the end states of the alchemical calculation are consistent
+    with energies that do not use the alchemical forces.
+    """
+    testsystem = testsystems.CharmmSolvated(ewald_tolerance=pme_tol)
+    if use_vswitch:
+        forcefactories.use_custom_vdw_switching_function(testsystem.system, 1.0*unit.nanometer, 1.2*unit.nanometer)
+
+    # utility functions
+    def calc_energy(testsystem, platform=platform):
+        dummy_integrator = openmm.VerletIntegrator(1.0*unit.femtosecond)
```
andrrizzi: Could you implement this using the existing `test_alchemy.compute_energy` function in here?

Issue #429, opened 2019-07-17
Split reporter storage over multiple files


andrrizzi:
I'm planning to make some changes in the `Reporter` to split the two monolithic `netcdf` files into more manageable chunks. This is what I'm thinking now:

- Solute trajectory: One `xtc` file per replica.
- Checkpoint trajectory: One `xtc` file per replica.
- Thermodynamic states: YAML (for thermodynamic parameters) and XML (for standard `System`) files.
- MCMC moves: One YAML file per move.
- `metadata`: One or many YAML files.
- Constructor options: One or many YAML files.
- everything else (energies, mixing statistics, logZ, `any_numeric_array_of_variable_dimension`): a single `netcdf` file for all.

I think splitting over multiple small files (whose directory structure is hidden by the `Reporter` class) means reading operations will be faster. Moreover, we'll be able to parallelize writing on disk, which is currently a big bottleneck for multi-replica methods.

**Question:** Should we keep the old reporter around and maybe change its name to `NetCDFReporter` to allow reading data generated with previous versions or do we anticipate installing a previous version of `openmmtools`/`yank` will suffice for our needs?

sroet:
I do not use this personally, so no strong opinion either way.

I did have a small remark on the proposal:

> Checkpoint trajectory: One xtc file per replica.

You probably do not want to use a lossy format like `xtc` (stores positions to about 1e-3 A) for checkpointing trajectories.

andrrizzi:
> You probably do not want to use a lossy format like `xtc` (stores positions to about 1e-3 A) for checkpointing trajectories.

Thank you so much for your input, @sroet. I wasn't aware of that, and I think it's a fair concern. We may want to save the checkpoint in `dcd` instead then.

We also need to think about how to handle the restraint reweighting which relies on the solute trajectory. A solution that would let us store it in low-precision would be to add the option to compute and store at runtime distances and energies. Actually, this feature was transferred here from YANK with the `multistate` module, but maybe we want to handle that only in YANK to keep things simpler here.

kyleabeauchamp:
1.  Are there parallel backends for netCDF that would also allow parallelism without a refactor?
2.  You might want to copy some ideas from xarray, which is now used by arviz as a backend for MCMC trace storage---it won't have exactly what you need, but maybe the overall structure would be inspring.  https://arviz-devs.github.io/arviz/notebooks/XarrayforArviZ.html

jchodera:
[...]

PR #434, opened 2019-07-25
Write sliced trajectories to XTC files

jaimergp:
## Description

Following the conversation at #429, this PR adds support for writing and reading (sliced) sampler states to `xtc` files using `mdtraj` IO helpers.

To do that, I have modified `openmmtools.multistate.multistatereporter.MultiStateReporter.{write,read}_sampler_states` so they use two new methods:

- `._write_sampler_states_to_xtc()`. Writes to a `<prefix>_replica.<replica_index>.xtc` file per `StateSampler` in the given iteration. `mdtraj.formats.XTCFileTrajectory` does not support `append` mode, so we have to binary concatenate a temporary `xtc` file to work around that, resulting in two write operations. This might not be ideal for performance, so we might need to add real `a` support if the benchmarks don't show an improvement.
- `._read_sampler_states_from_xtc()`. Traverses the `<prefix>_replica.*.xtc` files and generates a `SamplerState` for each requested frame/iteration index.

## Todos
- [x] Benchmark performance
- [X] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)
    - Note: The existing tests already hit the new methods, but maybe we need to add more subtleties (box vectors, etc).
- [ ] Update [changelog](https://github.com/choderalab/openmmtools/blob/master/docs/releasehistory.rst) with notable points that this PR has either accomplished or will accomplish.

## Status
- [ ] Ready to go

### DIFF
--- a/examples/trajectory-io-benchmarks/run.sh
+++ b/examples/trajectory-io-benchmarks/run.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+git checkout master > /dev/null
+python trajectory_io_benchmarks.py benchmark_master
+
+echo ""
+echo "****"
+echo ""
+git checkout multistate-xtc-traj > /dev/null
+python trajectory_io_benchmarks.py benchmark_multistate-xtc-traj
--- a/examples/trajectory-io-benchmarks/trajectory_io_benchmarks.py
+++ b/examples/trajectory-io-benchmarks/trajectory_io_benchmarks.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+
+"""
+Benchmark the performance of several trajectory output strategies.
+
+This script will test the performance of the strategy proposed in PR #434
+against the default monolythic NetCDF file.
+"""
+
+from argparse import ArgumentParser
+import time
+
+import numpy as np
+from openmmtools.tests.test_sampling import TestReporter
+from openmmtools import testsystems, states
+from simtk import unit
+
+import warnings
+warnings.simplefilter("ignore", UserWarning)
+
+
+def run_simulation(
+    test_system=testsystems.AlanineDipeptideVacuum,
+    n_states=10,
+    n_iterations=10,
+    analysis_particle_indices=list(range(1, 10)),
+    checkpoint_interval=2,
+):
+    """
+    Writes n_states SamplerStates over n_iterations to disk
+    and read them back to back.
+
+    Returns a tuple of two timings: write and read, per iteration and state
+    """
+    with TestReporter.temporary_reporter(
+        analysis_particle_indices=analysis_particle_indices,
+        checkpoint_interval=checkpoint_interval,
+    ) as reporter:
+        test = test_system()
+        positions = test.positions
+        box_vectors = unit.Quantity(
+            [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]], unit=unit.nanometer
+        )
+        sampler_states = [
+            states.SamplerState(positions=positions, box_vectors=box_vectors)
+            for _ in range(n_states)
+        ]
+        t0 = time.time()
+        for iteration in range(n_iterations):
+            reporter.write_sampler_states(sampler_states, iteration=iteration)
+        reporter.write_last_iteration(iteration)
+        t1 = time.time()
+        for iteration in range(n_iterations):
+            restored_sampler_states = reporter.read_sampler_states(iteration=iteration, analysis_particles_only=True)
+        t2 = time.time()
+
+    return (t1 - t0) / (n_states * n_iterations), (t2 - t1) / (n_states * n_iterations)
+
+def main():
+    tests = (
+        {
+            'test_system': testsystems.AlanineDipeptideVacuum,
+            'n_states': 10,
+            'n_iterations': 10,
+            'analysis_particle_indices': list(range(10)),
+            'checkpoint_interval': 20
+        },
+        {
+            'test_system': testsystems.SrcImplicit,
+            'n_states': 10,
+            'n_iterations': 10,
+            'analysis_particle_indices': list(range(2000)),
+            'checkpoint_interval': 20
+        },
+        {
+            'test_system': testsystems.DHFRExplicit,
+            'n_states': 10,
+            'n_iterations': 10,
+            'analysis_particle_indices': list(range(10000)),
+            'checkpoint_interval': 20
+        }
+    )
+    print('Timings per iteration and state, averaged over three attempts')
+    for options in tests:
+        read_timings, write_timings = [], []
+        for _ in range(3):
+            write, read = run_simulation(**options)
+            read_timings.append(read)
+            write_timings.append(write)
+        print(" ->", options['test_system'].__name__, f"({len(options['analysis_particle_indices'])}-atom subset)",
+              "\n\tW:", np.mean(write_timings), "+-", np.std(write_timings),
+              "\n\tR:", np.mean(read_timings), "+-", np.std(read_timings))
+
+
+if __name__ == '__main__':
+    main()
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -96,6 +96,15 @@ class MultiStateReporter(object):
 
         The reporter internally tracks what data goes into which file, so its transparent to all other classes
         In the future, this will be able to take Storage classes as well
+
+    sliced_trajectory_storage : str or None, optional
+        Optional name of the sliced trajectory file, as provided in ``analysis_particle_indices``. This file is used to
+        save the sliced trajectory in XTC format.
+
+        This should NOT be a full path, but just the desired placeholder filename (replica indices will be appended to the
+        suggested name). If None is provided, the derived name will be the same as ``storage``, but adding ``_replica``
+        before the ``.xtc`` extension.
+
     analysis_particle_indices : tuple of ints, Optional. Default: () (empty tuple)
         Indices of particles which should be treated as special when manipulating read and write functions.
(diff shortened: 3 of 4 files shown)


jaimergp:
Benchmarks are still pending, but in case we need to improve performance, some ideas:

- A. Write the temporary `xtc` to `/dev/shm` when available. Depending on the available RAM and the system size, this could be unfeasible, but I don't know if a single positions array (one frame) will pose such a problem.
- B. Use a ton of individual files and concatenate only every `N` frames, or at the end of the calculation.
- C. Implement `a` mode in `XTCFileTrajectory`. Apparently, the `xdrlib` [supports this mode](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/xtc/src/xdrfile.c#L203-L206), but [it's not exposed in the Python wrapper](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/xtc/xtc.pyx#L275). It should be [relatively easy](https://github.com/mdtraj/mdtraj/issues/1313#issuecomment-347982625).

jaimergp:
Added some benchmarks comparing against `master`. Although write operations are slower as expected, we are having more trouble reading (!) the data back. This is my laptop (WSL, which might be slower in IO, but still applies for the relative differences).

The script is located at `/examples/trajectory-io-benchmarks`. I have used git checkout to change between branches, which would require the `.py` file to be git untracked if you want it to survive the checkouts.

```
Switched to branch 'master'
Timings per iteration and state, averaged over three attempts
 -> AlanineDipeptideVacuum (10 atoms)
        W: 0.00021750926971435549 +- 2.392853074589234e-05
        R: 0.0005726496378580729 +- 4.084119558895955e-05
 -> SrcImplicit (2000 atoms)
        W: 0.0008469923337300619 +- 0.00039442946856827886
        R: 0.0005366365114847819 +- 2.7153077114797376e-05
 -> DHFRExplicit (10000 atoms)
        W: 0.0029479670524597164 +- 0.00015016260792103403
        R: 0.000729363759358724 +- 7.63090047609842e-06

****

Switched to branch 'multistate-xtc-traj'
Timings per iteration and state, averaged over three attempts
 -> AlanineDipeptideVacuum (10 atoms)
        W: 0.003030669689178467 +- 7.005560284900752e-05
        R: 0.001007688840230306 +- 1.0001335781663186e-05
 -> SrcImplicit (2000 atoms)
        W: 0.0035883164405822753 +- 8.677391591527427e-05
        R: 0.0688159728050232 +- 0.00560091105373269
 -> DHFRExplicit (10000 atoms)
        W: 0.010596606731414793 +- 0.00020410040438298147
        R: 0.3884154311815898 +- 0.043372317409459606
```

jaimergp:
I suspected that we might be wasting some precious time when converting to `unit.Quantity` objects, and apparently we are. I have attached the profiling results for both `master` and this PR (`xtc.prof`).

#  master

Writing:
image

Reading:
image

***

# This PR

Writing:

We seem to spend a long time initializing (and removing) the temporary directories. Actual writing seems to be at ~0.1s, if I am reading the timings correctly.

image

Reading:

There must be something very wrong here with [how I initialize](https://github.com/choderalab/openmmtools/pull/434/files#diff-372aa17c69074986d5586b694a40c199R1773) `Quantity` objects.

image

***

I used `snakeviz` to visualize the profiles:
[profiling.zip](https://github.com/choderalab/openmmtools/files/3447471/profiling.zip)

andrrizzi:
This is awesome!

I think you are right and the read operation is slowed down by the absence of random access. This might be mitigated if we need to read them in sequence, which is probably the most common use case.

I'm more worried about the extra order of magnitude in writing. Implementing appending is probably the way to go. Possibly, keeping the xtc files open and closing them on `__del__` might also help, although this is probably a very small improvement.

If we can't make `xtc` writing faster for a single MPI process, we could think of having multiple AMBER-style netcdf solute trajectories instead of xtc.

andrrizzi:
The time we spend to copy the positions is also a good catch! I wonder if that happens [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/states.py#L2113-L2118) when we slice the positions to obtain only the positions of the solute atoms. If that's the case, I'd be +1 on adding the extra argument `__getitem__(item, copy=True)` that allows toggling the copy to improve performance.

andrrizzi:
> There must be something very wrong here with how I initialize Quantity objects

Yeah, I seem to remember the `alanine_dipeptide.positions` variable being of the worst possible type. It may even be a list of `Quantity`. You probably want to convert it to a quantity-wrapped numpy array/list before starting the benchmark because I remember that can throw timings off.

jaimergp:
> Yeah, I seem to remember the alanine_dipeptide.positions variable being of the worst possible type. It may even be a list of Quantity. You probably want to convert it to a quantity-wrapped numpy array/list before starting the benchmark because I remember that can throw timings off.

But then we would also have that problem when benchmarking `master`. What's different in terms of `SamplerState` initialization between both branches?

jaimergp:
# Good news!

That last commit addresses the `read` bottleneck. `unit.Quantity` spent a lot of time trying to guess the value / unit of the passed list. Being explicit about helps... _a lot_:

## Master

```
Timings per iteration and state, averaged over three attempts
 -> AlanineDipeptideVacuum (10-atom subset) 
	W: 0.0004067365328470866 +- 1.0960072317113647e-05 
	R: 0.0010199570655822752 +- 1.5879270658969786e-05
 -> SrcImplicit (2000-atom subset) 
	W: 0.001061562697092692 +- 9.563426464632019e-06 
	R: 0.001057159105936686 +- 6.712241785818989e-06
 -> DHFRExplicit (10000-atom subset) 
	W: 0.004576456546783447 +- 5.45391163817968e-05 
	R: 0.001435536543528239 +- 1.0289301823887973e-05
```

Screenshot from 2019-07-31 10-51-01

Screenshot from 2019-07-31 10-47-46


## This PR

```
Timings per iteration and state, averaged over three attempts
 -> AlanineDipeptideVacuum (10-atom subset) 
	W: 0.0005906271934509277 +- 6.760185491223273e-06 
	R: 0.0001976950963338216 +- 3.987598232796745e-06
 -> SrcImplicit (2000-atom subset) 
	W: 0.0011950039863586426 +- 4.749539061513433e-07 
	R: 0.0003932054837544759 +- 1.2144305913238907e-06
 -> DHFRExplicit (10000-atom subset) 
	W: 0.004053071339925131 +- 2.855186750600774e-05 
	R: 0.0010753647486368815 +- 3.0104090569905125e-06
```
Screenshot from 2019-07-31 10-49-41

Screenshot from 2019-07-31 10-44-09

***

## But...

As you can see, performance has improved in both cases, and now reading is even faster for `xtc` than `nc`. Why? First benchmark was done on a last-gen Lenovo laptop with a high-end SSD, but **on Ubuntu running on the Windows 10 Linux Subsystem**, which is known to have a huge IO penalty. I assumed this penalty would be the same for both file formats (`xtc` and `netcdf`) but turns out `netcdf` must be doing something smart to avoid that problem.

This last benchmark has been performed on a 2017 Asus laptop with a standard SSD, but using a native Ubuntu installation. The difference is gone! Writing is now on the same order of magnitude (and there's room for improvement), and reading is even faster.

I guess we should test more platforms to get an idea on how performance can vary, but I'd say WSL will be always worse.

***

[profiles.zip](https://github.com/choderalab/openmmtools/files/3450672/profiles.zip)

jchodera:
Thanks for all the great work here!
I'll respond more fully shortly, but it may be worth posting some of the profiling info in the OpenMM issue tracker. One thing we had always discussed is allowing retrieval of unitless quantities with a special variant function call from the Python API, like _function. This might be good justification for the need.

jaimergp:
> I'll respond more fully shortly, but it may be worth posting some of the profiling info in the OpenMM issue tracker. One thing we had always discussed is allowing retrieval of unitless quantities with a special variant function call from the Python API, like _function. This might be good justification for the need.

Thanks, John! Are you referring to [this issue](https://github.com/pandegroup/openmm/issues/1490)? Should I just add a comment on that one?

jaimergp:
Timings for `lilac` (`cpuqueue`):

# Master
```
Timings per iteration and state, averaged over three attempts
 -> AlanineDipeptideVacuum (10-atom subset)
        W: 0.00034163713455200197 +- 8.698151755411477e-06
        R: 0.0008952434857686361 +- 1.7459108083169662e-05
 -> SrcImplicit (2000-atom subset)
        W: 0.0008839257558186849 +- 1.9250867516056536e-05
        R: 0.0009127648671468099 +- 1.0260630332895122e-05
 -> DHFRExplicit (10000-atom subset)
        W: 0.0038567423820495606 +- 0.0002045357621465894
        R: 0.0011864296595255534 +- 4.044621937473161e-05
```

image

image


# This PR
```
Timings per iteration and state, averaged over three attempts
 -> AlanineDipeptideVacuum (10-atom subset)
        W: 0.00102842648824056 +- 0.00046570606555826474
        R: 0.00023671547571818034 +- 5.1293549804258624e-05
 -> SrcImplicit (2000-atom subset)
        W: 0.0017322484652201335 +- 0.0009600916513871555
        R: 0.00037200450897216796 +- 3.7705866249522058e-06
 -> DHFRExplicit (10000-atom subset)
        W: 0.0044581524531046544 +- 7.981870646437672e-05
        R: 0.0010082292556762695 +- 3.4921978568712295e-05
```
image

image

# Conclusions

`master` writes faster in small systems, but is within the same order of magnitude with realistic system sizes (per-frame estimations). Absolute times seem for writing to be ~1.75x worse for this PR. We might need to test against a _real_ calculation instead of this synthetic test.

andrrizzi:
Great! From here we can only get better so I'd propose we merge this into the `parallel-writing` branch and keep going from here.

jaimergp:
Fine!

Issue #436, opened 2019-09-24
Barostat support in ThermodynamicState


Olllom:
Hi,

I would like to use the replica exchange facility for systems that have a `MonteCarloAnisotropicBarostat` or a `MonteCarloMembraneBarostat`.

Right now `ThermodynamicState` only supports the `MonteCarloBarostat` although it looks to me like upgrading it to the other barostat classes should be fairly straightforward.

From what I see in the code the anisotropic barostat should be **supported without any changes** to the code (other than adding it to the list of supported barostats). The `MonteCarloBarostat` can remain the go-to class if the system does not have a barostat upon initialization, but if an anisotropic barostat exists, it could get handled exactly as the isotropic barostat. Or am I missing something? 

Same for a membrane barostat with zero surface tension. 

It looks like the only non-trivial change would be to support surface tensions (getters, setters, keyword arguments in the constructor, and the gamma*A term in the reduced energy).

For now we could just check that the surface tension is zero when initializing the ThermodynamicState and add support for non-zero surface tensions later. 

What do you think?

Olllom:
* anisotropic barostat with equal pressure in all dimensions that are scaled

andrrizzi:
This is a very welcomed feature, thank you! I'll review the PR shortly.

PR #437, opened 2019-09-24
ThermodynamicState support for more barostats

Olllom:
## Description
Thermodynamic state supports
- MonteCarloAnistropicBarostat with same pressure along all axes
- MonteCarloMembraneBarostat with zero surface tension

see #436

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/states.py
+++ b/openmmtools/states.py
@@ -281,11 +281,13 @@ class ThermodynamicsError(Exception):
      MULTIPLE_BAROSTATS,
      NO_BAROSTAT,
      UNSUPPORTED_BAROSTAT,
+     UNSUPPORTED_ANISOTROPIC_BAROSTAT,
+     UNSUPPORTED_MEMBRANE_BAROSTAT,
      INCONSISTENT_BAROSTAT,
      BAROSTATED_NONPERIODIC,
      INCONSISTENT_INTEGRATOR,
      INCOMPATIBLE_SAMPLER_STATE,
-     INCOMPATIBLE_ENSEMBLE) = range(12)
+     INCOMPATIBLE_ENSEMBLE) = range(14)
 
     error_messages = {
         MULTIPLE_THERMOSTATS: "System has multiple thermostats.",
@@ -294,6 +296,10 @@ class ThermodynamicsError(Exception):
         INCONSISTENT_THERMOSTAT: "System thermostat is inconsistent with thermodynamic state.",
         MULTIPLE_BAROSTATS: "System has multiple barostats.",
         UNSUPPORTED_BAROSTAT: "Found unsupported barostat {} in system.",
+        UNSUPPORTED_ANISOTROPIC_BAROSTAT:
+            "MonteCarloAnisotropicBarostat is only supported if the pressure along all scaled axes is the same.",
+        UNSUPPORTED_MEMBRANE_BAROSTAT:
+            "MonteCarloAnisotropicBarostat is only supported if the surface tension is zero.",
         NO_BAROSTAT: "System does not have a barostat specifying the pressure.",
         INCONSISTENT_BAROSTAT: "System barostat is inconsistent with thermodynamic state.",
         BAROSTATED_NONPERIODIC: "Non-periodic systems cannot have a barostat.",
@@ -696,7 +702,7 @@ class ThermodynamicState(object):
         # Update the internally stored standard system, and restore the old
         # pressure if something goes wrong (e.g. the system is not periodic).
         try:
-            self._pressure = new_barostat.getDefaultPressure()
+            self._pressure = self._get_barostat_pressure(new_barostat)
             self._unsafe_set_system(system, fix_state=False)
         except ThermodynamicsError:
             self._pressure = old_pressure
@@ -1235,9 +1241,12 @@ class ThermodynamicState(object):
         # If pressure is None, we try to infer the pressure from the barostat.
         barostat = self._find_barostat(system)
         if pressure is None and barostat is not None:
-            self._pressure = barostat.getDefaultPressure()
+            self._pressure = self._get_barostat_pressure(barostat)
         else:
             self._pressure = pressure  # Pressure here can also be None.
+        self._barostat_type = openmm.MonteCarloBarostat
+        if barostat is not None:
+            self._barostat_type = type(barostat)
 
         # If temperature is None, we infer the temperature from a thermostat.
         if temperature is None:
@@ -1386,7 +1395,7 @@ class ThermodynamicState(object):
         # Here we push the barostat at the end.
         barostat = self._pop_barostat(system)
         if barostat is not None:
-            barostat.setDefaultPressure(self._STANDARD_PRESSURE)
+            self._set_barostat_pressure(barostat, self._STANDARD_PRESSURE)
             system.addForce(barostat)
 
     def _compute_standard_system_hash(self, standard_system):
@@ -1422,7 +1431,7 @@ class ThermodynamicState(object):
         # Apply pressure and temperature to barostat.
         if update_pressure:
             self._set_barostat_pressure(barostat, self.pressure)
-            context.setParameter(barostat.Pressure(), self.pressure)
+            self._set_barostat_pressure_in_context(barostat, self.pressure, context)
         if update_temperature:
             self._set_barostat_temperature(barostat, self.temperature)
             # TODO remove try except when drop openmm7.0 support
@@ -1544,7 +1553,7 @@ class ThermodynamicState(object):
     # Internal-usage: barostat handling
     # -------------------------------------------------------------------------
 
-    _SUPPORTED_BAROSTATS = {'MonteCarloBarostat'}
+    _SUPPORTED_BAROSTATS = {'MonteCarloBarostat', 'MonteCarloAnisotropicBarostat', 'MonteCarloMembraneBarostat'}
 
     @classmethod
     def _find_barostat(cls, system, get_index=False):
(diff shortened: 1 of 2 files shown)


openmmtools/states.py
```diff
@@ -1235,9 +1241,12 @@ def _initialize(self, system, temperature=None, pressure=None):
         # If pressure is None, we try to infer the pressure from the barostat.
         barostat = self._find_barostat(system)
         if pressure is None and barostat is not None:
-            self._pressure = barostat.getDefaultPressure()
+            self._pressure = self._get_barostat_pressure(barostat)
         else:
             self._pressure = pressure  # Pressure here can also be None.
+        self._barostat_type = openmm.MonteCarloBarostat
```
andrrizzi: What do you think about making `_barostat_type` a private property that simply returns `type(self.barostat)`? Because this attribute is accessed very rarely, I'd prefer to make it slightly slower but avoid redundancies.

openmmtools/states.py
```diff
@@ -1573,6 +1582,19 @@ def _find_barostat(cls, system, get_index=False):
             if barostat.__class__.__name__ not in cls._SUPPORTED_BAROSTATS:
                 raise ThermodynamicsError(ThermodynamicsError.UNSUPPORTED_BAROSTAT,
                                           barostat.__class__.__name__)
+            elif isinstance(barostat, openmm.MonteCarloAnisotropicBarostat):
+                # support only if pressure in all scaled directions is equal
+                pressures = barostat.getDefaultPressure().value_in_unit(unit.bar)
+                scaled = [barostat.getScaleX(), barostat.getScaleY(), barostat.getScaleY()]
+                if sum(scaled) == 0:
+                    raise ThermodynamicsError(ThermodynamicsError.UNSUPPORTED_ANISOTROPIC_BAROSTAT)
+                active_pressures = [pressure for pressure, active in zip(pressures, scaled)]
```
andrrizzi: `active` here is unused

andrrizzi:
As an extension (but feel free to postpone this to a future PR if you don't need it) we could allow `self.pressure` to be a vector to support different values of pressures for the axes of the anisotropic MC barostat.

openmmtools/states.py
Olllom: I would rather like to keep it as it is. The reason is that `_initialize` calls `_is_barostat_consistent` indirectly (through `_unsafe_set_system` and `_check_system_consistency`_). At this point, `self.barostat` is not available, since `self._standard_system` is not initialized, yet.

openmmtools/states.py
Olllom: Good catch! Thank you.

Olllom:
An easier and more relevant addition for us would be nonzero surface tension. I will add some code for this to the pull request.

openmmtools/states.py
andrrizzi: I see what you mean. It may be safe after all though because, if I understand correctly the code I wrote, `_is_barostat_consistent` shouldn't be called when `self.barostat` is unavailable (I'm looking at these lines: https://github.com/Olllom/openmmtools/blob/other_barostats/openmmtools/states.py#L1341-L1342).

andrrizzi:
Good point about the pV work. Feel very free to implement only the bits you need. This is already very helpful.

Olllom:
I added
1. support for non-zero surface tension
2. support for setting the platform properties in the context cache and in the thermodynamic state

Point 2. should enable multistate simulations on CUDA in mixed and double precision,
by binding the MCMC moves to caches created with the 'precision'

It also makes it unnecessary to reduce CUDA_VISIBLE_DEVICES to a single device in multinode-gpu simulations. Instead the device ID can be set as a platform property.

Olllom:
This PR is ready to go on my end.

openmmtools/states.py
```diff
@@ -2032,6 +2105,11 @@ def volume(self):
         """The volume of the box (read-only)"""
         return _box_vectors_volume(self.box_vectors)
 
+    @property
+    def area(self):
```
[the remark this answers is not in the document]
Olllom: Done

openmmtools/cache.py
```diff
@@ -447,13 +453,17 @@ def __getstate__(self):
...
         return dict(platform=platform_serialization, capacity=self.capacity,
-                    time_to_live=self.time_to_live)
+                    time_to_live=self.time_to_live, platform_properties=self._platform_properties)
 
     def __setstate__(self, serialization):
         if serialization['platform'] is None:
             self._platform = None
         else:
             self._platform = openmm.Platform.getPlatformByName(serialization['platform'])
+        if not 'platform_properties' in serialization:
```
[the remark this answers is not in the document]
Olllom: Done

openmmtools/cache.py
```diff
@@ -303,8 +306,11 @@ class ContextCache(object):
 
     """
 
-    def __init__(self, platform=None, **kwargs):
+    def __init__(self, platform=None, platform_properties=None, **kwargs):
         self._platform = platform
+        self._platform_properties = platform_properties
+        if platform_properties is not None and platform is None:
```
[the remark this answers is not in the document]
Olllom: Done

docs/releasehistory.rst
```diff
@@ -1,6 +1,11 @@
 Release History
 ***************
 
+Enhancements
+------------
+- Added support for anisotropic and membrane barostats in `ThermodynamicState`
+- Added support for platform properties in ContextCache (e.g. for mixed and double precision CUDA in multistate sampler)
```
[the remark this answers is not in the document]
Olllom: Done

Olllom:
Hi Andrea,

thanks for your suggestions and sorry for the delay. I have incorporated all of your requests. 

In order to allow compatibility between states with different surface tensions, I basically mimicked the pressure implementation. This should be the most consistent way to standardize the system, since the surface tension needs to be defined by the ThermodynamicState rather than the system barostat.

Instead of having a field for the barostat type, I added a private method `_is_barostat_type_consistent`. This should avoid the inconsistencies you mentioned above.

jchodera:
Since this PR has added significant complexity, is there anything we could implement at the OpenMM API  for these barostats that would significantly simplify this code? We're looking for features to add for the next OpenMM release, so if there is an easy way to extend the OpenMM API to facilitate this, please let us know!

andrrizzi:
Thanks @Olllom ! I'll take a look at the changes before the end of the week. Now that we merged the multiple alchemical region features, I should be able to take a look at your other PR modifying the alchemical factory too next week.

openmmtools/states.py
```diff
@@ -1303,9 +1315,15 @@ def _initialize(self, system, temperature=None, pressure=None):
             self._pressure = self._get_barostat_pressure(barostat)
         else:
             self._pressure = pressure  # Pressure here can also be None.
-        self._barostat_type = openmm.MonteCarloBarostat
-        if barostat is not None:
-            self._barostat_type = type(barostat)
+
+        # If surface tension is None, we try to infer the pressure from the barostat.
```
andrrizzi: "infer the pressure" -> "infer the surface tension"

openmmtools/states.py
```diff
@@ -1757,6 +1804,32 @@ def _set_barostat_temperature(barostat, temperature):
...
+        if gamma is not None and not system.usesPeriodicBoundaryConditions():
+            raise ThermodynamicsError(ThermodynamicsError.BAROSTATED_NONPERIODIC)
+        barostat = self._find_barostat(system)
+        if (gamma is None) == isinstance(barostat, openmm.MonteCarloMembraneBarostat):
+            raise ThermodynamicsError(ThermodynamicsError.INCOMPATIBLE_ENSEMBLE)
+        self._set_barostat_surface_tension(barostat, gamma)
+
+
+    def _set_barostat_surface_tension(self, barostat, gamma):
+        # working around a bug in the unit conversion https://github.com/openmm/openmm/issues/2406
```
andrrizzi: Nice catch!

openmmtools/states.py
```diff
@@ -1757,6 +1804,32 @@ def _set_barostat_temperature(barostat, temperature):
...
+        if isinstance(gamma, unit.Quantity):
+            gamma = gamma.value_in_unit(unit.bar * unit.nanometer)
+        # barostat = self._find_barostat(system)
```
andrrizzi: Probably forgotten here.

openmmtools/states.py
```diff
@@ -1618,6 +1648,11 @@ def set_temp(_integrator):
 
     _SUPPORTED_BAROSTATS = {'MonteCarloBarostat', 'MonteCarloAnisotropicBarostat', 'MonteCarloMembraneBarostat'}
 
+    @property
+    def _barostat_type(self):
```
andrrizzi: Probably forgotten here since it's unused.

openmmtools/tests/test_states.py
```diff
@@ -437,7 +437,7 @@ def test_property_system(self):
                       (self.barostated_toluene, TE.BAROSTATED_NONPERIODIC),
                       (self.multiple_barostat_alanine, TE.MULTIPLE_BAROSTATS),
                       (self.unsupported_anisotropic_barostat_alanine, TE.UNSUPPORTED_ANISOTROPIC_BAROSTAT),
-                      (self.supported_anisotropic_barostat_alanine, TE.INCONSISTENT_BAROSTAT),
+                      #(self.supported_anisotropic_barostat_alanine, TE.INCONSISTENT_BAROSTAT),
```
andrrizzi: Did you mean to comment this test case?

Olllom:
Great!
All fixed.

andrrizzi:
Almost! It looks there's one of the new tests failing: https://travis-ci.org/choderalab/openmmtools/jobs/603266322#L3314

andrrizzi:
There are also these other two errors in the "allowed failures" that I haven't seen before, but I'm not sure this is related to your changes so I'd address in a separate PR unless you know why that is happening: https://travis-ci.org/choderalab/openmmtools/jobs/603266325#L3185

andrrizzi:
I confirm that the latter tests are unrelated as they appear also in a PR without these changes so there's just the surface tension test to fix.

jchodera:
Is there anything we could implement at the OpenMM API for these barostats that would significantly simplify this code?

andrrizzi:
>Is there anything we could implement at the OpenMM API for these barostats that would significantly simplify this code?

Nothing particularly useful comes to mind. I think most of the complexity here comes from the fact that we're providing a unified API to a bunch of objects that have different parameters and (consequently) different APIs. We could create a general `Barostat` Python object in OpenMM that does these checks for us, but I think the total complexity (i.e., OpenMM + OpenMMTools) will be identical in the end.

Olllom:
@jchodera Something like this old PR of yours would be useful to simplify the API: https://github.com/openmm/openmm/issues/1440 

In this way, OpenMM could provide a simplified API to set things like temperature, pressure, surface tension, but also step size, etc. But this would only simplify the present code to a certain degree. IMHO a lot of the complexity in the `ThermodynamicState` is due to the `_standardized_system`, from which the data stored explicitly in the `ThermodynamicState` (pressure, temperature, surface tension) is excluded. However, I do not see a good way around this without hurting performance.

On a side note, one rather boring feature that would be very useful for us is native OpenMM support for the van-der-Waals force switch (see https://github.com/choderalab/openmmtools/pull/431). We have to use custom forces for the vdW in most of our simulations to get the correct lipid areas, which is quite a significant performance hit.

Olllom:
@andrrizzi should be fixed now.

andrrizzi:
Fantastic. Thanks again for the really nice work on tests too. Merging.

PR #438, opened 2019-09-25
Multiple alchemical regions

adw62:
Provides feature for multiple alchemical regions, now with tests.

Todo:
Tests do not cover the interaction between alchemical regions.
Using reaction field causes errors in testing.


openmmtools/alchemy.py
```diff
@@ -1984,91 +2323,155 @@ def check_energy_expression(custom_force, parameter):
...
             return found
-
+        
         force_labels = {}
-        nonbonded_forces = []
-        sterics_bond_forces = []
-        electro_bond_forces = []
+        nonbonded_forces = {'' :[], 'zero': [], 'one': [], 'two': []}
+        sterics_bond_forces = {'' :[], 'zero': [], 'one': [], 'two': []}
+        electro_bond_forces = {'' :[], 'zero': [], 'one': [], 'two': []}
```
andrrizzi: It may be possible to avoid hardcoding these by having `check_parameters` finding the region name with something like
```python
if parameter_name.startswith(parameter):
    region_name = parameter_name[len(parameter)+1:] if len(parameter_name) > len(parameter) else ''
    return True, region_name
```

openmmtools/alchemy.py
```diff
@@ -1403,309 +1584,453 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
+                    at_least_one_alchemical = iatom in alchemical_atomset_0 or jatom in alchemical_atomset_1
+
+                    # Check if this is an exception or an exclusion
+                    is_exception_epsilon = abs(epsilon.value_in_unit_system(unit.md_unit_system)) > 0.0
+                    is_exception_chargeprod = abs(chargeprod.value_in_unit_system(unit.md_unit_system)) > 0.0
+
+                    if use_exact_pme_treatment and both_alchemical and is_exception_chargeprod:
+                        # Not sure how to deal with this case. Exception should be scaled by lam0*lam1.
+                        # However we can only have one Offset parameter per exception?
+                        raise ValueError('Cannot have exception that straddles two alchemical regions')
```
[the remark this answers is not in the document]
adw62: Okay I have just amended the comment here to reflect this for the future.

openmmtools/alchemy.py
```diff
@@ -1403,309 +1584,453 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
+                        if is_exception_epsilon:
+                            aa_sterics_custom_bond_force.addBond(iatom, jatom, [sigma, epsilon])
+                        if is_exception_chargeprod and not use_exact_pme_treatment:
+                            aa_electrostatics_custom_bond_force.addBond(iatom, jatom, [chargeprod, sigma])
+
+                    # Turn off all exception contributions from alchemical atoms in the NonbondedForce
+                    # modelling non-alchemical atoms only
+                    if at_least_one_alchemical:
+                        nonbonded_force.setExceptionParameters(exception_index, iatom, jatom,
+                                                               abs(0.0*chargeprod), sigma, abs(0.0*epsilon))
```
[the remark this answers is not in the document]
adw62: Yes, I think you are correct. I have taken these lines out.

openmmtools/alchemy.py
```diff
@@ -1403,309 +1584,453 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
+                    [iatom, jatom, chargeprod, sigma, epsilon] = reference_force.getExceptionParameters(exception_index)
+
+                    # Exclude this atom pair in CustomNonbondedForces. All nonbonded forces
+                    # must have the same number of exceptions/exclusions on CUDA platform.
+                    for force in all_custom_nonbonded_forces:
+                        force.addExclusion(iatom, jatom)
+
+                    # Check how many alchemical atoms we have
+                    both_alchemical = iatom in alchemical_atomset_0 and jatom in alchemical_atomset_1
+                    at_least_one_alchemical = iatom in alchemical_atomset_0 or jatom in alchemical_atomset_1
```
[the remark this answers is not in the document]
adw62: Possibly I can't convince myself this couldn't happen so I've added an or condition to cover this case.

openmmtools/alchemy.py
```diff
@@ -1403,309 +1584,453 @@ def _alchemically_modify_NonbondedForce(self, reference_force, alchemical_region
...
+
+            # ---------------------------------------------------------------
+            # Distribute exceptions contributions in appropriate bond forces
+            # ---------------------------------------------------------------
+
+            all_custom_nonbonded_forces = all_sterics_custom_nonbonded_forces + all_electrostatics_custom_nonbonded_forces
+
+            # Move all NonbondedForce exception terms for alchemically-modified particles to CustomBondForces.
+            if len(lambda_var_suffixes) > 1:
+                for exception_index in range(nonbonded_force.getNumExceptions()):
```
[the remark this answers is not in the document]
adw62: Made some simplifications here. Let me know if this is along the lines you were thinking.

openmmtools/alchemy.py
adw62: Do you thin this can remove all the hard coding in '_find_force_components' or just the lines you highlighted?

openmmtools/alchemy.py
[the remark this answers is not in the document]
andrrizzi: Sounds good!

openmmtools/alchemy.py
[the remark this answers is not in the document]
andrrizzi: That made things much better, thank you! I think I managed to remove some more redundant code. Could you check my last commit to see if I made a mistake?

openmmtools/alchemy.py
andrrizzi: I think it could remove the hard coding in `_find_force_components` altogether. You'd have to substitute this pattern
```python
if check_energy_expression(force, 'lambda_sterics'):
    if check_energy_expression(force, 'lambda_sterics_zero'):
        sterics_bond_forces['zero'].append([force_index, force])
    ...
```
with something like
```python
if check_energy_expression(force, 'lambda_sterics'):
    # region_name is None if the parameter is not found in the force and '' if it has no suffix.
    region_name = check_parameter(force, 'lambda_sterics_zero')
    sterics_bond_forces[region_name].append([force_index, force])
```
unless there's somehting I don't see.

andrrizzi: Ah! Sorry, I remember now that I used `check_energy_expression` for the alchemical-alchemical custom forces that are decoupled instead of annihilated (and have no `lambda_sterics` parameter). To make it work we'll have to modify `check_energy_expression` too, not just `check_parameter`. I'm thinking using regex.

andrrizzi: Something like this works for me (see https://stackoverflow.com/questions/15340582/python-extract-pattern-matches)
```python
>>> a = 'asdfa sdfalambda_sterics_zero*asdf'
>>> import re
>>> p = re.compile(r'lambda_sterics_([A-Za-z0-9]+)')
>>> p.search(a).group(0)
'lambda_sterics_zero'
>>> p.search(a).group(1)
'zero'
```
I think you can do a simple `parameter in custom_force.getEnergyFunction()` as it is done now, and then try that regex to see if it has a region associated.

andrrizzi:
I see what you mean. Yeah, I think I'd prefer to simplify the code in this case as the cost of the if statement will be negligible w.r.t. the rest of the operations inside the foor loop.

> Although I have changed it a little to remove one of the if statements

Much better now, thanks!

openmmtools/alchemy.py
adw62: I've reworked `_find_force_components` to remove the hard coded region names. It's not final and is causing HostGuestExplicit with exact PME to fail currently. But is this rework similar to what you were thinking?

andrrizzi:
Just looked at this briefly, but it looks like what I had in mind 👍

openmmtools/alchemy.py
adw62: Finished off `_find_force_components` think it's working as intended. Let me know if there are any more changes that need to be made.

andrrizzi:
Merging! 🎉

adw62:
Great. Thank you for all your help with this!

andrrizzi:
No problem!

PR #439, opened 2019-09-26
[WIP] Solvent-solute splitting

maxentile:
## Description
Intention: Efficient implementations of force factories for splitting solvent-solvent interactions from other interactions, for use in multiple-timestepping schemes.

## Status
Incomplete. Currently the splitter using exceptions and the splitter using interaction groups are both giving pretty big force errors, and the one using subtraction is giving forces that agree within a very tight tolerance with the original system. However, *dynamics* with the one using subtraction is unstable at 2fs... 

CC'ing @c-matthews and @jchodera

## Todos
  - [x] Draft implementations
  - [ ] Fix possible double-counting in interaction-group-based implementation
- [ ] Add tests
  - [x] Add test asserting that forces are similar between reference system and split system (currently failing for 2 of 3 draft implementations)
  - [x] Make sure test is using `CutoffPeriodic` nonbonded method
  - [ ] Test nonbonded-parameter cloning method separately


openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+    new_force.addPerParticleParameter('charge')
+    new_force.addPerParticleParameter('sigma')
+    new_force.addPerParticleParameter('epsilon')
+
+
+    # go through all of the setter and getter methods
+    new_force.setCutoffDistance(nonbonded_force.getCutoffDistance())
+    #new_force.setEwaldErrorTolerance(nonbonded_force.getEwaldErrorTolerance())
+    #new_force.setForceGroup(nonbonded_force.getForceGroup())
+    new_force.setNonbondedMethod(nonbonded_force.getNonbondedMethod())
```
jchodera: The nobonded methods may not directly translate from NonbondedForce to CustomNonbondedForce.
See: https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L1488-L1491

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+    new_force.setSwitchingDistance(nonbonded_force.getSwitchingDistance())
+    new_force.setUseLongRangeCorrection(nonbonded_force.getUseDispersionCorrection())
+
+
+    # now add all the particle parameters
+    num_particles = nonbonded_force.getNumParticles()
+    for i in range(num_particles):
+        new_force.addParticle(nonbonded_force.getParticleParameters(i))
+
+    # now add all the exceptions? # TODO: check if we want to do this...
```
jchodera: I think you definitely want to copy all exceptions -> exclusions, or else you'll end up in NaNdyland.

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+    group 1.
+    """
+
+    # create a copy of the original system: only touch this
+    new_system = copy.deepcopy(system)
+
+    # find the default nonbonded force
+    force_index, nb_force = forces.find_forces(new_system, openmm.NonbondedForce, only_one=True)
+    # create copies for each interaction. Only half in solvent/solvent and solute/solute as we double-count.
+    nb_only_solvent_solvent = clone_nonbonded_parameters(nb_force, energy_prefactor='0.5*')
```
jchodera: Why do you want to multiple the energy by 0.5?

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+
+    # NOTE: these need to be python ints -- not np.int64s -- when passing to addInteractionGroup later!
+    solvent_indices = list(map(int, md_topology.select('water')))
+    solute_indices = list(map(int, md_topology.select('not water')))
+
+    #all_indices = list(range(md_topology.n_atoms))
+
+    nb_only_solvent_solvent.addInteractionGroup(set1=solvent_indices, set2=solvent_indices)
+
+    nb_solute.addInteractionGroup(set1=solute_indices, set2=solute_indices)
```
jchodera: You can just use
```python
    all_indices = solute_indices.union(solvent_indices)
    nb_solute.addInteractionGroup(set1=solute_indices, set2=all_indices)
```

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+    nb_solute.addInteractionGroup(set1=solute_indices, set2=solvent_indices)
+
+    #nb_solute.addInteractionGroup(set1=solute_indices, set2=all_indices)
+
+    # remove original force, add new forces
+    new_system.removeForce(force_index)
```
jchodera: I think you want to leave in the original force so you don't have to copy all the exceptions (nonzero 1-4 interactions) to a CustomBondForce.

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+    new_system = copy.deepcopy(system)
+
+    # find the default nonbonded force
+    force_index, nb_force = forces.find_forces(new_system, openmm.NonbondedForce, only_one=True)
+    # assert('Cutoff' in nb_force.getNonbondedMethod()) # TODO: this, less jankily
+
+    # create copies for each interaction. Only half in solute/solute as we double count.
+    nb_only_solvent_solvent = nb_force
+    nb_only_solvent_solute = clone_nonbonded_parameters(nb_force)
+    nb_only_solute_solute = clone_nonbonded_parameters(nb_force,energy_prefactor='0.5*')
```
jchodera: Why would you multiply the energy by 0.5?

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+
+    # NOTE: these need to be python ints -- not np.int64s -- when passing to addInteractionGroup later!
+    solvent_indices = list(map(int, md_topology.select('water')))
+    solute_indices = list(map(int, md_topology.select('not water')))
+
+    # for each (i,j) set chargeprod, sigma, epsilon to 0, causing these pairs to be omitted completely from force calculation
+    # Remove solute-solute interactions
+    for i in solute_indices:
+        for j in solute_indices:
+            nb_only_solvent_solvent.addException(i, j, 0, 0, 0, replace=True)
```
jchodera: This will probably be slow.

openmmtools/forcefactories.py
```diff
@@ -174,6 +174,240 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
...
+        if 'Nonbonded' not in force.__class__.__name__:
+            force.setForceGroup(1)
+
+    return new_system
+
+
+
+# surrogate could deviate in functional form from the NonbondedForce defaults (e.g. using soft-core)
+surrogate_energy_expression = default_energy_expression
+half_surrogate_energy_expression = '0.5*' + surrogate_energy_expression
```
jchodera: Why use a factor of 0.5?

maxentile:
Thanks!

openmmtools/forcefactories.py
maxentile: Thanks! Will follow this example more closely

openmmtools/forcefactories.py
maxentile: Will do!

openmmtools/forcefactories.py
maxentile: Mistakenly assumed double-counting -- will revert

openmmtools/forcefactories.py
maxentile: Will uncomment this equivalent way (L249 and L256)

openmmtools/forcefactories.py
maxentile: Okay!

openmmtools/forcefactories.py
maxentile: Mistakenly assumed double-counting -- will revert

openmmtools/forcefactories.py
maxentile: Yes!

openmmtools/forcefactories.py
maxentile: Mistakenly assumed double-counting -- will revert

jchodera:
It's probably a good idea to make sure the total potential energy matches the unmodified system pretty closely!

maxentile:
>It's probably a good idea to make sure the total potential energy matches the unmodified system pretty closely!

This test is currently failing, but we're aiming to match within the same tolerance as the alchemical factories https://github.com/choderalab/openmmtools/blob/be65ec6b48ff54328af247ba5d3c85e4f5c6a46f/openmmtools/tests/test_forcefactories.py#L201-L210

jchodera:
It's probably much simpler just to compare the potential energy.

openmmtools/forcefactories.py
jaimergp: May I suggest `itertools.product(solute_indices, repeats=2)` instead of the nested `for`? Nothing to do with performance but I'd say it's prettier :sparkles:

PR #442, opened 2019-10-19
Fix travis test failures and test nightly and conda-forge openmm builds on travis

jchodera:
This PR adds travis test branches for nightly and conda-forge testing openmm builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,10 +57,10 @@ install:
     # Build and install package
   - python setup.py develop --no-deps
     # Install desired OpenMM version
-  - if [ "$OPENMM" == "latest" ]; then echo "Using latest release OpenMM."; conda remove --yes --force openmm; conda clean -tipsy; conda install --yes -c omnia openmm; fi
-  - if [ "$OPENMM" == "beta" ]; then echo "Using OpenMM beta"; conda remove --yes --force openmm; conda clean -tipsy; conda install --yes --override-channels -c omnia/label/beta openmm; fi
-  - if [ "$OPENMM" == "nightly" ]; then echo "Using OpenMM nightly dev build."; conda remove --yes --force openmm; conda clean -tipsy; conda install --yes --override-channels -c omnia-dev openmm; fi
-  - if [ "$OPENMM" == "conda-forge" ]; then echo "Using OpenMM conda-forge testing build."; conda remove --yes --force openmm; conda clean -tipsy; conda install --yes --override-channels -c conda-forge/label/testing openmm; fi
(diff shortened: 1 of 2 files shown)


jaimergp:
Looks like `openmm` is trying to use OpenCL [here](https://travis-ci.org/choderalab/openmmtools/jobs/599950704#L3169), but cannot find any OpenCL device (no implementation is present, which makes sense). This might have to do with `ocl-icd` being in the environment, but that shouldn't trigger OpenCL being chosen as the platform. Indeed, this is not happening in `perses`...

I can think of two fixes while we investigate whether this is a problem in `openmm` or not.

A) Install `pocl` to provide an OpenCL CPU implementation
B) Force the tests to use the `CPU` platform.

jchodera:
@andrrizzi : Have you seen this kind of OpenCL behavior before on travis?

@jaimergp : I don't think we should be using `pocl` to provide an OpenCL CPU implementation---this is exactly what @peastman was worried about. If the OpenCL platform is being accidentally used, then this is a huge problem, isn't it, since that's _not_ supposed to happen with the `ocl-icd` build of OpenMM from `conda-forge`.

jaimergp:
What happens if you do `conda remove --force ocl-icd ocl-icd-system`?

jchodera:
Good idea. But hm, for some reason travis is not running.

jaimergp:
It's still failing but for a different reason now, so I'd say that the OpenCL problem is gone?

jchodera:
OK, so this means there _are_ significant issues with including `ocl-icd` and `ocl-icd-system` in the build, right?

jaimergp:
Looks like `openmm` thinks it's a real implementation and not simply a loader. I'd feel more confident on that diagnosis if we had a specific test for this, so I'll see if I can come up with something.

jchodera:
Are you able to try running the benchmarks locally? OpenMM installs in `$MINICONDA/share/openmm/examples/` a `benchmark.py` script:
```bash
python benchmark.py --test=rf --seconds=30 --platform=OpenCL
```

jaimergp:
Yep, that's enough to reproduce:

```
conda create -n openmm-cf-ocl-icd -c conda-forge/label/testing -c conda-forge openmm
conda activate openmm-cf-ocl-icd
cd $CONDA_PREFIX/share/openmm/examples

# This will fail
python benchmark.py --test=rf --seconds=30 --platform=OpenCL

# If we remove ocl-icd, it works
conda remove --force ocl-icd ocl-icd-system
python benchmark.py --test=rf --seconds=30 --platform=OpenCL
```

jchodera:
Woohoo!

PR #444, opened 2019-10-19
Fix reduced_potential_at_states

andrrizzi:
Fix the shortcut function `reduced_potential_at_states` that didn't actually return anything -.-


jchodera:
Good catch! Should we add a test, since it looks like we were missing test coverage for this previously?

andrrizzi:
Good idea! I'll add the test before merging.

PR #446, opened 2019-10-26
Rework harmonic oscillator free energy tests for multistate samplers to run on travis

### LINKED ISSUE #441: Extend unit tests for multistate sampler harmonic oscillator tests to ensure that unsampled endstates are tested
We think there is an issue where unsampled endstates cause issues with SAMS.
cc: https://github.com/choderalab/perses/issues/542

We should extend [this test](https://github.com/choderalab/openmmtools/blob/master/openmmtools/tests/test_sampling.py#L245) to include some unsampled endstates to make sure they are not NaN.

jchodera:
## Description
This PR reworks the harmonic oscillator free energy tests for multistate samplers, testing them both with and without unsampled states and ensuring they are fast enough to run on travis (fixes #441).

I've also converted the `warnings.warn` messages about the experimental API to `logger.warn` so they can be more easily suppressed with the logger level.


docs/releasehistory.rst
```diff
@@ -8,6 +8,10 @@ New features
 ------------
 - Added support in ``AbsoluteAlchemicalFactory`` for handling multiple independent alchemical regions (`#438 <https://github.com/choderalab/openmmtools/pull/438>`_).
 
+Bugfixes
+--------
+- The multistate samplers now issue experimental API warnings via `logger.warn()` rather than `warnings.warn()` (#446)
```
andrrizzi: This should be in sphinx syntax so code font should be surrounded by `` instead of a single one, and the link should be
```
(`#446 <https://github.com/choderalab/openmmtools/pull/446>`_)
```

jchodera:
> Just for curiosity, what's the reason why you had to re-generate the cython code? 

That was accidental. It appears that installing the code locally caused that to happen.
Should we exclude that `.c` file from being included in [`.gitignore`](https://github.com/choderalab/openmmtools/blob/master/.gitignore) so that it is built at install time?

> Do you think this might be related to the mixing problems with multiple GPUs we observed (choderalab/yank#1130)?

Hm. It could be that the view into the arrays somehow skipped every other element. But didn't we only see this issue if we were using MPI? How would MPI be interacting with this?

docs/releasehistory.rst
jchodera: Fixed!

jchodera:
The [cython changelog](https://github.com/cython/cython/blob/master/CHANGES.rst) doesn't list any bugfixes that seem suspiciously like ours since 0.29.4.

andrrizzi:
> Should we exclude that .c file from being included in .gitignore so that it is built at install time?

I think it's fine to track the .c file. It probably shortens slightly installation and, in principle, it allows us to not rely on the cython dependency.

> But didn't we only see this issue if we were using MPI? How would MPI be interacting with this?

Yes, you're probably right. I'm going to take another look at the problem in the next few days though. If this didn't change anything, maybe there was something that changed in mpich that fixed it.

PR #463, opened 2020-04-05
Option to return fastest double-supporting platform

hannahbrucemacdonald:
## Description
double=True can be set. This is to help with issue
https://github.com/choderalab/perses/issues/657

default behaviour of function is unchanged

## Todos

## Status
- [ ] Ready to go


hannahbrucemacdonald:
closing this - @jchodera pointed out this wouldn't necessarily work for the OpenCL platform, which will return `True` even if it may not

PR #464, opened 2020-04-10
Loopy `CustomIntegrator`

dominicrufa:
## Description
add a loopy custom integrator that does neq switching back and forth

## Todos
- [ ] test `LoopyIndependentAlchemicalNonequilibriumLangevinIntegrator`

## Status
- [ ] almost ready to go

PR #466, opened 2020-04-15
Add PeriodicNonequilibriumIntegrator

jchodera:
## Description
This PR adds a `PeriodicNonequilibriumIntegrator`, a simple extension of `AlchemicalNonequilibriumLangevinIntegrator` that supports periodic alchemical protocols.

@dominicrufa : This is heavily based on your `Loopy` and bidirectional integrators, but since this is simpler and more compatible with the current Folding@home infrastructure, I'm going to merge this first and then return to #464 in a day or two.

## Todos

## Status
- [x] Ready to go


jchodera:
Weird---there are issues that seem unrelated:
```
testing alchemically modified BondForce for alchemical/alchemical electrostatics exceptions
```
Any ideas here @dominicrufa ?

openmmtools/integrators.py
```diff
@@ -1895,6 +1905,145 @@ def _add_alchemical_reset_step(self):
...
+    >>> integrator = PeriodicAlchemicalNonequilibriumLangevinIntegrator(
+    ...                  temperature=300*unit.kelvin, collision_rate=1.0/unit.picoseconds, timestep=1.0*unit.femtoseconds,
+    ...                  alchemical_functions=alchemical_functions, splitting="V R H O R V",
+    ...                  nsteps_eq=nsteps_eq, nsteps_neq=nsteps_neq)
+    >>> # Create a Context
+    >>> context = openmm.Context(testsystem.system, integrator)
+    >>> # Run one periodic cycle: (eq 0) > (neq 0->1) > (eq 1) > (neq 1->0)
+    >>> context.setPositions(testsystem.positions)
+    >>> nsteps_per_period = 2*nsteps_eq + 2*nsteps_neq
+    >>> integrator.step(nsteps_per_period)
```
dominicrufa, line 1942: it seems like this class was written to allow for single `integrator.step()` calls which don't complete the nonequilibrium _cycle_; is this correct?

openmmtools/integrators.py
```diff
@@ -1895,6 +1905,145 @@ def _add_alchemical_reset_step(self):
...
+
+        """
+        # Store initial potential energy
+        self.addComputeGlobal("Eold", "energy")
+
+        # Update lambda and increment that tracks updates.
+        lambda_control_expression = "lambda"\
+            " + delta_lambda*step((step+0.5)-n_steps_eq)*step((n_steps_eq+n_steps_neq) - (step+0.5))"\
+            " - delta_lambda*step((step+0.5)-(n_steps_eq+n_steps_neq+n_steps_eq))*step(n_steps_per_cycle - (step+0.5));"\
+            "delta_lambda = 1/n_lambda_steps;" # Note that there may be multiple H per integrator string, so we correct for this here
```
dominicrufa, line 2036: could you split this expression into substrings and provide a short comment on each part?  it's not too easy to read in this format

dominicrufa:
> Weird---there are issues that seem unrelated:
> 
> ```
> testing alchemically modified BondForce for alchemical/alchemical electrostatics exceptions
> ```
> 
> Any ideas here @dominicrufa ?



> Weird---there are issues that seem unrelated:
> 
> ```
> testing alchemically modified BondForce for alchemical/alchemical electrostatics exceptions
> ```
> 
> Any ideas here @dominicrufa ?

hm. never seen this before.  can you post the whole traceback?

jchodera:
@dominicrufa : I was just picking the errors out of the [travis failures](https://travis-ci.org/github/choderalab/openmmtools/builds/675168958), but it looks like there are a variety of odd errors I haven't seen before.

openmmtools/integrators.py
jchodera: That's correct! It's the same way that `AlchemicalNonequilibriumLangevinIntegrator` works, except instead of halting after `n_steps_neq`, it continues periodic cycling infinitely. This way, we can also measure the work at various points through the switching, and this allows us to capture useful information about the efficiency of the protocol.

I think your integrators are super useful as well, so I'll come back to those in a couple of days.

openmmtools/integrators.py
jchodera: Great idea!

jchodera:
Thanks for the great feedback, @dominicrufa !

PR #467, opened 2020-04-21
Fix an issue with accidentally resetting PeriodicNonequilibriumIntegrator

jchodera:
We should be accumulating work as we go around the cycle, not resetting the integrator each cycle!

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -1644,7 +1644,7 @@ class NonequilibriumLangevinIntegrator(LangevinIntegrator):
         Manually reset protocol work and other statistics.
         """
         self.reset_protocol_work()
-        super(NonequilibriumLangevinIntegrator, self).reset()
+        super().reset()
 
 class AlchemicalNonequilibriumLangevinIntegrator(NonequilibriumLangevinIntegrator):
     """Allows nonequilibrium switching based on force parameters specified in alchemical_functions.
@@ -1876,14 +1876,14 @@ class AlchemicalNonequilibriumLangevinIntegrator(NonequilibriumLangevinIntegrato
         if self._n_steps_neq == 0:
             # If nsteps = 0, we need to force execution on the first step only.
             self.beginIfBlock('step = 0')
-            super(AlchemicalNonequilibriumLangevinIntegrator, self)._add_integrator_steps()
+            super()._add_integrator_steps()
             self.addComputeGlobal("step", "step + 1")
             self.endBlock()
         else:
             #call the superclass function to insert the appropriate steps, provided the step number is less than n_steps
             self.beginIfBlock("step >= 0")
             self.beginIfBlock("step < n_steps_per_cycle")
-            super(AlchemicalNonequilibriumLangevinIntegrator, self)._add_integrator_steps()
+            super()._add_integrator_steps()
             self.addComputeGlobal("step", "step + 1")
             self.endBlock()
             self.endBlock()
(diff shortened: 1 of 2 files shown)


openmmtools/tests/test_integrators.py
```diff
@@ -774,7 +774,7 @@ def run_alchemical_langevin_integrator(nsteps=0, splitting="O { V R H R V } O"):
     if nsigma > NSIGMA_MAX:
         raise Exception("The free energy difference for the nonequilibrium switching for splitting '%s' and %d steps is not zero within statistical error." % (splitting, nsteps))
 
-def test_periodic_langevin_integrator(splitting="V R H O R V", ncycles=40, nsteps_neq=1000, nsteps_eq=1000):
+def test_periodic_langevin_integrator(splitting="H V R O R V H", ncycles=40, nsteps_neq=1000, nsteps_eq=1000):
```
dominicrufa, line 777: outside of the test, would it be advisable to just put an 'H' step at the beginning (or end) of the protocol?  in the AIS regime, i think 'H' would go at the beginning.

jchodera: Just having `H` at the beginning would make the protocol asymmetric, leading to large errors if you are trying to use BAR with one-step protocols. Placing at the beginning and end makes the protocol symmetric, and the bookkeeping automatically splits the perturbation into two smaller pieces. The main difference will be the total number of energy evaluations per integrator cycle.

jchodera:
The `PeriodicNonequilibriumIntegrator` should now correctly set the alchemical functions at the beginning of the integration cycle.

openmmtools/integrators.py
```diff
@@ -1995,15 +1995,17 @@ def _add_integrator_steps(self):
         self.beginIfBlock('step = 0')
         self.addConstrainPositions()
         self.addConstrainVelocities()
+        self._add_reset_protocol_work_step()
```
dominicrufa, line 1998: okay, so this will reset the protocol work after a cycle is complete...

openmmtools/integrators.py
```diff
@@ -1993,9 +1993,10 @@ def _add_integrator_steps(self):
         """
         # First step: Constrain positions and velocities and reset work accumulators and alchemical integrators
         self.beginIfBlock('step = 0')
-        self._add_update_alchemical_parameters_step() # sets alchemical parameters based on master lambda
         self.addConstrainPositions()
         self.addConstrainVelocities()
+        self._add_reset_protocol_work_step()
+        self._add_alchemical_reset_step() # set alchemical parameters too
```
dominicrufa, line 1999: but this does that same, making the `reset_protocol_work_step` redundant once the cycle is complete

dominicrufa: i guess the only thing to make sure is that we extract the protocol work after every half cycle (forward or backward annealing) because the backward annealing continues to write to the protocol work variable; it might be advisable to zero the `protocol_work` once the `step` has reached `n_steps_eq + n_steps_neq` to avoid having to subtract out forward protocol work to get the backward protocol work?

jchodera:
@dominicrufa : On the odd chance that the [`PeriodicNonequilibriumIntegrator` I use in this script](https://github.com/choderalab/PersesAtFoldingAtHome/pull/1) is to blame for the failures we're seeing, perhaps you could add a couple of tests to this PR just to make sure everything looks good with `PeriodicNonequilibriumIntegrator` to you?

openmmtools/integrators.py
jchodera: The trajectories may contain multiple cycles, so we would lose information that way.

I'm currently just extracting the work by 
```
forward_work = protocol_work[t1] - protocol_work[t0]
backward_work = protocol_work[t3] - protocol_work[t2]
```
etc., so it's pretty straightforward.

And yes, we can definitely remove redundancies and clean up the code here to make it faster and cleaner!

PR #469, opened 2020-05-04
Update neq integrator tests to generate trajectory.

jchodera:
## Description
This adds an option for `test_integrators.py:test_periodic_langevin_integrator` to generate a trajectory for `PeriodicNonequilibriumIntegrator` if requested.

To run:
```python
nosetests -vv --nocapture openmmtools/tests/test_integrators.py:test_periodic_langevin_integrator
```
after setting the `write_trajectory=True` option.


jchodera:
The only failures are those same barostat segfaults, so I'll merge this.

PR #471, opened 2020-05-22
Add travis branches for omnia/label/rc

jchodera:
Add travis test branches for new openmm 7.4.2 rc under allowed_failures


jchodera:
@mikemhenry has already added this.

Issue #480, opened 2020-09-03
Ensemble validation issue for temperature replica exchange


cwalker7:
I am running temperature replica exchange simulations on small coarse-grained oligomer systems (containing a single chain), and have been using an ensemble validation check on the energies for thermodynamic state pairs. This is from the physical-validation python package (https://physical-validation.readthedocs.io/en/latest/userguide.html#ensemble-validation). I am finding that for a system with 12 replicas spaced logarithmically from 200K to 300K, using LangevinDynamicsMove with replica_mixing_scheme='swap-all', n_steps=100, collision frequency=1/ps, and time_step=10fs, the ratio ln(P(U)_1/P(U)_2) is consistently about 4-5 standard deviations off from the analytical value. The physical validation code uses the pymbar DetectEquilibration function to trim the energies for each state. 

I don't think it's an issue of the time step being too large, as reducing to 1fs (with n_steps set to 1000) still leads to 4-5 standard deviations for the above system.

I've tried a number of different replica exchange parameter variations: Upon increasing n_steps to 1000, increasing the friction to 10/ps, or setting the mixing scheme to 'swap-neighbor', there is much better agreement (~1 standard deviation or less from analytical value). Those are comparable to the results for replica_mixing_scheme=None (no exchanges), and for independent simulations run in openmm.

**My question is - is this a matter of simply not equilibrating each replica long enough after an exchange, or could there be an underlying issue somewhere with the exchanges?**

These plots summarize the results, where quantiles is standard deviation from the analytical ratio:
![ensemble_check_langevinmove_coll_rate1_vs_time_step](https://user-images.githubusercontent.com/64790444/92157717-cfab3e00-edf8-11ea-943d-5b8ee3a188eb.png)
![ensemble_check_langevinmove_step10fs_collision1_vs_exch_freq](https://user-images.githubusercontent.com/64790444/92156779-7e4e7f00-edf7-11ea-9a84-02b3e7efc929.png)
![ensemble_check_langevinmove_step10fs_exch_freq1k_vs_coll_rate](https://user-images.githubusercontent.com/64790444/92156781-7e4e7f00-edf7-11ea-9cd6-3bdc5a46f3d1.png)
![ensemble_check_langevinmove_step10fs_exch_freq100_coll_rate1_vs_mixing_scheme](https://user-images.githubusercontent.com/64790444/92156782-7e4e7f00-edf7-11ea-847c-b3c5a2cddc83.png)
[...]

PR #483, opened 2020-09-17
Fix failing tests

zhang-ivy:
## Description
We are seeing the following error in `test_mixing`:
```
  File "/home/runner/work/openmmtools/openmmtools/openmmtools/tests/test_mixing.py", line 11, in <module>
    import openmmtools.multistate.mixing._mix_replicas as mixing
ModuleNotFoundError: No module named 'openmmtools.multistate.mixing._mix_replicas'
```
To fix this, I've added the option to build extension inplace in the `setup.cfg` file.

FYI: There may be more bugs to fix in the failing tests.

~~Note: I'm not sure if this addresses [#480](https://github.com/choderalab/openmmtools/issues/480), but may be worth a shot to see if the large standard deviations disappear?~~
EDIT: Though this change fixed the problem locally, the gh actions tests are still failing.. 

## Todos

## Status
- [ ] Ready to go


Lnaden:
You'll want to use `..` instead of `~` since these builds are always constructed in isolated little directories which are spun up, not necessarily the HOME dir.

Lnaden:
There is also a chance the script is failing because the `temp` directory does not exist. Instead of `working-directory`, try adding the `mkdir` and `cd` lines into the bit before the `nosetest` command.

Not all of the keys in GHA are well documented

jchodera:
This appears to no longer be relevant, so closing. Reopen if necessary.

PR #493, opened 2020-11-10
Make softcore parameters in custom forces hardcoded.

adw62:
This hard codes the soft core parameters in the sterics and electrostatic custom forces. This should make it easier for OpenMM to simplify the function, this is especially useful if you want to take the gradient of these forces.

### DIFF
--- a/openmmtools/alchemy/alchemy.py
+++ b/openmmtools/alchemy/alchemy.py
@@ -516,7 +516,7 @@ class AbsoluteAlchemicalFactory(object):
     switch_width : float, optional, default = 1.0 * angstroms
         Default switch width for electrostatics in periodic cutoff systems
         used in alchemical interactions only.
-    alchemical_pme_treatment : str, optional, default = 'direct-space'
+    alchemical_pme_treatment : str, optional, default = 'exact'
         Controls how alchemical region electrostatics are treated when PME is used.
         Options are ['direct-space', 'coulomb', 'exact'].
         - 'direct-space' only models the direct space contribution


jchodera:
Yikes! Has it really been a year since we've been able to fix these things? Huge apologies---COVID has had us really distracted.

Exposing these parameters as global parameters can be useful in many cases, such as automated optimization of the alchemical pathway or dynamics in alchemical parameter space. A better way to address this would be to hard-code all  parameters by default, but optionally allow the designated parameters to be global parameters. That way, maximum performance could be obtained for the desired use case.

adw62:
No problem. Some more info if it's useful: in my testing these changes to the code gave more stable calculations of the gradients but when it came to production time I was still getting spurious result from the analytic gradient. I ended up using the numerical gradient in my code. I'm not sure what the remaining problem was, if I can dig out an example case I'll post it here.

jchodera:
Here's how we will create a new PR based on this idea:

* The default values of parameters come from the `AlchemicalRegion`
* The factory will accept a list of global parameters the user wants to expose as *adjustable* global parameters (e.g. `["softcore_beta", "softcore_e"]`.
* The Lepton energy expressions would still be constructed starting with the fully general expression like
```python
energy_electrostatics = "sigma*((softcore_beta*(1.0-({0}))^softcore_e + (r/sigma)^softcore_f))^(1/softcore_f);"
```
* Parameters that should _not_ be exposed as global variables will then be appended as constants:
```python
for parameter_name in nonglobal_parameters:
   energy_expression += "{parameter_name} = {getattr(alchemical_region, parameter_name)};"
```
* Parameters that _should_ be exposed as global variables will be set using `addGlobalVariable`:
```python
for parameter_name in global_parameters:
    force.addGlobalParameter(parameter_name, getattr(alchemical_region, parameter_name))
```

PR #494, opened 2020-11-14
Bump setup-miniconda to v2

jaimergp:
Just bumping setup-miniconda to v2 to workaround those GH deprecation warnings that will error out on Nov 16


jaimergp:
These failures are known bugs and being worked on separately. Merging.

PR #504, opened 2021-03-17
remove cython from setup.py

mikemhenry:
## Description
Remove cython from setup.py

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -14,7 +14,6 @@ dependencies:
   - python
   - numpy
   - scipy
-  - six
   - parmed
   - mdtraj
   - netcdf4 >=1.4.2 # after bugfix: "always return masked array by default, even if there are no masked values"
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - python
     - numpy
     - scipy
-    - six
     - openmm >=7.3.1
     - parmed
     - mdtraj
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -9,7 +9,6 @@ dependencies:
     - cython
     - numpy
     - scipy
-    - six
     - parmed
     - mdtraj
     - numpydoc
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -1,6 +1,10 @@
 Release History
 ***************
 
+0.20.2 - Remove leftover support for python 2.7
+
+- Remove leftover `six` imports and `xrange` (`#504 <https://github.com/choderalab/openmmtools/pull/504>`_)
+
 0.20.1 - Bugfix release
 ========================================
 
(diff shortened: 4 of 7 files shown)


mikemhenry:
@jaimergp anything else that should end up in this PR?

Regardless on what we decide on https://github.com/conda-forge/openmmtools-feedstock/pull/6 we should make a 0.20.2 release that does things better!

jaimergp:
Now that you mention it, I'd like to remove the `six` dependency. We claim to support Pyrthon 3.6+ only, but https://github.com/choderalab/openmmtools/blob/master/openmmtools/sobol.py has this weird `xrange` py2k->py3k leftover. Removing that `six` import and replacing `range` with `xrange` would be nice to have here too.

Lnaden:
> six import and replacing `range` with `xrange`

Did you mean "replacing `xrange` with `range`" instead?

jaimergp:
> > six import and replacing `range` with `xrange`
> 
> Did you mean "replacing `xrange` with `range`" instead?

Oops, yes, exactly.

mikemhenry:
That sounds good, I don't want to explode this release too much since there is some python 2.7 cruft (see `openmmtools/utils.py` line 691) But I can remove the dependency and use range.

Lnaden:
> (see openmmtools/utils.py line 691)

That works for Py3 as well, so there isnt explicit harm in leaving it for now, but it probably could be cleaned up at a future time. 

> But I can remove the dependency and use range.

Yes to this. Anything which further and fully severs our ties to Python 2 the better, that is until we have to start using `twelve`.

mikemhenry:
Agreed! I'll add @jaimergp and @Lnaden as reviewers -- let me know if there are any more changes you think we should make.

Lnaden (approved):
LGTM. We'll keep this simple and move on.

jaimergp:
We should note that the last conda-forge release depends on `ambertools`, which is not available on Windows. I am not sure where this is used in the package, but maybe it should be reassessed? If we have find something, we need soft-imports and safeguards for missing executables. 

There's also another `xrange` instance in `openmmtools/scripts/test_openmm_platforms.py`.

Lnaden:
The ambertools stuff comes from the `data` dirs: `https://github.com/choderalab/openmmtools/tree/483765d3b87395834745e79275469ce6c2c3fa13/openmmtools/data`

jaimergp:
Those scripts are not run often, right?

jaimergp:
I am also seeing `mdtraj` and `pandas` required in `testsystems`, but they are try-except guarded. The OpenEye toolkit is also there, but not available on `conda-forge`.

mikemhenry:
> There's also another `xrange` instance in `openmmtools/scripts/test_openmm_platforms.py`.

I don't see it? What line?

mikemhenry:
> The OpenEye toolkit is also there, but not available on `conda-forge`.

I dug a bit and while that will throw an error -- it isn't clear to me how to trivially replace the functionality or disable the functionality (it is really limited in use and a quick grok shows that it is used in test systems) but I can throw a different error if we want other than the one openeye will throw when it says it is unlicensed. 

Is this fix something you wanted in this patch or can we make this an issue for another day @jaimergp ?

jaimergp:
> I don't see it? What line?

Ah, you got it already! Disregard then.

> Is this fix something you wanted in this patch or can we make this an issue for another day?

Definitely a different issue.

mikemhenry:
Okay! I will merge this and cut a new release, thank you both for all of your help!

mikemhenry:
ALMOST FORGOT TO UPDATE THE CHANGELOG!

PR #506, opened 2021-03-23
Fix critical bug in which GPU contexts would fail to recognize mixed precision was available

### LINKED ISSUE #505: On newer versions of openmmtools, GPU can't be used if using mixed precision due to mixed precision check
https://github.com/choderalab/openmmtools/blob/3f2cfc0d9ea5ee69b270b1f7705867548f8ca037/openmmtools/utils.py#L513

This function always fails due to this part:

```python
    if platform.getName() in ['CUDA', 'OpenCL']:
        from simtk import openmm
        properties = { 'Precision' : precision }
        system = openmm.System()
        integrator = openmm.VerletIntegrator(0.001)
        try:
            context = openmm.Context(system, integrator, properties)
            del context, integrator
            return True
        except Exception as e:
            return False
```

It always returns 

```
TypeError: Wrong number or type of arguments for overloaded function 'new_Context'.
  Possible C/C++ prototypes are:
    OpenMM::Context::Context(OpenMM::System const &,OpenMM::Integrator &)
    OpenMM::Context::Context(OpenMM::System const &,OpenMM::Integrator &,OpenMM::Platform &)
    OpenMM::Context::Context(OpenMM::System const &,OpenMM::Integrator &,OpenMM::Platform &,std::map< std::string,std::string,std::less< std::string >,std::allocator< std::pair< std::string const,std::string > > > const &)
    OpenMM::Context::Context(OpenMM::Context const &)
```

Try it:

```python
precision = "mixed"
from simtk import openmm
properties = { 'Precision' : precision }
system = openmm.System()
integrator = openmm.VerletIntegrator(0.001)
context = openmm.Context(system, integrator, properties)
```

Thus, whenever configuring the platform, if trying to check available platforms that allow mixed precision, the CUDA and OpenCL ones will always fail.

jchodera:
## Description
This PR fixes #505, where GPU contexts would fail to recognize mixed precision was available.
This may mean that applications were silently failing to use mixed precision on GPUs.

I fixed the logic to match the function docstring, but I'm not sure if this will break anything else. We'll have to wait for the tests to see.
I've verified the new `test_platform_supports_precision()` test passes on an NVIDIA GTX 1080.

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -1,8 +1,20 @@
 Release History
 ***************
 
-0.20.2 - Remove leftover support for python 2.7
+0.20.3 - Bugfix release
+======================
 
+Bugfixes
+--------
+- Fixes [#505](https://github.com/choderalab/openmmtools/issues/505): GPU contexts would silently fail to enable 'mixed' precision; corrects reporting of available precisions
+
+0.20.2 - Bugfix release
+=======================
+
+Remove leftover support for python 2.7
+
+Cleanup
+-------
 - Remove leftover `six` imports and `xrange` (`#504 <https://github.com/choderalab/openmmtools/pull/504>`_)
 
 0.20.1 - Bugfix release
--- a/openmmtools/tests/test_utils.py
+++ b/openmmtools/tests/test_utils.py
@@ -21,6 +21,28 @@ from openmmtools.utils import *
 
 
 # =============================================================================
+# TEST CONTEXT UTILITIES
+# =============================================================================
+
+def test_platform_supports_precision():
+    """Test that platform_supports_precision works correctly."""
+    from simtk import openmm
+    
+    for platform_index in range(openmm.Platform.getNumPlatforms()):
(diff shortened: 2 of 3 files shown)


mikemhenry:
Works on my `RTX 2060` as well!

mikemhenry (commented):
LGTM!

mikemhenry:
I'm running all the tests locally now and once they pass I will merge this + cut another release! 

@jchodera do we have a mailing list to announce new releases or should I use the lab account and tweet we have a new release and users should upgrade?

mikemhenry (approved):

jchodera:
> @jchodera do we have a mailing list to announce new releases or should I use the lab account and tweet we have a new release and users should upgrade?

We don't have a mailing list, but let's make sure we have folks in the lab test it with perses to make sure it doesn't break anything first. I can tweet from the openmm twitter account later!

Issue #510, opened 2021-05-12
Charmm36 nonbonded forces with Alchemy


AJZein:
Hello,

I am trying to perform alchemy using the charmm36 forcefield but I've noticed that there are remaining forces when lambda is set to 0. In the LysozymeImplicit test system for example, the pxylene_atoms wont separate from its binding pocket when lambda is set to 0 in charmm36, but it will when using the test system's forcefield. I've added a minimal example for both cases at the bottom.

I've tried looking into the problem and I believe it might be due to the LJ Parameters being implemented in a CustomNonbondedForce when using charmm36, which allows for adding nonbonded fix (NBFIX) as a lookup table. However I don't know how to get alchemy to work with this.

Thanks for any help,
Zein

Working example
```
from openmmtools import alchemy, testsystems, integrators
import simtk.openmm.app as app
from sys import stdout

# Create the reference OpenMM System that will be alchemically modified.
lysozyme_pxylene = testsystems.LysozymeImplicit()
t4_system = lysozyme_pxylene.system


# Define the region of the System to be alchemically modified.
pxylene_atoms = lysozyme_pxylene.mdtraj_topology.select('resname TMP')
alchemical_region = alchemy.AlchemicalRegion(alchemical_atoms=pxylene_atoms)

# Set up simulation
factory = alchemy.AbsoluteAlchemicalFactory()
alchemical_system = factory.create_alchemical_system(t4_system, alchemical_region)
integrator = integrators.LangevinIntegrator()
simulation = app.Simulation(lysozyme_pxylene.topology, alchemical_system, integrator)
simulation.context.setPositions(lysozyme_pxylene.positions)

# Add reporting
simulation.reporters.append(app.StateDataReporter(stdout, 1000, step=True, potentialEnergy=True, temperature=True))
app.PDBFile.writeFile(simulation.topology, lysozyme_pxylene.positions, open(f'working.pdb', 'w'))
simulation.reporters.append(app.DCDReporter(f'working.dcd', 1000))

# Remove interactions by setting lambda=0
alchemical_state = alchemy.AlchemicalState.from_system(alchemical_system)
alchemical_state.lambda_electrostatics = 0
alchemical_state.lambda_sterics = 0
alchemical_state.apply_to_context(simulation.context)

simulation.step(20000)
```

Failing example with charmm36
```
from openmmtools import alchemy, testsystems, integrators
import simtk.openmm.app as app
from sys import stdout

[...]

PR #511, opened 2021-05-26
Catch LennardJonesGenerator made NB functions.

Lnaden:
## Description
Correctly raise a NYI error when the nonbonded force's LJ terms are
handled by a tabulated potential generated by the OpenMM forcefield.py's
LennardJonesGenerator function (Such as CHARMM36). This will prevent
the alchemical factory from continuing and act like everything is
correct when it isn't.

Temporary catch until #510 can be fixed.

Fix the case there is a CustomNB force that doesn't have the tabulated
expression.

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/alchemy.py
+++ b/openmmtools/alchemy.py
@@ -2031,6 +2031,37 @@ class AbsoluteAlchemicalFactory(object):
             forces_by_lambda[''] = [nonbonded_force]
         return forces_by_lambda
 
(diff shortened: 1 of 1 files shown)


jchodera (approved):
Thanks! This is a good solution for safety now. See #510 for my notes on how we can properly address this.

jchodera:
Whoops---forgot to catch that we should update the changelog too!

PR #515, opened 2021-06-29
Remove logic for missing file when retrying to open a dataset

mikemhenry:
## Description

What should we do about these now defunct arguments to `_open_dataset_robustly`
`catch_io_error=False` and `io_error_warning=None`? Since the method is private, we should be able to just remove them from the function signature and update the code where we use it, thoughts?

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -362,13 +362,6 @@ class MultiStateReporter(object):
         for attempt in range(n_attempts-1):
             try:
                 return netcdf.Dataset(*args, **kwargs)
-            except IOError as e:
-                # If the file does not exist, it doesn't make sense to try again.
-                if catch_io_error:
-                    if io_error_warning is not None:
-                        logger.warning(io_error_warning)
-                    return None
-                raise e
             except:
                 logger.debug('Attempt {}/{} to open {} failed. Retrying '
                              'in {} seconds'.format(attempt+1, n_attempts, sleep_time))
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -360,11 +360,14 @@ class MultiStateReporter(object):
         """
 
         # Check if file exists and warn if asked
-        if catch_io_error:
-            if not os.path.isfile(args[0]):
+        # raise IOError otherwise
+        if not os.path.isfile(args[0]):
+            if catch_io_error:
                 if io_error_warning is not None:
                     logger.warning(io_error_warning)
                 return None
+            raise IOError
+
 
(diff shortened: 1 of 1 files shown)


openmmtools/multistate/multistatereporter.py
```diff
@@ -362,13 +362,6 @@ def _open_dataset_robustly(self, *args, n_attempts=5, sleep_time=2,
...
-            except IOError as e:
-                # If the file does not exist, it doesn't make sense to try again.
-                if catch_io_error:
-                    if io_error_warning is not None:
-                        logger.warning(io_error_warning)
-                    return None
-                raise e
             except:
                 logger.debug('Attempt {}/{} to open {} failed. Retrying '
                              'in {} seconds'.format(attempt+1, n_attempts, sleep_time))
```
jchodera, line 367: There are four `{}` items in this string but only three arguments passed, which is now causing test failures:
```
ERROR: Test that the analysis file can open without the checkpoint file
  File "/usr/local/miniconda/envs/test/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/runner/work/openmmtools/openmmtools/openmmtools/tests/test_sampling.py", line 1340, in test_analysis_opens_without_checkpoint
    self.REPORTER(storage_path, checkpoint_storage=cp_file_mod, open_mode='r')
  File "/Users/runner/work/openmmtools/openmmtools/openmmtools/multistate/multistatereporter.py", line 136, in __init__
    self.open(open_mode)
  File "/Users/runner/work/openmmtools/openmmtools/openmmtools/multistate/multistatereporter.py", line 302, in open
    version=netcdf_format)
  File "/Users/runner/work/openmmtools/openmmtools/openmmtools/multistate/multistatereporter.py", line 367, in _open_dataset_robustly
    'in {} seconds'.format(attempt+1, n_attempts, sleep_time))
IndexError: tuple index out of range
```

mikemhenry: Your right, but what is weird is the code I removed didn't change that, so this exception would never have worked. I'm going to fix the code + double check what the tests are doing now as well as remove those defunct kwargs.

mikemhenry:
Should be good to go! I had to switch the order we check if a file exits since this function is also used to create files. Now we try and open it a few times, check to see if it exits, then set the envar to disable locking and try one more time. That should fix the issue @zhang-ivy was having.

mikemhenry:
@jchodera when you get the chance, can you give this another look?

mikemhenry:
@jchodera just a friendly bump to check this when you got a chance

jchodera (approved):
LGTM!

PR #524, opened 2021-10-26
Overriding deepcopying behavior to avoid creating new ContextCaches.

### LINKED ISSUE #521: Getting `CUDA_ERROR_OUT_OF_MEMORY` error when running replica exchange (with lots of replicas) on large system
I'm trying to run replica exchange (rest) on a system with 185K atoms, and it runs fine when I use a rest region of ~70 atoms with 12 replicas. However, I get the following error when I run with a rest region of ~1000 atoms with 24 or 36 replicas.

I'm currently using this as the context cache:
`context_cache = cache.ContextCache(capacity=None, time_to_live=None)`

```
DEBUG:openmmtools.utils:Mixing of replicas took    3.747s
DEBUG:openmmtools.multistate.replicaexchange:Accepted 12284/93312 attempted swaps (13.2%)
DEBUG:openmmtools.multistate.multistatesampler:Propagating all replicas...
DEBUG:mpiplus.mpiplus:Running _propagate_replica serially.
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/cache.py", line 445, in get_context
    context = self._lru[context_id]
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/cache.py", line 147, in __getitem__
    entry = self._data.pop(key)
KeyError: (366470656340979585, 4946973717435834380)

During handling of the above exception, another exception occurred:

  File "generate_rest2_cache_interface.py", line 153, in <module>
    simulation.run()
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/multistate/multistatesampler.py", line 681, in run
    self._propagate_replicas()
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/utils.py", line 90, in _wrapper
    return func(*args, **kwargs)
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/multistate/multistatesampler.py", line 1196, in _propagate_replicas
    propagated_states, replica_ids = mpiplus.distribute(self._propagate_replica, range(self.n_replicas),
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/mpiplus/mpiplus.py", line 512, in distribute
    all_results = [task(job_args, *other_args, **kwargs) for job_args in distributed_args]
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/mpiplus/mpiplus.py", line 512, in <listcomp>
    all_results = [task(job_args, *other_args, **kwargs) for job_args in distributed_args]
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/multistate/multistatesampler.py", line 1223, in _propagate_replica
    mcmc_move.apply(thermodynamic_state, sampler_state)
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/mcmc.py", line 1114, in apply
    super(LangevinDynamicsMove, self).apply(thermodynamic_state, sampler_state)
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/mcmc.py", line 655, in apply
    context, integrator = context_cache.get_context(thermodynamic_state, integrator)
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/cache.py", line 447, in get_context
    context = thermodynamic_state.create_context(integrator, self._platform, self._platform_properties)
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/openmmtools/states.py", line 1172, in create_context
    return openmm.Context(system, integrator)
  File "/home/zhangi/miniconda3/envs/perses-rbd-ace2/lib/python3.8/site-packages/simtk/openmm/openmm.py", line 4948, in __init__
    _openmm.Context_swiginit(self, _openmm.new_Context(*args))
simtk.openmm.OpenMMException: Error creating array savedForces: CUDA_ERROR_OUT_OF_MEMORY (2)
DEBUG:mpiplus.mpiplus:Single node: executing <bound method MultiStateReporter.close of <openmmtools.multistate.multistatereporter.MultiStateReporter object at 0x2b547fb44ac0>>
```

To reproduce the issue, I can point to where the files are located on lilac (the files are large, so it doesn't make sense to drop them here):
In `/data/chodera/zhangi/perses_benchmark/neq/14/147/for_debugging/`:
- ` perses-rbd-ace2-direct.yml` -- contains a yaml of the env I'm using
- `generate_rest_cache_interface.py` -- rest script
- `run_rest2_complex_for_debugging.sh` -- bash script
- `147_complex_1.pickle `-- pickled RepartitionedHybridTopologyFactory needed for rest script
- `147_complex_1_rest.pickle` -- pickled RESTTopologyFactory needed for rest script
- Note that to use the above two pickled factories with the rest script, you'll need to create a dir called `147/` and place the factories and scripts in that directory
    - You'll also need to make sure you set outdir in the rest script to point to your `147/` directory
There are also xmls for the system (the REST system), state (at thermodynamic state with temp 298 K), and integrator (attached to the context cache --though i dont think this is actually being used anywhere).

Some observations I made:
- The error happens more often when I use job arrays than when I don't use job arrays. However, I do see the error in both scenarios.
- From glancing at which GPUs have been pulled for my jobs, it doesn't seem like one type is responsible for the error. I've seen the error happen on lt, ly, lx , and lu . I've also seen absence of an error on these nodes.

ijpulidos:
## Description
When you specify a context_cache to be used in the mcmc_move for the simulation it gets deep copied into every replica. This avoid creating new ContextCache for each replica, instead it uses a reference to the same given and hopefully compatible ContextCache.

Fixes #521 

## Todos
- [ ] Check how this affects serialization of objects.

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -1279,6 +1279,21 @@ class LangevinSplittingDynamicsMove(LangevinDynamicsMove):
                                               measure_shadow_work=self.measure_shadow_work,
                                               measure_heat=self.measure_heat)
 
+    def __deepcopy__(self, memo):
+        """Overriding default deep copy behavior to avoid creating new ContextCache objects without need."""
+
+        if memo is None:
+            memo = {}
+
+        cls = self.__class__
+        new_state = cls.__new__(cls)
+        memo[id(self)] = new_state
+        for k, v in self.__dict__.items():
+            if k != 'context_cache':
+                new_state.__dict__[k] = copy.deepcopy(v, memo)
+        new_state.__dict__['context_cache'] = self.context_cache
+        return new_state
+
 
 # =============================================================================
 # GENERALIZED HYBRID MONTE CARLO MOVE
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -762,6 +762,21 @@ class BaseIntegratorMove(MCMCMove):
         else:
             self.context_cache = utils.deserialize(serialization['context_cache'])
 
+    def __deepcopy__(self, memo):
+        """Overriding default deep copy behavior to avoid creating new ContextCache objects without need."""
+
+        if memo is None:
+            memo = {}
+
+        cls = self.__class__
+        new_state = cls.__new__(cls)
+        memo[id(self)] = new_state
+        for k, v in self.__dict__.items():
+            if k != 'context_cache':
+                new_state.__dict__[k] = copy.deepcopy(v, memo)
+        new_state.__dict__['context_cache'] = self.context_cache
+        return new_state
+
 
 # =============================================================================
 # METROPOLIZED MOVE BASE CLASS
@@ -1278,21 +1293,6 @@ class LangevinSplittingDynamicsMove(LangevinDynamicsMove):
                                               constraint_tolerance=self.constraint_tolerance,
                                               measure_shadow_work=self.measure_shadow_work,
                                               measure_heat=self.measure_heat)
-
-    def __deepcopy__(self, memo):
-        """Overriding default deep copy behavior to avoid creating new ContextCache objects without need."""
-
-        if memo is None:
-            memo = {}
-
-        cls = self.__class__
-        new_state = cls.__new__(cls)
-        memo[id(self)] = new_state
-        for k, v in self.__dict__.items():
-            if k != 'context_cache':
-                new_state.__dict__[k] = copy.deepcopy(v, memo)
-        new_state.__dict__['context_cache'] = self.context_cache
-        return new_state
 
 
 # =============================================================================
--- a/openmmtools/tests/test_mcmc.py
+++ b/openmmtools/tests/test_mcmc.py
@@ -13,7 +13,9 @@ Test State classes in mcmc.py.
 # GLOBAL IMPORTS
 # =============================================================================
 
+import math
 import pickle
+import tempfile
 from functools import partial
 
 import nose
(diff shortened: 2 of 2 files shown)


mikemhenry:
This PR doesn't fix the issue with cycling behavior, but it also doesn't break our current work around that issue. Since it seems to fix the issue with memory, we should merge in this fix and keep exploring the root cause of the cycling issue when using a global context cache defined at the module level. 

I think the place to dive in is to ask:
How are the code paths different when using the context cache defined in the module instead of instantiating my own instance?

We already found one difference, which is the way the object gets copied, are there more?

ijpulidos:
@mikemhenry I totally agree. We have yet to figure out why this fix doesn't show the cycling behavior as with using the global context cache. I don't really see a difference, we define `global_context_cache` [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L775).

But then if we manually specify a context_cache using keywords, `context_cache = cache.ContextCache(capacity=None, time_to_live=None)` (which is the same as the previous `global_context_cache` is defined), it works. And I don't see why the behavior would fork, since the only difference is [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/mcmc.py#L645-L648), as far as I can see.

ijpulidos:
I'll try to move this to the base class.

ijpulidos:
I was thinking about this and I'm not totally sure this is what we want. Is there a case where we DO want context caches to be different for different replicas? What about when we use the `DummyContextCache`? Do we want to share the same `DummyContextCache` for all replicas as well in this case?

This does fix the memory error in #521 but I'm not sure if this is exactly what we want everytime. Even though the tests seem to be passing.

jchodera:
> I was thinking about this and I'm not totally sure this is what we want. Is there a case where we DO want context caches to be different for different replicas? 

While I doubt it, there may be a use case where we _do_ want separate context caches for different replicas on different hardware, but we never want a `ContextCache` that was not explicitly constructed to replicate itself under the hood. Each `ContextCache` should fully own the device it is using---otherwise, there is no point to its existence.

> What about when we use the DummyContextCache? Do we want to share the same DummyContextCache for all replicas as well in this case?

If I recall correctly, the `DummyContextCache` takes no configuration arguments and creates-and-destroys single-use `Context`s, so it does not matter if there are one or many of these objects.

ijpulidos:
@jchodera Do you know a quick way to test serialization here? That's the only remaining test I would like to do just to be sure this PR is safe. Thanks!

mikemhenry (approved):
LGTM! :tada:

PR #528, opened 2021-11-14
deal with openmm 7.6 package unwrapping

sroet:
## Description
importing OpenMMTools raises warinings about importing `openmm` from `simtk`, which make OpenPathSampling's notebook tests fail (unexpected warning in the stderr output) 

This updates all imports to the new standard.

I went through  `grep -r -n "simtk" --include="*.py" -l` and updated all references to `simtk`. For the doc-tests I decided to only include the current advised method of importing openmm. For all real code, I wrapped the new method in a try/except clause that drops back to the old style imports.

## Todos
- [NA] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)
- [NA] Update [documentation](https://github.com/choderalab/openmmtools/tree/master/docs) as needed

## Status
- [x] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -7,6 +7,7 @@ Release History
 Enhancement
 -----------
 - Use of CODATA 2018 constants information from OpenMM 7.6.0. (#522)
+- Use new way of importing OpenMM >= 7.6. (#528)
 
 0.20.3 - Bugfix release
 ======================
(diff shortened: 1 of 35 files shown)


jchodera:
Thanks so much for updating this in a backwards-compatible manner!

@ijpulidos can you review and merge when able?

sroet:
test failure is identical to master. (and is for an unsupported python, based on [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html))

jchodera:
Thanks! And thanks for pointing out [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)---I hadn't seen that before, but we will try to standardize around that!

ijpulidos (approved):
This looks great, thanks!

PR #537, opened 2021-12-24
Shared context cache deserialization on resume

### LINKED ISSUE #536: Upon resuming a simulation from a .nc, new context caches are created for each move
This is a follow up to the fix here: https://github.com/choderalab/openmmtools/pull/524, which prevents the creation of a new context cache for each move.

The fix works fine for starting a simulation, but when I resume a simulation from a .nc file, I run into memory issues. 

This is because [when mcmc moves are read from the reporter](https://github.com/choderalab/openmmtools/blob/master/openmmtools/multistate/multistatereporter.py#L752-L768), each move is deserialized (`deserialize()` defined [here](https://github.com/choderalab/openmmtools/blob/456af23d91d12d570ec755674716ab0003b69e5b/openmmtools/utils.py#L652)).  Each `deserialize()` call involves a `__setState__()` call. For `BaseIntegratorMove`, that's defined [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/mcmc.py#L760-L767). In the `BaseIntegratorMove`'s  `__setState__()` , the context cache is then deserialized. However, in the `ContextCache`'s `__setState__()` (defined [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L472-L482)), a new context cache is created each time. This means that when deserializing the mcmc moves from the reporter, we are re-introducing the problem that the aforementioned PR solves (i.e. each move now has its own context cache).

Btw, you can check that each move has a separate context cache from this code:
```python
from openmmtools.multistate import MultiStateReporter
path = '4_apo.nc'
reporter = MultiStateReporter(path, 'r')
moves = reporter.read_mcmc_moves()
print(id(moves[0].context_cache))
print(id(moves[1].context_cache))
```

Attached are files to reproduce: 
[debug-resume-memory-issue.zip](https://github.com/choderalab/openmmtools/files/7738057/debug-resume-memory-issue.zip)

You can try resuming the simulation with this code:
```python
import argparse
from openmmtools.multistate import MultiStateReporter
from perses.samplers.multistate import HybridRepexSampler
import logging
import os

# Load repex simulation
_logger = logging.getLogger()
_logger.setLevel(logging.DEBUG)
reporter_file = "4_apo.nc"
reporter = MultiStateReporter(reporter_file, checkpoint_interval=10)
simulation = HybridRepexSampler.from_storage(reporter)

# Determine how many more iterations are needed
total_iterations = 5000
iterations =  total_iterations - simulation.iteration

# Resume simulation
simulation.extend(n_iterations=iterations)
```

ijpulidos:
## Description
Fixes #536

Added helper functions to tests in new `test/utils.py` file (is this a good place?).

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -762,9 +762,18 @@ class MultiStateReporter(object):
 
         # Retrieve all moves in order.
         mcmc_moves = list()
-        for move_id in range(n_moves):
+        # manually add first move and get context from it
+        serialized_move = self.read_dict('mcmc_moves/move0')  # first serialized move
+        deserialized_move = deserialize(serialized_move)
+        mcmc_moves.append(deserialized_move)
+        # get context_cache of first move to use in the rest of the moves
+        first_context_cache = deserialized_move.context_cache
+        # loop through the rest of the moves
+        for move_id in range(1, n_moves):
             serialized_move = self.read_dict('mcmc_moves/move{}'.format(move_id))
-            mcmc_moves.append(deserialize(serialized_move))
+            deserialized_move = deserialize(serialized_move)
+            deserialized_move.context_cache = first_context_cache  # Use same context_cache in all the moves
+            mcmc_moves.append(deserialized_move)
         return mcmc_moves
 
     def write_mcmc_moves(self, mcmc_moves):
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -761,6 +761,7 @@ class BaseIntegratorMove(MCMCMove):
         self.n_steps = serialization['n_steps']
         self.reassign_velocities = serialization['reassign_velocities']
         self.n_restart_attempts = serialization['n_restart_attempts']
+        # TODO: self.context_cache defaults to None in constructor, do we need this check?
         if serialization['context_cache'] is None:
             self.context_cache = None
         else:
--- a/openmmtools/tests/test_mcmc.py
+++ b/openmmtools/tests/test_mcmc.py
@@ -310,7 +310,31 @@ def test_moves_shared_contextcache_resume():
                                                             n_steps=10,
                                                             n_iterations=10,
                                                             checkpoint_interval=5,
-                                                            do_run=True)
+                                                            do_run=True,
+                                                            local_context_cache=True)
+        # Resume simulation from storage and check identity
+        reporter = MultiStateReporter(storage_path, 'r')
+        moves = reporter.read_mcmc_moves()
+        id_1 = id(moves[0].context_cache)
+        id_2 = id(moves[1].context_cache)
+        assert id_1 == id_2, f"Identities mismatch between context cache objects."
+
+
+def test_moves_shared_none_contextcache_resume():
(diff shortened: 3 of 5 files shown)


openmmtools/multistate/multistatereporter.py
```diff
@@ -762,9 +762,18 @@ def read_mcmc_moves(self):
...
+        mcmc_moves.append(deserialized_move)
+        # get context_cache of first move to use in the rest of the moves
+        first_context_cache = deserialized_move.context_cache
+        # loop through the rest of the moves
+        for move_id in range(1, n_moves):
             serialized_move = self.read_dict('mcmc_moves/move{}'.format(move_id))
-            mcmc_moves.append(deserialize(serialized_move))
+            deserialized_move = deserialize(serialized_move)
+            deserialized_move.context_cache = first_context_cache  # Use same context_cache in all the moves
+            mcmc_moves.append(deserialized_move)
```
ijpulidos, line 776: Unrolled the loop and extracted first move context_cache to used it in others. It does the trick but I must confess it feels a bit dirty. I don't have better ideas, though.

ijpulidos:
We need to let the tests pass first (passing locally). But @zhang-ivy when you have the chance, can you test this branch's code in a venv in your system and confirm the memory issue is not happening after resuming your simulations? As a reminder, you might have to install it using pip with something like `pip install "git+https://github.com/choderalab/openmmtools.git@shared-context-from-resume"` and check you get `openmmtools               0.20.3+18.gf179898          pypi_0    pypi` when you use `conda list | grep openmmtools`.

ijpulidos:
Ok, tests are not passing and I think I know why. There are possible moves that don't have a `context_cache`, i.e. `SequenceMove`. I have to fix this because I am assuming moves have this attribute.

zhang-ivy:
@ijpulidos : I tested this branch out -- it fixes my issue! The fix looks fine to me -- I can't think of a better solution.

@jchodera : what do you think?

jchodera:
@ijpulidos : Shouldn't the choice of which context caches to use be a choice made at runtime, and have nothing to do with the serialized form? I'm honestly not sure we ever intended for `ContextCache` objects to be stored when serialized to disk.

I think the correct behavior here is that both the creation of a new simulation from scratch and the resuming a simulation should 
* use the user-specified caches if provided, or failing that
* call the same helper method to set up MCMCMove and energy evaluation `ContextCache` objects 

Does that make sense? Happy to set up a quick video chat to talk through it otherwise!

ijpulidos:
@jchodera It does makes sense, but I think I am going to think about this a bit more, I'm checking how the serialization and deserialization of these objects is happening in detail and will let you know. Because there is metadata for the `context_cache` attribute of the mcmc move getting serialized, but not the objects themselves.

jchodera:
We should not be serializing anything about runtime execution choices (which is what `ContextCache` is---a runtime execution optimization) in the serialized form of the calculation. If we're serializing anything about runtime execution choices, it's a mistake, and we should eliminate it.

ijpulidos:
Closing this one. Superseded by #538

PR #538, opened 2022-01-13
ContextCache usage cleanup

### LINKED ISSUE #519: Issue with ContextCache
See this issue
https://github.com/choderalab/perses/issues/613

And this comment:
https://github.com/choderalab/perses/issues/613#issuecomment-902512392

The current workaround is to create a instance of `ContextCache` and use that instead of the default global one, ie.

```python
context_cache = cache.ContextCache(capacity=None, time_to_live=None)
move = mcmc.LangevinSplittingDynamicsMove(
    timestep=4.0 * unit.femtoseconds,
    n_steps=250,
    collision_rate=5.0 / unit.picosecond,
    reassign_velocities=False,
    n_restart_attempts=20,
    constraint_tolerance=1e-06,
    context_cache=context_cache,  # This is the important part of the example
)
```

### LINKED ISSUE #536: Upon resuming a simulation from a .nc, new context caches are created for each move

ijpulidos:
## Description
Given that the `ContextCache` is a performance _trick_. It is better that it doesn't get serialized at all and that it is only used when the moves are applied or energies are computed. 

The `mcmc` module classes won't have a `context_cache` kwarg, instead the `apply` method for each will get fed the `context_cache` from the `multistatesampler` module objects.

This should tackle the memory issues when dealing with multiple replicas (fixes #536  ) and also work around the cycling issue (fixes #519 ). 

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -148,7 +148,7 @@ class MCMCMove(SubhookedABCMeta):
     """
 
     @abc.abstractmethod
-    def apply(self, thermodynamic_state, sampler_state):
+    def apply(self, thermodynamic_state, sampler_state, context_cache):
         """Apply the MCMC move.
 
         Depending on the implementation, this can alter the thermodynamic
@@ -162,6 +162,7 @@ class MCMCMove(SubhookedABCMeta):
         sampler_state : openmmtools.states.SamplerState
            The initial sampler state before applying the move. This may
            be modified depending on the implementation.
+        context_cache : opemmtools.cache.ContextCache
 
         """
         pass
@@ -631,7 +632,7 @@ class BaseIntegratorMove(MCMCMove):
            The thermodynamic state to use to propagate dynamics.
         sampler_state : openmmtools.states.SamplerState
            The sampler state to apply the move to. This is modified.
-        context_cache : openmmtools.context.ContextCache
+        context_cache : openmmtools.cache.ContextCache
             context cache to be used during propagation with the integrator.
 
         See Also
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -201,7 +201,7 @@ class MCMCMove(SubhookedABCMeta):
             context_cache = context_cache_input
         else:
             raise ValueError("Context cache input is not a valid ContextCache or None type.")
-        # update private attribute
+        # update private attribute - expected ONLY to be updated by this method!
         self._propagation_context_cache = context_cache
         return context_cache
 
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -420,8 +420,11 @@ class SequenceMove(MCMCMove):
         context_cache : opemmtools.cache.ContextCache, optional
 
         """
+        # Get context cache to be used and update attribute
+        self.context_cache = self._get_context_cache(context_cache)
         for move in self.move_list:
-            move.apply(thermodynamic_state, sampler_state, context_cache=context_cache)
+            # Apply each move with the specified local context
+            move.apply(thermodynamic_state, sampler_state, context_cache=self.context_cache)
 
     def __str__(self):
         return str(self.move_list)
@@ -511,7 +514,9 @@ class WeightedMove(MCMCMove):
         """
         moves, weights = zip(*self.move_set)
         move = np.random.choice(moves, p=weights)
-        move.apply(thermodynamic_state, sampler_state, context_cache=context_cache)
+        # Get context cache to be used and update attribute
+        self.context_cache = self._get_context_cache(context_cache)
+        move.apply(thermodynamic_state, sampler_state, context_cache=self.context_cache)
 
     def __getstate__(self):
         serialized_moves = [utils.serialize(move) for move, _ in self.move_set]
@@ -691,12 +696,13 @@ class BaseIntegratorMove(MCMCMove):
         # Create integrator.
         integrator = self._get_integrator(thermodynamic_state)
 
-        # Get context cache
-        local_context_cache = self._get_context_cache(context_cache)
+        # Get context cache and update attribute
+        self.context_cache = self._get_context_cache(context_cache)
 
         # Create context.
         timer.start("{}: Context request".format(move_name))
-        context, integrator = local_context_cache.get_context(thermodynamic_state, integrator)
+        # TODO: Is this still needed now that we are specifying the context?
+        context, integrator = self.context_cache.get_context(thermodynamic_state, integrator)
         timer.stop("{}: Context request".format(move_name))
         #logger.debug("{}: Context obtained, platform is {}".format(
         #    move_name, context.getPlatform().getName()))
@@ -884,11 +890,12 @@ class MetropolizedMove(MCMCMove):
         benchmark_id = 'Applying {}'.format(self.__class__.__name__ )
         timer.start(benchmark_id)
 
-        # Get context cache
-        local_context_cache = self._get_context_cache(context_cache)
+        # Get context cach and update attribute
+        self.context_cache = self._get_context_cache(context_cache)
 
+        # TODO: Is this still needed now that we are specifying the context?
         # Create context, any integrator works.
-        context, unused_integrator = local_context_cache.get_context(thermodynamic_state)
+        context, unused_integrator = self.context_cache.get_context(thermodynamic_state)
 
         # Compute initial energy. We don't need to set velocities to compute the potential.
         # TODO assume sampler_state.potential_energy is the correct potential if not None?
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -235,11 +235,16 @@ class MCMCSampler(object):
 
     """
 
-    def __init__(self, thermodynamic_state, sampler_state, move):
+    def __init__(self, thermodynamic_state, sampler_state, move,
+                 energy_context_cache=None, integrator_context_cache=None):
         # Make a deep copy of the state so that initial state is unchanged.
         self.thermodynamic_state = copy.deepcopy(thermodynamic_state)
         self.sampler_state = copy.deepcopy(sampler_state)
         self.move = move
+        # Create context caches for energy minimization and integrator
+        self._energy_context_cache = energy_context_cache
+        self._integrator_context_cache = integrator_context_cache
+        self._create_context_caches()
 
     def run(self, n_iterations=1):
         """
@@ -253,9 +258,8 @@ class MCMCSampler(object):
         """
         # Apply move for n_iterations.
         for iteration in range(n_iterations):
-            self.move.apply(self.thermodynamic_state, self.sampler_state, context_cache)
+            self.move.apply(self.thermodynamic_state, self.sampler_state, self._integrator_context_cache)
 
-    # TODO: This method is only used in test_mcmc.py, what do we need it for?
     def minimize(self, tolerance=1.0*unit.kilocalories_per_mole/unit.angstroms,
                  max_iterations=100, context_cache=None):
         """Minimize the current configuration.
@@ -295,6 +299,19 @@ class MCMCSampler(object):
         self.sampler_state.update_from_context(context)
 
         #timer.report_timing()
+
+    def _create_context_caches(self):
+        """
+        Handle the creation of context cache for energy minimization and move integrator.
+
+        Useful private method if we want to change the default behavior.
+        """
+        # Create independent unlimited context cache for energy minimization, if None
+        if self._energy_context_cache is None:
+            self._energy_context_cache = cache.ContextCache(capacity=None, time_to_live=None)
+        # Create independent unlimited context cache for move/integrator, if None
+        if self._integrator_context_cache is None:
+            self._integrator_context_cache = cache.ContextCache(capacity=None, time_to_live=None)
 
 
 # =============================================================================
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -205,6 +205,20 @@ class MCMCMove(SubhookedABCMeta):
         self._propagation_context_cache = context_cache
         return context_cache
 
+    def __deepcopy__(self, memo):
+        """Overriding default deep copy behavior to avoid creating new ContextCache objects without need."""
+        if memo is None:
+            memo = {}
+
+        cls = self.__class__
+        new_state = cls.__new__(cls)
+        memo[id(self)] = new_state
+        for k, v in self.__dict__.items():
+            if k != 'context_cache':
+                new_state.__dict__[k] = copy.deepcopy(v, memo)
+        new_state.__dict__['context_cache'] = self.context_cache
+        return new_state
+
 
 # =============================================================================
 # MARKOV CHAIN MONTE CARLO SAMPLER
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -175,7 +175,7 @@ class MCMCMove(SubhookedABCMeta):
         pass
 
     @staticmethod
-    def _get_context_cache(self, context_cache_input):
+    def _get_context_cache(context_cache_input):
         """
         Method to return context to be used for move propagation.
 
@@ -418,7 +418,6 @@ class SequenceMove(MCMCMove):
         sampler_state : openmmtools.states.SamplerState
            The sampler state to apply the move to.
         context_cache : opemmtools.cache.ContextCache, optional
-
 
         """
         for move in self.move_list:
@@ -692,9 +691,12 @@ class BaseIntegratorMove(MCMCMove):
         # Create integrator.
         integrator = self._get_integrator(thermodynamic_state)
 
+        # Get context cache
+        local_context_cache = self._get_context_cache(context_cache)
+
         # Create context.
         timer.start("{}: Context request".format(move_name))
-        context, integrator = context_cache.get_context(thermodynamic_state, integrator)
+        context, integrator = local_context_cache.get_context(thermodynamic_state, integrator)
         timer.stop("{}: Context request".format(move_name))
         #logger.debug("{}: Context obtained, platform is {}".format(
         #    move_name, context.getPlatform().getName()))
@@ -882,8 +884,11 @@ class MetropolizedMove(MCMCMove):
         benchmark_id = 'Applying {}'.format(self.__class__.__name__ )
         timer.start(benchmark_id)
 
+        # Get context cache
+        local_context_cache = self._get_context_cache(context_cache)
+
         # Create context, any integrator works.
-        context, unused_integrator = context_cache.get_context(thermodynamic_state)
+        context, unused_integrator = local_context_cache.get_context(thermodynamic_state)
 
         # Compute initial energy. We don't need to set velocities to compute the potential.
         # TODO assume sampler_state.potential_energy is the correct potential if not None?
@@ -1537,6 +1542,8 @@ class HMCMove(BaseIntegratorMove):
            The thermodynamic state to use when applying the MCMC move.
         sampler_state : openmmtools.states.SamplerState
            The sampler state to apply the move to. This is modified.
+        context_cache : openmmtools.cache.ContextCache
+            Context cache to be used during propagation with the integrator.
 
         """
         # Explicitly implemented just to have more specific docstring.
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -359,7 +359,8 @@ class SequenceMove(MCMCMove):
     False
 
     """
-    def __init__(self, move_list):
+    def __init__(self, move_list, **kwargs):
+        super(SequenceMove, self).__init__(**kwargs)
         self.move_list = list(move_list)
 
     @property
@@ -445,7 +446,8 @@ class WeightedMove(MCMCMove):
     False
 
     """
-    def __init__(self, move_set):
+    def __init__(self, move_set, **kwargs):
+        super(WeightedMove, self).__init__(**kwargs)
         self.move_set = move_set
 
     @property
@@ -630,7 +632,8 @@ class BaseIntegratorMove(MCMCMove):
 
     """
 
-    def __init__(self, n_steps, reassign_velocities=False, n_restart_attempts=4):
+    def __init__(self, n_steps, reassign_velocities=False, n_restart_attempts=4, **kwargs):
+        super(BaseIntegratorMove, self).__init__(**kwargs)
         self.n_steps = n_steps
         self.reassign_velocities = reassign_velocities
         self.n_restart_attempts = n_restart_attempts
@@ -816,7 +819,8 @@ class MetropolizedMove(MCMCMove):
     1
 
     """
-    def __init__(self, atom_subset=None):
+    def __init__(self, atom_subset=None, **kwargs):
+        super(MetropolizedMove, self).__init__(**kwargs)
         self.n_accepted = 0
         self.n_proposed = 0
         self.atom_subset = atom_subset
--- a/openmmtools/mcmc.py
+++ b/openmmtools/mcmc.py
@@ -254,6 +254,7 @@ class MCMCSampler(object):
         for iteration in range(n_iterations):
             self.move.apply(self.thermodynamic_state, self.sampler_state)
 
+    # TODO: This method is only used in test_mcmc.py, what do we need it for?
     def minimize(self, tolerance=1.0*unit.kilocalories_per_mole/unit.angstroms,
                  max_iterations=100, context_cache=None):
         """Minimize the current configuration.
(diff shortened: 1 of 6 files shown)


ijpulidos:
@jchodera According to our discussion on how to refactor this so we can come up with a cleaner use of the `ContextCache`. I realized that the way I understood this should be change breaks the API in at least two ways:

* Move objects won't have now a `context_cache` attribute, since this will be fed by the `MultiStateSampler`
* Calling the move method `.apply()` now requires a new argument for the `ContextCache` object. This one can be worked around by making it a keyword argument and specify a default behavior, but we would then have to do this in every move that implements the `apply` method.

Is it okay breaking the API in this way or maybe I missed something and we should be able to make these changes without breaking the API at all?

jchodera:
After our discussion, here is what we could do:
1. In order to support backwards-compatibility`MCMCMove` would receive an initializer that would accept an optional `context_cache` that would issue a warning if specified:
```python
def __init__(self, context_cache=None):
    if context_cache is not None:
        warning('Specifying context_cache on initialization of an MCMCMove is deprecated; please pass to apply() instead')
```
2. All `MCMCMove` direct subclasses would call `super().__init__(**kwargs)` and accept `**kwargs` in order to continue to allow `context_cache` as an optional argument, which would issue this warning. No base class initializers would store or directly accept `context_cache` anymore
3. All `MCMCMove` implementations of `apply()` would now accept optional keyword argument `context_cache=None`, and then use an `MCMCMove` base implementation of `_get_context_cache()` to return which context cache to use:
```python
context_cache = self._get_context_cache(context_cache)
```

ijpulidos:
With this approach I'm seeing a potential issue with the `mcmc_move.context_cache` attribute. We have to make sure we are syncing it and updating it correctly. Can we expect this to be done **exclusively** by invoking the `apply()` method?

As we have it in this moment we would need to sync this attribute in the `apply` method but we can get strange/undefined behavior when we just initialize the object without having run `apply` (the attribute is fixed as `None` in this case).

ijpulidos:
From our dev sync, we will be returning the **global** context cache for the attribute (maintaining backwards compatibility), raising a warning since we expect users to use the `context_cache`  parameter of the `apply()` method, instead.

ijpulidos:
This should be pretty close to be finished. Of course, the CI workflow test pass has the last word. Locally tests_mcmc.py are passing but we also need test_sampling to pass (that one takes a long time, so I'll just wait for CI to catch anything). Opening again for review.

ijpulidos:
@jchodera @mikemhenry  Now that the tests are happy with the changes, can you take a look at it? I think this is ready to go, but it resulted in a somewhat bigger change than expected, so reviews are very appreciated :).

ijpulidos:
I will be testing this further to see if the cycling issue is not shown (with the default behavior, #519 ) and that the context caches are shared when resuming simulations( #536 )

openmmtools/mcmc.py
```diff
@@ -146,9 +146,19 @@ class MCMCMove(SubhookedABCMeta):
     statistics such as number of attempted moves and acceptance rates.
 
     """
+    def __init__(self, context_cache=None):
+        if context_cache is not None:
+            logger.warning("Ignoring context_cache argument. Specifying context_cache on" 
+                           "initialization of an MCMCMove is deprecated; please pass to"
+                           " apply() method as kwarg instead.")
+        # hardcoding unused attribute to global context cache. For compatibility.
+        self.context_cache = cache.global_context_cache
```
jchodera, line 155: Is this only to support subclasses of `MCMCMove` that may reference `self.context_cache`? 
If so, I wonder if we want to make this a `@property` so we can also issue the warning when someone tries to access `self.context_cache`?
```python
@property
def context_cache(self):
    logger.warning('The MCMCMove.context_cache field has been deprecated. The API now requires context_cache be passed to apply(). Please update your code.')
    from openmmtools import cache
    return cache.global_context_cache
```

openmmtools/mcmc.py
```diff
@@ -146,9 +146,19 @@ class MCMCMove(SubhookedABCMeta):
...
+        # private attribute for propagation context cache - useful for testing
+        self._propagation_context_cache = None
```
jchodera, line 157: What is this used for? Do we need it?

openmmtools/mcmc.py
```diff
@@ -162,10 +172,55 @@ def apply(self, thermodynamic_state, sampler_state):
         sampler_state : openmmtools.states.SamplerState
            The initial sampler state before applying the move. This may
            be modified depending on the implementation.
+        context_cache : opemmtools.cache.ContextCache
```
jchodera, line 175: This is incomplete. It should be:
```
        context_cache_input : openmmtools.cache.ContextCache or None
            Context cache to be used in the propagation of the mcmc move. If None,
            it will create a new unlimited ContextCache object.
```

openmmtools/mcmc.py
```diff
@@ -162,10 +172,55 @@ def apply(self, thermodynamic_state, sampler_state):
...
+        """
+        Method to return context to be used for move propagation.
+
+        .. note:: centralized API point to deal with context cache behavior.
+
+        Parameters
+        ----------
+        context_cache_input : openmmtools.cache.ContextCache or None
+            Context cache to be used in the propagation of the mcmc move. If None,
+            it will create a new unlimited ContextCache object.
```
jchodera, line 190: If `None`, it should return the global ContextCache, not create a new one!

jchodera (commented):
@ijpulidos : I think there may be some confusion regarding `MCMCMove._get_context_cache()`. 

When we discussed this, we wanted to remove any methods or fields specifying `MCMCMove.context_cache` and localize it _only_ in the `MCMCMove.apply()` method. 

However, the `MultistateSampler` needs a `_get_context_caches(propagation_cache=None, energy_cache=None)` method that will either use the specified context cache or the global context cache.

I'm not sure why we need an `MCMCMove._get_context_cache()`. Perhaps we can have a brief chat to sort this out?

openmmtools/multistate/multistatesampler.py
```diff
@@ -1606,14 +1636,57 @@ def _is_completed_static(iteration_limit, iteration, last_err_free_energy,
...
+
             def __init__(self, error_message):
                 super().__init__(error_message)
                 # Critical messages which have halted a simulation, badly
                 logger.critical(error_message)
+
         raise RestorationError(message)
 
+    @staticmethod
+    def _create_context_caches(energy_context_cache_input, propagation_context_cache_input):
```
jchodera, line 1651: We can just call the arguments `energy_context_cache` and `propagation_context_cache`. We don't need the `_input` suffix.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1606,14 +1636,57 @@ def _is_completed_static(iteration_limit, iteration, last_err_free_energy,
...
+            raise ValueError("Energy context cache input is not a valid ContextCache or None type.")
+        # Handling propagation context cache
+        if propagation_context_cache_input is None:
+            # Default behavior, unlimited ContextCache
+            propagation_context_cache = cache.ContextCache(capacity=None, time_to_live=None)
+        elif isinstance(propagation_context_cache_input, cache.ContextCache):
+            propagation_context_cache = propagation_context_cache_input
+        else:
+            raise ValueError("MCMC move context cache input is not a valid ContextCache or None type.")
+        return energy_context_cache, propagation_context_cache
```
jchodera, line 1688: This method is internal, so it can just store these as `self.energy_context_cache` and `self.propagation_context_cache`. It doesn't need to return them.

openmmtools/multistate/sams.py
```diff
@@ -363,8 +363,10 @@ def _pre_write_create(self, thermodynamic_states: list, sampler_states: list, st
         # Update log weights
         self._update_log_weights()
 
-    def _restore_sampler_from_reporter(self, reporter):
-        super()._restore_sampler_from_reporter(reporter)
+    def _restore_sampler_from_reporter(self, reporter, energy_context_cache=None, propagation_context_cache=None):
```
jchodera, line 366: Should we just use `**kwargs` so any kwargs can be passed to the base class?

openmmtools/mcmc.py
```diff
@@ -162,10 +172,55 @@ def apply(self, thermodynamic_state, sampler_state):
         sampler_state : openmmtools.states.SamplerState
            The initial sampler state before applying the move. This may
            be modified depending on the implementation.
+        context_cache : opemmtools.cache.ContextCache
 
         """
         pass
 
+    def _get_context_cache(self, context_cache_input):
```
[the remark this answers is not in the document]
ijpulidos: From our discussion, we find this useful but we want to change the behavior to use a global context cache if None.

openmmtools/mcmc.py
ijpulidos: So, we needed this for the test [here](https://github.com/choderalab/openmmtools/blob/clean-context-cache-usage/openmmtools/tests/test_mcmc.py#L255-L265). I'm refactoring this test to use the number of accesses instead of having to use this private attribute that's only useful in testing. In the end this attribute won't be needed, thanks!

mikemhenry:
How does this PR interact with:
https://github.com/choderalab/openmmtools/pull/537

ijpulidos:
> How does this PR interact with: #537

@mikemhenry This PR should resolve the issue that #537 solves, so this one supersedes it. Thanks for noticing, I will update the description and close PRs accordingly, once this PR gets completed.

openmmtools/multistate/multistatesampler.py
ijpulidos: Here I was just thinking about explicitly setting the attributes (even if private/internal) in the `__init__`, because, as far as I know, it is commonly frowned upon to set them in another method, even if the method is called in the `__init__` I think it makes it a little cryptic. What do you think?

jchodera: That's fine---I agree your approach is less confusing.

openmmtools/mcmc.py
ijpulidos: Should be fixed now in latest version.

openmmtools/mcmc.py
ijpulidos: Now we don't need this attribute and testing is now using number of accesses.

openmmtools/mcmc.py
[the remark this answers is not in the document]
ijpulidos: It's now using the global context cache by default.

openmmtools/multistate/multistatesampler.py
ijpulidos: Fixed in latest changes.

openmmtools/multistate/sams.py
ijpulidos: Using arbitrary keyword arguments (kwargs) now. Do we want to do the same for non-keyword arguments(`reporter`) here? Seems like it might make it less readable :thinking:

openmmtools/mcmc.py
jchodera: Shouldn't we delete the attribute then?

ijpulidos:
I think this is expected, but just so that we are clear. Since we **were** serializing some information (metadata) about the context caches before these changes, we are now not being able to deserialize `.nc` files that were serialized with previous versions of our software. I wonder how much of an issue this is. We could try doing something with the deserialization but it feels like it is just doing some monkey-patching and it's probably not worth the effort. Related to #536 

The good news is that if I use the script/files attached to this comment, we now get shared context caches upon resuming/deserialization of simulations (see again #536). I'm editing the description of this issue to match that.

ijpulidos:
On other good news, the default behavior in this branch now avoids the cycling issue in #519 which is also related to https://github.com/choderalab/perses/issues/613 (where the main discussion is held). Also, it doesn't affect the performance when a context cache is specified. As far as I can see, then issue #519 is fixed (in terms of working around the cycling issue).

I am now waiting for the review on the new CI tests for the more up to date python versions (#542 ) which would also help the test passing on this PR.

jchodera:
> I think this is expected, but just so that we are clear. Since we were serializing some information (metadata) about the context caches before these changes, we are now not being able to deserialize .nc files that were serialized with previous versions of our software. I wonder how much of an issue this is. We could try doing something with the deserialization but it feels like it is just doing some monkey-patching and it's probably not worth the effort. Related to #536

@ijpulidos : I think we had discussed that this could be handled by writing our own `MCMCMove` base class deserialization routine which would just silently drop the cache-related entries from the `dict` and then call `__setstate__` for the parent object. 

You're right that this may not be worth the effort, but it might mean we should not break backward-compatibility in a bigfix patch release, and should instead call this 0.10.0.

openmmtools/mcmc.py
```diff
@@ -234,11 +290,16 @@ class MCMCSampler(object):
 
     """
 
-    def __init__(self, thermodynamic_state, sampler_state, move):
+    def __init__(self, thermodynamic_state, sampler_state, move,
+                 energy_context_cache=None, integrator_context_cache=None):
```
jchodera, line 294: The `MCMCSampler` isn't actually used in `MultiStateSampler`, and doesn't need separate `_energy_context_cache` or `_propagation_context_cache` attributes. 

In fact, we don't need to store any context caches at all---we can just accept an optional `context_cache` argument in `run()`, where we would default to `cache.global_context_cache` if none is specified.

openmmtools/mcmc.py
```diff
@@ -294,6 +355,19 @@ def minimize(self, tolerance=1.0*unit.kilocalories_per_mole/unit.angstroms,
 
         #timer.report_timing()
 
+    def _create_context_caches(self):
```
jchodera, line 358: We wouldn't need this method either, since we would just either use the provided `context_cache` in `run()` or fall back to the `cache.global_context_cache`.

jchodera (changes requested):
Thanks for the hard work on this PR, @ijpulidos!

For `MCMCSampler`: This is not actually used by `MultiStateSampler`---it's just an independent driver for sampling with `MCMCMove` that folks could use if they just want to simulate from a single thermodynamic state. As such, we don't need it to store any `ContextCache` objects internally---we can just ask the user to specify an optional `context_cache` when they call `run()` or fall back to the `cache.global_context_cache` otherwise.

For `MultiStateSampler`, I discovered an important issue: We don't have a way to ensure that the `ContextCache` we automatically create has the platform properties we need (e.g. `mixed` precision for CUDA or OpenCL; `DeterminsticForces` set to `true` for CUDA) if we need to create our own `ContextCache` objects, and there's nothing in `perses` that handles this either. 

Here's my suggested solution: Let's only accept `propagation_context_cache` and `energy_context_cache` arguments in `MultiStateSampler.run()`, and strip out all the other infrastructure that stores these as attributes. That will vastly simplify the infrastructure we need here.

`run()` would then fall back to `cache.global_context_cache` for any caches that are left as the default (`None`). This will cause periodic cycling if nothing else is done, but that's OK---we will create and configure these `ContextCache` arguments in perses explicitly and pass them in when we call `run()`.

We can also clean up some issues with `ContextCache` while we're at it:

We should fix [these lines](https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L678-L680) in `ContextCache` to make sure they clean up:
```
        try:
            openmm.Context(dummy_system, dummy_integrator, platform, platform_properties)
            return True
```
by adding an explicit `del context`:
```
        try:
            context = openmm.Context(dummy_system, dummy_integrator, platform, platform_properties)
            del context
            return True
```

I also note that the `global_context_cache` is [created without any platform properties specified](https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L780), so defaults to `single` precision on the fastest platform (e.g. CUDA if available). If we can figure out how to at least change this to `mixed` precision and set `DeterministicForces` to `true` by default, that would be ideal, but we may need to save that for a separate PR in the future and just write an issue to capture that future action item.

To create and configure the `ContextCache` objects in perses, we'll want to do something like
```python
# Attempt to get Platform
# NOTE: See openmmtools.utils.get_fastest_platform() if we just want the fastest platform that supports mixed precision
platform = openmm.Platform.getPlatformByName(platform_name)
# Configure platofrm
if platform.getName() in ['CUDA', 'OpenCL']:
    platform.setPropertyDefaultValue('Precision', 'mixed')
if platform.getName() == 'CUDA':
    platform.setPropertyDefaultValue('DeterministicForces', 'true')
# Create ContextCache for this platform
context_cache = ContextCache(platform=platform)
```

openmmtools/mcmc.py
```diff
@@ -162,10 +166,62 @@ def apply(self, thermodynamic_state, sampler_state):
...
+            # Default behavior, gobal Context Cache
+            context_cache = cache.global_context_cache
+        elif isinstance(context_cache_input, cache.ContextCache):
+            context_cache = context_cache_input
+        else:
+            raise ValueError("Context cache input is not a valid ContextCache or None type.")
+        return context_cache
+
+    def __deepcopy__(self, memo):
+        """Overriding default deep copy behavior to avoid creating new ContextCache objects without need."""
```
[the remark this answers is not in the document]
ijpulidos: You are right, now that the `context_cache` property defaults to the global context cache we shouldn't be needing this,

ijpulidos:
@jchodera respect the deserialization issue, you are right we did talk about this and I did make this work. I was just having a different deserialization issue (because I was using the wrong branch for the test, this one needed the RESTCapable branch of perses) and thought it was due to my changes. So, serialization/deserialization should be working just fine. Sorry for the noise.

mikemhenry:
@ijpulidos give me a ping on slack when this is ready for review

ijpulidos:
@jchodera Your suggestions sound reasonable, this will reduce the complexity of the infrastructure when dealing with `ContextCache` objects. 

I think we still need the `MultiStateSampler.from_storage` method to accept both `energy_context_cache` and `propagation_context_cache` (so it would be both `.run` and `.from_storage` that will accept these). And we have to make users specify them when they try to resume a simulation (or load one from serialized objects) IF they don't want to hit the cycling issue. Did I get that correctly?

openmmtools/mcmc.py
ijpulidos: As discussed we now have `energy_context_cache` and `sampler_context_cache` as attributes that are passed in `create()` and default to using the global context cache if `None`.

openmmtools/mcmc.py
ijpulidos: As discussed, this method is useful to have a centralized point where to handle the context caches behavior. This method is accesed from `create()` and `from_storage()`, so it reduces redundant code.

openmmtools/mcmc.py
ijpulidos: Ok, I'll make the changes, this one is still missing in the current status of the PR.

openmmtools/mcmc.py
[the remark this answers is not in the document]
ijpulidos: We still need it, because `MultiStateSampler` makes a deepcopy of the `mcmc_moves`, so we don't want each copy to have its own `context_cache` because it will end up eating memory without need (issue in #521 )

openmmtools/mcmc.py
jchodera: OK!

openmmtools/mcmc.py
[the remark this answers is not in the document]
jchodera: Can you elaborate a bit more?`MCMCMove` and derived classes no longer contain a `context_cache` attribute.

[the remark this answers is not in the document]
ijpulidos: Oh, you are correct! it's not a property and this one shouldn't be affected by the deepcopying. I agree, I don't think we need it.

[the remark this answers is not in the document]
ijpulidos: Changed in current state of the PR.

openmmtools/mcmc.py
ijpulidos: Changed in current state of the PR.

openmmtools/multistate/multistatesampler.py
```diff
@@ -950,6 +978,13 @@ def _read_options(check_iteration):
         self._last_mbar_f_k = last_mbar_f_k
         self._last_err_free_energy = last_err_free_energy
 
+        # Handle with context caches as specified
+        # update context caches attributes handling inputs
+        self.energy_context_cache, self.sampler_context_cache = self._initialize_context_caches(
```
jchodera, line 983: I think my only comment is that we discussed this could be simplified into a single `self._initialize_context_cache()` that just returns the global context cache, and there is no need to accept the kwargs for context caches since we can just set them as public attributes now, but I don't feel very strongly about this either way. I think it's just simpler and more streamlined if we don't have to pass these through various ways of creating or initializing the class, but it's not too critical either way.

jchodera (approved):
This is looking good now, and it looks like everything passes!
It would be good to do some quick testing on lilac to make sure we have fixed periodic cycling when initialized appropriately from perses, and have @mikemhenry give it a look over before merging and cutting the new openmmtools release!

openmmtools/multistate/multistatesampler.py
ijpulidos: I think I understand what you mean. For the sake of this release I think we are good with having this as it is now, but I'll raise a new issue where we can discuss enhancing this behavior.

ijpulidos:
Cycling issue is not shown when initialized properly. @mikemhenry can you please review, thanks!

openmmtools/mcmc.py
```diff
@@ -645,18 +686,16 @@ def apply(self, thermodynamic_state, sampler_state):
...
         # Create integrator.
         integrator = self._get_integrator(thermodynamic_state)
 
+        # Get context cache
+        local_context_cache = self._get_context_cache(context_cache)
+
         # Create context.
         timer.start("{}: Context request".format(move_name))
-        context, integrator = context_cache.get_context(thermodynamic_state, integrator)
+        # TODO: Is this still needed now that we are specifying the context?
```
mikemhenry, line 697: Is this a future TODO or did you figure it out?

openmmtools/multistate/multistatesampler.py
mikemhenry: Did you raise the issue? If you did can you link to John's comment?

mikemhenry (approved):
Just a few comments

openmmtools/mcmc.py
ijpulidos: It's still to be figured out, it just seems a bit redundant to specify a context cache and then use this `get_context`, but we have you to find out if there's a better way to accomplish this. Maybe for a future quick enhancement?

PR #544, opened 2022-02-03
Add support for receptor dimers

jfennick:
## Description

I'm performing a simulation using Yank of a system in which the receptor is a dimer. This causes the following error to be thrown from openmmtools https://github.com/choderalab/openmmtools/blob/master/openmmtools/forcefactories.py#L152

The call site in Yank https://github.com/choderalab/yank/blob/master/Yank/experiment.py#L2504 passes in ALL of the carbon alpha and/or carbon atoms, i.e. from both chains. However, in openmmtools, the chains are treated separately and rejected.

The original justification for not supporting more than one chain is that, at the time, it caused issues with the Barostat: https://github.com/openmm/openmm/issues/1854 

It appears that the issue was fixed by https://github.com/choderalab/openmmtools/issues/310 

It also appears that removing support for more than one chain was only a temporary workaround, and that the workaround was simply never removed. Issue #1854 was created on 7/5/2017 and issue #310 was resolved on 11/24/2017. In between, the workaround was added to version 0.13.1 on 9/20/2017 and the fix was added to version 0.13.4 on 11/27/2017. It appears that the justification for the workaround got lost between these two separate tickets, and someone added the fix while thinking about issue #310 and didn’t think to remove the workaround due to issue #1854.

The fact that this feature could cause problems with Yank was identified in https://github.com/choderalab/openmmtools/pull/290. (In that issue, the two molecules were incorrectly identified as the ligand and a receptor monomer, but the failure of the single molecule check still remains for a receptor dimer.)

In light of #310, I suggest the single molecule check be removed.

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/forcefactories.py
+++ b/openmmtools/forcefactories.py
@@ -113,10 +113,8 @@ def restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, sigma=3
         The thermodynamic state with the system. This will be modified.
     sampler_state : openmmtools.states.SamplerState
         The sampler state with the positions.
-    topology : mdtraj.Topology or openmm.Topology
-        The topology of the system.
-    atoms_dsl : str
-        The MDTraj DSL string for selecting the atoms to restrain.
+    restrained_atoms : List[int]
+        The indices of the atoms to be restrained.
     sigma : openmm.unit.Quantity, optional
         Controls the strength of the restrain. The smaller, the tighter
         (units of distance, default is 3.0*angstrom).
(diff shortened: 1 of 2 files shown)


mikemhenry:
Thanks @jfennick for this PR :100: 
I'll have @jchodera take a look at this :eyes:

mikemhenry:
Not sure why CI isn't running :upside_down_face:

jfennick:
bump?

jchodera:
There are still some potential issues with imaging of multimers (e.g. https://github.com/openmm/openmm/issues/2913) so I recommend we provide some simple way to add virtual (zero energy) bonds between multimers and bound cofactors at some point. (Perses does this now [here](https://github.com/choderalab/perses/blob/main/perses/annihilation/relative.py#L2059-L2063)). Perhaps we could create an issue to remind us to implement some general facility for ensuring all components of a complex are imaged together that we can reuse in perses, yank, and friends later?

PR #547, opened 2022-02-11
More streamlined context caches usage

### LINKED ISSUE #546: More streamlined `ContextCache` usage
I think my only comment is that we discussed this could be simplified into a single `self._initialize_context_cache()` that just returns the global context cache, and there is no need to accept the kwargs for context caches since we can just set them as public attributes now, but I don't feel very strongly about this either way. I think it's just simpler and more streamlined if we don't have to pass these through various ways of creating or initializing the class, but it's not too critical either way.

_Originally posted by @jchodera in https://github.com/choderalab/openmmtools/pull/538#discussion_r799199073_

ijpulidos:
## Description
Context caches for propagation and energy computations are expected to be handled only via `MultiStateSampler` instance attributes. Default behavior is to use global cache. If users want to avoid cycling they are expected to specify different context caches via the instance attributes.

Since we are not serializing context cache information, resuming simulations requires users to specify context cache attributes upon resuming.

Resolves #546 

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/tests/test_sampling.py
+++ b/openmmtools/tests/test_sampling.py
@@ -1460,14 +1460,6 @@ class TestMultiStateSampler(object):
             return
 
         with self.temporary_storage_path() as storage_path:
-            # For this test to work, positions should be the same but
-            # translated, so that minimized positions should satisfy
-            # the same condition.
-            original_diffs = [np.average(sampler_states[i].positions - sampler_states[i + 1].positions)
-                              for i in range(n_replicas - 1)]
-            assert not np.allclose(original_diffs,
-                                   [0 for _ in range(n_replicas - 1)]), "sampler %s failed" % self.SAMPLER
-
             # Create a replica exchange that propagates only 1 femtosecond
             # per iteration so that positions won't change much.
             move = mmtools.mcmc.IntegratorMove(openmm.VerletIntegrator(1.0 * unit.femtosecond), n_steps=1)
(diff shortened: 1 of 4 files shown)


ijpulidos:
Hmm, the test I designed for this is failing. It seems that using `context_cache._lru._n_access` is not the way to test this. Any ideas how to test this behavior? I'm just basically trying to test that propagating the replicas and the energy computation are using the context caches being specified.

openmmtools/tests/test_sampling.py
```diff
@@ -1442,6 +1443,58 @@ def test_online_analysis_stops(self):
...
+            self.call_sampler_create(sampler, reporter,
+                                     thermodynamic_states, sampler_states,
+                                     unsampled_states)
+            # Set context cache attributes
+            sampler.energy_context_cache = cache.ContextCache(capacity=None, time_to_live=None)
+            sampler.sampler_context_cache = cache.ContextCache(capacity=None, time_to_live=None)
+            # Compute energies
+            sampler._compute_energies()
+            # Check only energy context cache has been accessed
+            assert sampler.energy_context_cache._lru._n_access == sampler.n_replicas
```
jchodera, line 1485: I'd have the assertion failure print out a more informative error message, such as
```python
            assert sampler.energy_context_cache._lru._n_access == sampler.n_replicas, f'sampler.energy_context_cache._lru._n_access = {sampler.energy_context_cache._lru._n_access} (expected {sampler.n_replicas})'
```
but I think the proper way to test this is just to test for `> 0` (== we are using this cache) vs `==0` (== we are not using this cache). That may be all we need to change.
```

ijpulidos: I see, that sounds good! And yes, I already made the changes for more informative error. Will push them along with this new way of testing it. Thanks!

ijpulidos:
I found out that `ParallelTemperingSampler` had its own `_compute_replica_energies` method that was hardcoded to use the global context cache. Fixed that and now all tests are passing locally. I'll wait for GHA workflow to give me the green flag, if so, this should be ready for revision.

ijpulidos:
I guess the change to `main` for the main branch messed up the `codeclimate` check. I think this has to be changed from the `codeclimate` side of things, @mikemhenry @jchodera can you check that please? I could also just merge it without this check using admin capabilities, in case we want to do that.

jchodera:
Not sure I know how to fix that,---can we continue without it and have
@mikemhenry look into it in a separate PR?

PR #551, opened 2022-02-21
add test to check platform precision using a string

### LINKED ISSUE #550: platform_supports_precision broken when using a string
`platform_supports_precision` works fine if you pass in an openmm platform to check the precision, but breaks if you use a string:

```
  File "/home/mmh/miniconda3/envs/openmmtools/lib/python3.9/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/mmh/Projects/openmmtools/openmmtools/tests/test_utils.py", line 50, in test_string_platform_supports_precision
    assert platform_supports_precision("CPU", "mixed")
  File "/home/mmh/Projects/openmmtools/openmmtools/utils.py", line 537, in platform_supports_precision
    platform = openmm.Platform.getPlatformByName(platform)
UnboundLocalError: local variable 'openmm' referenced before assignment
```

mikemhenry:
## Description
Fixes #550 

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/tests/test_utils.py
+++ b/openmmtools/tests/test_utils.py
@@ -45,6 +45,11 @@ def test_platform_supports_precision():
             if supported_precisions != set(['mixed']):
                 raise Exception(f"'CPU' platform should support 'mixed' precision, but platform_supports_precision reports {supported_precisions}")
 
+
+def test_string_platform_supports_precision():
+    """Test that if we use a string for the platform name, it works"""
+    assert platform_supports_precision("CPU", "mixed")
+
 # =============================================================================
 # TEST STRING MATHEMATICAL EXPRESSION PARSING UTILITIES
 # =============================================================================


mikemhenry:
Okay this was a tricky one! 

We did import openmm at the top of `utils.py` but because we had an openmm import inside the function, it confused python:
https://github.com/choderalab/openmmtools/blob/main/openmmtools/utils.py#L547-L550

See this example
```
>>> import sys
>>> def foo():
...  print(sys)
...  if True:
...   import sys
... 
>>> foo()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in foo
UnboundLocalError: local variable 'sys' referenced before assignment
>>> 
```

jchodera:
Shouldn't we favor local imports (imports inside functions) so that we incur the overhead of importing OpenMM (which sometimes takes several seconds) only when we need it?

sroet:
> Shouldn't we favor local imports (imports inside functions) so that we incur the overhead of importing OpenMM (which sometimes takes several seconds) only when we need it?

(Just my opinion, feel free to ignore)
In general, yes. However, for [at least one of the classes defined in that file](https://github.com/choderalab/openmmtools/blob/7e6926f4719a331d06a66bc26be41d77779c3029/openmmtools/utils.py#L348) `unit` is required to be imported. So you can't really get around the module level `openmm` import for `openmm >= 7.6` (as that is done implicitly when calling `from openmm import unit`)

@mikemhenry, sorry for not catching this in #528 and thanks for adding a test.

jchodera (approved):
Thanks!

PR #557, opened 2022-03-09
Enhance stability of minimization

jfennick:
## Description
I have been experiencing some instabilities which, after much difficulty, I was able to track down to the minimizer. The observed behavior is that the minimizer will finish 'successfully', but some time later, during equilibration and/or propagating replicas, the simulation may or may not randomly crash with NaNs or even throw CUDA_ERROR_ILLEGAL_ADDRESS (700).

Nearly identical behavior can be found here https://github.com/openmm/openmm/issues/3414. In that ticket, the issue was that the box vectors were not being updated correctly. For stability, during minimization the box vectors should not be allowed to change at all. Subsequent equilibration at NPT should be used to adjust the box vectors.

In my case(s), the box vectors were changing only at the fourth significant figure, and yet that was sufficient to (usually) crash the simulations. Inserting Gradient Descent before FIRE seems to greatly ameliorate but not completely eliminate the instabilities. Replacing FIRE with L-BFGS (with or without Gradient Descent) also seems to fix the instabilities.

This PR adds three changes to improve the stability of minimization:

1. First and foremost, it temporarily disables the barostat during minimization to keep the box vectors fixed. This is probably the only change that is absolutely necessary to fix the instabilities.
2. It inserts a very short (24 timestep) Gradient Descent minimization before the main FIRE minimization. Gradient Descent is not asymptotically fast, but it is excellent for preconditioning a 'better' minimizer.
3. I have replaced FIRE with L-BFGS. Unlike FIRE, L-BFGS does not modify the box vectors (even when the pressure is not None) and it's a fine minimizer anyway.

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1395,33 +1395,20 @@ class MultiStateSampler(object):
         logger.debug('energy {:8.3f}kT box_volume {:6.4f}nm**3 bxx {:2.4f}nm  byy {:2.4f}nm  bzz {:2.4f}nm'.format(
             energy, volume, bxx, byy, bzz))
 
-        # Use the FIRE minimizer
-        integrator = FIREMinimizationIntegrator(tolerance=tolerance)
-
         # Get context and bound integrator from energy_context_cache
-        context, integrator = self.energy_context_cache.get_context(thermodynamic_state, integrator)
+        context, integrator = self.energy_context_cache.get_context(thermodynamic_state, integrator_grad)
         # inform of platform used in current context
         logger.debug(f"{type(integrator).__name__}: Minimize using {context.getPlatform().getName()} platform.")
 
         # Set initial positions and box vectors.
         sampler_state.apply_to_context(context)
 
-        # Minimize energy.
-        try:
-            if max_iterations == 0:
-                logger.debug('Using FIRE: tolerance {} minimizing to convergence'.format(tolerance))
-                while integrator.getGlobalVariableByName('converged') < 1:
-                    integrator.step(50)
-            else:
-                logger.debug('Using FIRE: tolerance {} max_iterations {}'.format(tolerance, max_iterations))
-                integrator.step(max_iterations)
-        except Exception as e:
-            if str(e) == 'Particle coordinate is nan':
-                logger.debug('NaN encountered in FIRE minimizer; falling back to L-BFGS after resetting positions')
-                sampler_state.apply_to_context(context)
-                openmm.LocalEnergyMinimizer.minimize(context, tolerance, max_iterations)
-            else:
-                raise e
+        # Use L-BFGS minimizer because (if pressure is not None) FIRE modifies
+        # the box vectors and can be unstable.
+        logger.debug('Using L-BFGS: tolerance {} max_iterations {}'.format(tolerance, max_iterations))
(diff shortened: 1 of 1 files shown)


jfennick:
See also https://github.com/choderalab/yank/pull/1267

mikemhenry:
@jfennick Thanks for this! I'll have @jchodera take a look!

jfennick:
FYI I just realized that I accidentally named this branch stable_equilibration instead of stable_minimization.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1309,45 +1309,73 @@ def _minimize_replica(self, replica_id, tolerance, max_iterations):
...
+        logger.debug('Sampler state {}/{}:'.format(replica_id + 1, self.n_replicas))
+
+        # Temporarily disable the barostat during minimization by setting the
+        # pressure to None. (Otherwise, the minimizer will modify the box
+        # vectors and may cause instabilities which are very difficult to debug.)
+        pressure = thermodynamic_state.pressure
+        thermodynamic_state.pressure = None
+
+        # Use Gradient Descent first for numerical stability
+        integrator_grad = GradientDescentMinimizationIntegrator()
```
jchodera, line 1323: The initial step size here is not guaranteed to be robust. I wonder if there's a way we can automatically select a more robust initial step size, or if this is not a critical issue.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1309,45 +1309,73 @@ def _minimize_replica(self, replica_id, tolerance, max_iterations):
...
+        context_grad = thermodynamic_state.create_context(integrator_grad)
```
jchodera, line 1324: It's good that we're creating the context and then cleaning it up again, since we don't repeatedly re-use the minimization context, but it would be useful if we also specified [platform properties to ensure we use mixed precision if needed](https://github.com/choderalab/openmmtools/blob/main/openmmtools/states.py#L1094) for the propagation or energy computation context caches.

jchodera (commented):
@ijpulidos : This is a great step forward, but there are a few risky things in here we should address before pulling it into a bugfix release. I'll look into this in more detail.

openmmtools/multistate/multistatesampler.py
jfennick: In some of my other work with gromacs, I use emstep 0.0001 (nanometers) which is 10 times smaller than the openmmtools default of 0.01 Angstroms. gromacs also has an adaptive step size, and I have some plots showing that starting with 0.001 Angstroms, the adaptive step size increases for the first ~25 iterations before stabilizing. In other words, you're only 'wasting' at most ~25 iterations, which isn't even a rounding error in the total runtime. Of course, those iterations are well spent if they can prevent minimization failures.

openmmtools/multistate/multistatesampler.py
jfennick: Done

jfennick:
I'm not sure why this is failing CI.
`pytest openmmtools/tests/test_sampling.py -k 'test_minimize'`
works on my machine.

ijpulidos:
@jfennick Thanks for your contributions. I'm tagging @jchodera for him to take another look into these, sine we need to see if the previously raised concerns are solved. I will also review this in detail, since this can potentially change many things in our calculations. For now I'm marking this to be released on our 0.22 milestone.

PR #561, opened 2022-03-18
Real time analysis information file

ijpulidos:
## Description
The purpose of this PR is to have a way to export a human readable file with real time analysis information.

CC https://github.com/choderalab/perses/issues/916

The main idea is taking the information straight from the reporter to avoid redundancy and synchronizing problems.

So far I've settled into using YAML for this file but I am still open for suggestions. With a yaml file being easy to append/update without having to read the whole file contents (as opposed to JSON). 

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -1313,6 +1313,9 @@ class MultiStateReporter(object):
         """
         Write real time YAML file with analysis data.
 
+        A real_time_analysis.yaml file will be generated in the same directory for the reporter netcdf file
+        (see :func:`~multistatereporter.MultiStateReporter` for more information).
+
         Overwrites file if it already exists.
 
         Parameters
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -137,6 +137,8 @@ class MultiStateReporter(object):
         self._analysis_particle_indices = tuple(analysis_particle_indices)
         if open_mode is not None:
             self.open(open_mode)
+        # Flag to check whether to overwrite real time statistics file
+        self._overwrite_statistics = True
 
     @property
     def filepath(self):
@@ -1321,13 +1323,23 @@ class MultiStateReporter(object):
         """
         Write real time YAML file with analysis data.
 
+        Overwrites file if it already exists.
+
         Parameters
         ----------
         data: dict
             Dictionary with the key, value pairs to store in YAML format.
         """
         reporter_dir, _ = os.path.split(self._storage_analysis_file_path)
-        with open(f"{reporter_dir}/real_time_analysis.yaml", "a") as out_file:
+        output_filepath = f"{reporter_dir}/real_time_analysis.yaml"
+        # Remove if it is a fresh reporter session
+        if self._overwrite_statistics:
+            try:
+                os.remove(output_filepath)
+            except FileNotFoundError:
+                pass
+            self._overwrite_statistics = False # Append from now on
+        with open(output_filepath, "a") as out_file:
             out_file.write(yaml.dump([data], sort_keys=False))
 
     # -------------------------------------------------------------------------
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -1317,6 +1317,19 @@ class MultiStateReporter(object):
         # All elements are one dimensional, return only the first element
         return {k: v.data[0] for k, v in nc_data.items()}
 
+    def write_real_time_yaml(self, data):
+        """
+        Write real time YAML file with analysis data.
+
+        Parameters
+        ----------
+        data: dict
+            Dictionary with the key, value pairs to store in YAML format.
+        """
+        reporter_dir, _ = os.path.split(self._storage_analysis_file_path)
+        with open(f"{reporter_dir}/real_time_analysis.yaml", "a") as out_file:
+            out_file.write(yaml.dump([data], sort_keys=False))
+
     # -------------------------------------------------------------------------
     # Internal-usage.
     # -------------------------------------------------------------------------
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -1307,6 +1307,16 @@ class MultiStateReporter(object):
         self.write_online_analysis_data(None, **kwargs)
         self.write_online_analysis_data(iteration, **kwargs)
 
+    def read_online_analysis_timing_info(self):
+        """Reads timing information from online_analysis group in reporter."""
+        # Timing keys in reporter
+        # TODO: Maybe a way not to have to manually specify the keys?
+        keys = ('iteration_time', 'time_per_iteration', 'estimated_time_remaining', 'estimated_total_time',
+                'estimated_finish_time')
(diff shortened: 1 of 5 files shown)


ijpulidos:
So far we have a basic/direct way of storing the information in the YAML file, I tried doing it from the reporter but I couldn't find an easy way to do it, so it's just a direct write to YAML file now. I wanted to go through the reporter machinery to avoid synchronization issues with it, I think that's still desired.

We have to double check if the variables we are storing are actually the ones we want, according to https://github.com/choderalab/perses/issues/916.

I have to fix the tests, there are some serialization/deserialization things that have to be changed in the tests to handle the new `_timing_info` attribute of the `MultiStateSampler`.

mikemhenry:
@ijpulidos Good timing! I've simulated this here https://github.com/choderalab/perses/pull/972/files#diff-2381ec1f5bdefb7f1bcd38a72d62f7e4d01cd05a04fde359f1c6d782058161c4R60-R72

mikemhenry:
But I agree, we will want to make sure we don't have to handel file locking and race conditions, so we should figure out a way to use a reporter

mikemhenry:
@ijpulidos can you post an example of what the real time analysis fie will look like? Then we can tag John in and see what he thinks. One thought I had is that we probably want to append to this file each iteration instead of overwriting it.

mikemhenry:
Oh and we should also have a section of the yaml document that has the units of each value so we can progamactially read what the units are instead of assuming them

jchodera:
> Oh and we should also have a section of the yaml document that has the units of each value so we can programmatically read what the units are instead of assuming them

Right now, the units are baked into the names of the fields. In a future update, it would be useful to use a scheme similar to [YANK](https://github.com/choderalab/yank-examples/blob/master/examples/binding/t4-lysozyme/all-ligands-explicit.yaml#L6-L7) or what OpenFF is moving toward so that we can read/write units programmatically.

ijpulidos:
@mikemhenry Sure, an example of it can be found [here](https://gist.github.com/ijpulidos/66923a3c7ed28fd34faf2d399a085659)

There are a couple  of issues with codeclimate that I might spend a bit of thinking to fix them, but this is good for review and for comments/suggestions. @jchodera @mikemhenry can you review it? Thanks!

openmmtools/multistate/multistatesampler.py
```diff
@@ -1711,6 +1725,35 @@ def _initialize_context_caches(self):
...
+        """
+        self._timing_data["iteration_seconds"] = iteration_time
+        self._timing_data["average_seconds_per_iteration"] = \
+            partial_total_time / (self._iteration - run_initial_iteration)
+        estimated_timedelta_remaining = datetime.timedelta(
+            seconds=self._timing_data["average_seconds_per_iteration"] * (iteration_limit - self._iteration)
+        )
+        estimated_finish_date = datetime.datetime.now() + estimated_timedelta_remaining
+        self._timing_data["estimated_time_remaining"] = str(estimated_timedelta_remaining)  # Putting it in dict as str
+        self._timing_data["estimated_iso_finish_date"] = estimated_finish_date.strftime("%Y-%b-%d-%H:%M:%S")
```
jchodera, line 1751: The time zone is not specified. This should be something like `estimated_localtime_finish_date` if you're using local time.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1711,6 +1725,35 @@ def _initialize_context_caches(self):
...
+        total_time_in_seconds = datetime.timedelta(
+            seconds=self._timing_data["average_seconds_per_iteration"] * iteration_limit
+        )
+        self._timing_data["estimated_total_iso_time"] = str(total_time_in_seconds)
```
jchodera, line 1755: I don't think you need `iso` in these tags.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1551,6 +1551,33 @@ def _offline_analysis(self):
...
+            performance = performance.value_in_unit(unit.nanosecond)
+        except KeyError:
+            performance = None
+        # Write real time offline analysis YAML file
+        # TODO: Specify units
+        data_dict = {"iteration": self._iteration,
+                     "percent_complete": self._iteration*100/self.number_of_iterations,
+                     "mbar_analysis": {"free_energy_in_kT": float(free_energy),
+                                       "standard_error_in_kT": float(self._last_err_free_energy),
+                                       "uncorrelated_samples": float(analysis._equilibration_data[-1])
```
jchodera, line 1575: I'd use `number_of_uncorrelated_samples`.

Is there anything else in `_equilibration_data` we could report too, like the initial production iteration or statistical inefficiency?

openmmtools/multistate/multistatesampler.py
```diff
@@ -1551,6 +1551,33 @@ def _offline_analysis(self):
...
+                                       },
+                     "timing_data": self._timing_data,
+                     "ns_per_day": performance
```
jchodera, line 1578: Shouldn't `ns_per_day` appear under the `timing_data` block?

openmmtools/tests/test_sampling.py
```diff
@@ -851,9 +851,12 @@ def test_from_storage(self):
 
         Checks that the static constructor MultiStateSampler.from_storage()
         restores the simulation object in the exact same state as the last
-        iteration.
+        iteration. Except from the reporter and timing data attributes, that
+        is _reporter and _timing_data, respectively.
 
         """
+        # We don't want to restore reporter and timing data attributes
+        __NON_RESTORABLE_ATTRIBUTES__ = ("_reporter", "_timing_data")
```
jchodera, line 859: This is an elegant solution! Thanks for adding this!

openmmtools/multistate/multistatereporter.py
```diff
@@ -1307,6 +1309,29 @@ def write_online_data_dynamic_and_static(self, iteration: int, **kwargs):
...
+
+        Overwrites file if it already exists.
+
+        Parameters
+        ----------
+        data: dict
+            Dictionary with the key, value pairs to store in YAML format.
+        """
+        reporter_dir, _ = os.path.split(self._storage_analysis_file_path)
+        output_filepath = f"{reporter_dir}/real_time_analysis.yaml"
```
jchodera, line 1324: Can you list the path of the YAML file that will be generated in the docstring as well?

jchodera (approved):
Looks great! Just have some minor requests for changes, but good to merge after that.

mikemhenry:
Don't worry about code climate @ijpulidos, didn't get a ton of input on this question https://openforcefieldgroup.slack.com/archives/CM9BA4AVA/p1647459523274309 but I'm in favor dropping things like code climate in favor for some CLI tool we can wire into CI like pyflakes

mikemhenry (approved):
This looks great! Maybe add a test that runs a few iterations then opens the document to make sure that the number of iterations match

mikemhenry:
@ijpulidos Oh and somewhere can you add an example yaml that this produces? Maybe somwhere here https://openmmtools.readthedocs.io/en/latest/devtutorial.html

ijpulidos:
The windows test is failing but it seems like it has nothing to do with the changes in this PR, as far as I can tell. I'm merging this as discussed.

PR #562, opened 2022-03-29
Returning and broadcasting _replica_thermodynamic_states

### LINKED ISSUE #449: MPI bug when multiple GPUs are used per calculation
I wanted to create an issue about this in openmmtools as well since we completed the transfering of the `multistate` code from YANK.

We're still experiencing mixing problem with the latest version of MPI when multiple GPUs are available (see choderalab/yank#1130 and #407). I've added a test to `test_sampling` checking this in #407, but I still haven't figure out the reason for the bug.

I'm sure this was working correctly in YANK during the SAMPLing challenge (i.e., YANK 0.20.1, right before adding the `multistate` module) so the next step would probably be trying binary search on the YANK versions after that to identify where the problem was introduced.

ijpulidos:
## Description
Fixes #449 by ensuring that `_mix_replicas` returns the `_replica_thermodynamic_states`. Also broadcasting it to the MPI context, according to [mpiplus broadcasting](https://github.com/choderalab/mpiplus/blob/216f1328d40f71e6554d909d470b6fe3104fde68/mpiplus/mpiplus.py#L226)

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -760,7 +760,7 @@ class MultiStateSampler(object):
             timer.start('Iteration')
 
             # Update thermodynamic states
-            self._mix_replicas()
+            self._replica_thermodynamic_states = self._mix_replicas()
 
             # Propagate replicas.
             self._propagate_replicas()
--- a/openmmtools/multistate/replicaexchange.py
+++ b/openmmtools/multistate/replicaexchange.py
@@ -290,6 +290,7 @@ class ReplicaExchangeSampler(multistate.MultiStateSampler):
             swap_fraction_accepted = float(n_swaps_accepted) / n_swaps_proposed
         logger.debug("Accepted {}/{} attempted swaps ({:.1f}%)".format(n_swaps_accepted, n_swaps_proposed,
                                                                        swap_fraction_accepted * 100.0))
+        return self._replica_thermodynamic_states
 
     @staticmethod
     @njit


jchodera (approved):
Indeed, this appears to have been the cause of the bug! Thanks for catching this.

At some point in the future, we may want to extend the `mpiplus` API to make `run()` more consistent---the inconsistency of
```python
self._do_action_1()
self._result = self._do_action_2()
```
bugs me a bit.

If we could later figure out how to extend the API of `mpiplus` so we can just have
```python
self._do_action_1()
self._do_action_2()
self._do_action_3()
```
that would be optimal. 

Alternatively, we could refactor to do 
```python
self._result1 = self._do_action_1()
self._result2 = self._do_action_2()
self._result3 = self._do_action_3()
```
but we would want to make sure the methods consistently always return new arguments without side-effects or cleanly update and distribute state no matter which solution we choose. Perhaps we can open an issue for that?

jchodera (changes requested):
Can you be sure to update the [revision history](https://github.com/choderalab/openmmtools/blob/main/docs/releasehistory.rst) and mark this as an important bugfix? 

We should cut a bugfix release after this.

jchodera:
Alternatively, we could rework [`self._propagate_replicas()`](https://github.com/choderalab/openmmtools/blob/637feeac38929da1cdb568f767dc6245851d3e4e/openmmtools/multistate/multistatesampler.py#L1274-L1324) to assume that the `self._replica_thermodynamic_states` is only correct on node 0 and to pass the appropriate state index along with each replica index, but I imagine this is what was happening originally and the non-obvious pattern here caused the propagation and mixing implementations to drift apart from each other.

mikemhenry:
Looking at our schedule CI, it looks like there may be an issue with the windows openmm builds

mikemhenry:
Now that we have the changelog notes added, do we want to get this merged and a bug fix cut? This would also get our real time anaylsis logging on conda-forge as well.

mikemhenry:
@ijpulidos is this ready to merge? It looks good RE what we talked about at our meeting!

mikemhenry (approved):

Issue #564, opened 2022-03-29
Dynamically create the name for real time analysis yaml file


ijpulidos:
Generate the name for the  real analysis yaml file from the reporter netcdf filenames, as it's done with checkpoints. To avoid collisions.

ijpulidos:
Solved by #565

PR #565, opened 2022-03-31
Real time analysis dynamic naming from reporter

ijpulidos:
## Description
Fixes naming convention for new real time analysis yaml file. It uses the reporter filename as a prefix, dynamically. Solves #564 

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/multistate/multistateanalyzer.py
+++ b/openmmtools/multistate/multistateanalyzer.py
@@ -1035,9 +1035,9 @@ class PhaseAnalyzer(ABC):
         Parameters
         ----------
         energy_matrix : array of numpy.float64, optional, default=None
-           Reduced potential energies of the replicas; if None, will be extracted from the ncfile
+           Reduced potential energies of the replicas.
         samples_per_state : array of ints, optional, default=None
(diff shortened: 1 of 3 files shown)


mikemhenry:
One quick question, do we want to do anything if the file exits already? I feel like there is some flag to overwrite files or not and I'm wondering if we should respect that?

ijpulidos:
> One quick question, do we want to do anything if the file exits already? I feel like there is some flag to overwrite files or not and I'm wondering if we should respect that?

Right now it's overwriting by default. I'm not entirely sure there's a flag for overwriting.I remember it always complaining about the `.nc` file existing and refusing to overwrite in that case, [here](https://github.com/choderalab/openmmtools/blob/51cad197abda420141c00c306bcd261206160a7d/openmmtools/multistate/multistatesampler.py#L578). For this new yaml file, the default behavior now is to overwrite, and this was agreed upon in a conversation with @jchodera

jchodera (approved):
Let's not forget to update the [changelog](https://github.com/choderalab/openmmtools/blob/main/docs/releasehistory.rst)!

Issue #567, opened 2022-04-08
Windows CI is currently failing


mjw99:
Currently seeing this error:

```
======================================================================
ERROR: Failure: ImportError (DLL load failed while importing _openmm: The specified module could not be found.)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\a\openmmtools\openmmtools\openmmtools\testsystems.py", line 57, in <module>
    import openmm
  File "C:\Miniconda\envs\test\lib\site-packages\openmm\__init__.py", line 19, in <module>
    from openmm.openmm import *
  File "C:\Miniconda\envs\test\lib\site-packages\openmm\openmm.py", line 13, in <module>
    from . import _openmm
ImportError: DLL load failed while importing _openmm: The specified module could not be found.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Miniconda\envs\test\lib\site-packages\nose\failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "C:\Miniconda\envs\test\lib\site-packages\nose\loader.py", line 416, in loadTestsFromName
    module = self.importer.importFromPath(
  File "C:\Miniconda\envs\test\lib\site-packages\nose\importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "C:\Miniconda\envs\test\lib\site-packages\nose\importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "C:\Miniconda\envs\test\lib\imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "C:\Miniconda\envs\test\lib\imp.py", line 171, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 702, in _load
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "D:\a\openmmtools\openmmtools\openmmtools\tests\test_alchemy.py", line 30, in <module>
    from openmmtools import testsystems, forces
  File "D:\a\openmmtools\openmmtools\openmmtools\__init__.py", line 8, in <module>
    from openmmtools import testsystems, integrators, alchemy, mcmc, states, cache, utils, constants, forces, forcefactories, storage, multistate
[...]

PR #568, opened 2022-04-08
Disable Windows CI

mjw99:
## Description
Disable the Windows CI since it is failing (see #567). Aim to fix this on a branch and re-enable later on.

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,13 +22,18 @@ jobs:
       matrix:
         cfg:
           - { os: macOS-latest,   python-version: '3.8', openmm: latest  }
+          - { os: windows-latest, python-version: '3.8', openmm: latest  }
           - { os: ubuntu-latest,  python-version: '3.8', openmm: latest  }
           - { os: ubuntu-latest,  python-version: '3.9', openmm: latest  }
           - { os: ubuntu-latest,  python-version: '3.8', openmm: nightly }
+      include:
+         os: windows-latest
+         continue-on-error: true
 
     env:
       OPENMM: ${{ matrix.cfg.openmm }}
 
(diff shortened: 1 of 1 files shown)


mikemhenry:
Good idea! Do you have a windows computer or VM to investigate why openmm might be having this issue? I'm wondering if we need to communicate upstream that there are issues with the windows build

mikemhenry:
Oh and I'm not sure if you have access to do this or not, but in the repo settings we need to remove the windows CI from the required checks.

mjw99:
Yes, I have a win10 image in my VirtualBox on my linux desktop; I can start to have a play there.

ijpulidos:
I remember trying this in a windows VM this a few days ago and as far as I can tell the build `py38hb8327ec_1` (for python 3.8) of `openmm` is having issues.

On the other hand, the `openmm 7.7.0 py39h6dc0912_1` (for py 3.9) doesn't show the issues. I think it doesn't hurt to communicate the issues with some builds upstream.

ijpulidos:
Wouldn't it be enough to mark the CI job as optional in this case? Since I think it helps the `openmm` folks that we are testing these configurations.

mjw99:
> Wouldn't it be enough to mark the CI job as optional in this case? Since I think it helps the openmm folks that we are testing these configurations.

I think that would be a good idea; I did not know that was possible.

ijpulidos:
Thanks for bringing this up! This had to be done changing the branch settings for the repository. I have changed this already. Feel free to close.

mjw99:
> Thanks for bringing this up! This had to be done changing the branch settings for the repository. I have changed this already. Feel free to close.

Awesome; thanks Iván. I think I was struggling with some of the CI.yml syntax.

PR #578, opened 2022-04-19
Missing the number of replicas in performance estimate

ijpulidos:
## Description
The number of replicas was a factor missing in the estimation of the raw performance for the real time analysis output.

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1756,12 +1756,14 @@ class MultiStateSampler(object):
         self._timing_data["estimated_total_time"] = str(total_time_in_seconds)
 
         # Estimate performance
-        # TODO: use units for timing information to easily convert between seconds and days
-        # there are some mcmc_moves that have no timestep attribute, catch exception
         moves_iterator = self._flatten_moves_iterator()
-        current_simulated_nanoseconds = sum([move.timestep.value_in_unit(unit.nanosecond) * move.n_steps for
-                                             move in moves_iterator if hasattr(move, "timestep") and hasattr(move, "n_steps")])
-        self._timing_data["ns_per_day"] = current_simulated_nanoseconds / (partial_total_time / 86400)
+        # Only consider "dynamic" moves (timestep and n_steps attributes)
+        moves_times = [move.timestep.value_in_unit(unit.nanosecond) * move.n_steps for move in moves_iterator if
+                       hasattr(move, "timestep") and hasattr(move, "n_steps")]
+        iteration_simulated_nanoseconds = sum(moves_times)
+        seconds_in_a_day = (1 * unit.day).value_in_unit(unit.seconds)
+        self._timing_data["ns_per_day"] = iteration_simulated_nanoseconds / (
+                    self._timing_data["average_seconds_per_iteration"] / seconds_in_a_day)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


openmmtools/multistate/multistatesampler.py
```diff
@@ -1759,7 +1759,8 @@ def _update_timing(self, iteration_time, partial_total_time, run_initial_iterati
         # TODO: use units for timing information to easily convert between seconds and days
         # there are some mcmc_moves that have no timestep attribute, catch exception
         try:
-            current_simulated_time = self.mcmc_moves[0].timestep * self._iteration * self.mcmc_moves[0].n_steps
+            current_simulated_time = self.mcmc_moves[0].timestep * self._iteration * self.mcmc_moves[0].n_steps * \
```
jchodera, line 1762: Shouldn't we actually be summing over _all_ `mcmc_moves`, in case there are multiple `MCMCMove` objects with dynamics involved, or appearing in different order?

And something still seems off here---do we want to use the estimate from `average_seconds_per_iteration` instead of the `partial_total_time` (which is the accumulated time since `run()` was started)?

openmmtools/multistate/multistatesampler.py
```diff
@@ -1774,6 +1772,15 @@ def _display_cuda_devices():
         cuda_devices_list = [entry.split(',') for entry in cuda_query_output.split('\n')]
         logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
 
+    def _flatten_moves_iterator(self):
```
mikemhenry, line 1775: Add a docstring (I know it is a private method but I it would be nice) describing what this method does.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1758,12 +1758,10 @@ def _update_timing(self, iteration_time, partial_total_time, run_initial_iterati
...
-        try:
-            current_simulated_time = self.mcmc_moves[0].timestep * self._iteration * self.mcmc_moves[0].n_steps
-        except AttributeError:
-            current_simulated_time = 0.0 * unit.nanosecond  # Hardcoding to 0 ns
-        performance = current_simulated_time.in_units_of(unit.nanosecond) / (partial_total_time / 86400)
-        self._timing_data["ns_per_day"] = performance.value_in_unit(unit.nanosecond)
+        moves_iterator = self._flatten_moves_iterator()
+        current_simulated_nanoseconds = sum([move.timestep.value_in_unit(unit.nanosecond) * move.n_steps for
+                                             move in moves_iterator if hasattr(move, "timestep") and hasattr(move, "n_steps")])
+        self._timing_data["ns_per_day"] = current_simulated_nanoseconds / (partial_total_time / 86400)
```
mikemhenry, line 1764: Might want to add something like `86400 (seconds in a day)` or something -- not everyone will know what that number means. Even better would be to use `unit` to do the conversion since we are using the line above it.

mikemhenry (changes requested):
Can you update the change log?

ijpulidos:
@mikemhenry Thanks for the comments, I've checked the numbers with the `perses` benchmarks and the estimates seem to be correct. Can you give this another look when you have the chance? Thanks!

mikemhenry (approved):
LGTM

Issue #560, opened 2022-03-17
Set up GPU CI


jchodera:
We should use the same perses infrastructure for nightly GPU CI.

PR #580, opened 2022-04-22
Add self hosted aws gpu testing

mikemhenry:
## Description
Resolves: #560 

I'm sure some secrets will be missing, which we can make org secrets instead of repo secrets so we can use this in other repos/ 

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/.github/workflows/self-hosted-gpu-test.yml
+++ b/.github/workflows/self-hosted-gpu-test.yml
@@ -1,0 +1,136 @@
+name: self-hosted-gpu-test
+on:
+  push:
+    branches:
+      - master
+      - main
+  workflow_dispatch:
+  schedule:
+    # nightly tests
+    - cron: "0 0 * * *"
+jobs:
+  start-runner:
+    name: Start self-hosted EC2 runner
+    runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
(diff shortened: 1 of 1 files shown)


mikemhenry:
@mjw99 I thought you would want to take a look at this one, please as any questions you have about how this works! I'd like to start using this as a template to adding GPU testing to other repos, so I'd like to add comments explaining what is going on in the file.

mikemhenry:
Also, in order to really test this and see what fails we will need to merge it, but I'd like to wait to do that until we get this file documented!

mjw99:
Hi Mike,

Generally looks good. 

I would pin the version of Ubuntu at 20.04 since 22.04 has just arrived; you probably do not want this to latently change whilst you're initially testing this. Also actions/checkout is now at version 3. Also, one super minor thing in devtools/conda-envs/test_env.yaml , it seems to have cython quoted twice.

mjw99 (changes requested):
Hi Mike,

Generally looks good.

I would pin the version of Ubuntu at 20.04 since 22.04 has just arrived; you probably do not want this to latently change whilst you're initially testing this. Also actions/checkout is now at version 3. Also, one super minor thing in devtools/conda-envs/test_env.yaml , it seems to have cython quoted twice

mikemhenry:
Good idea! Github is kinda slow when rolling these updates out, but I rather not worry about it https://github.com/actions/virtual-environments/issues/5490

mikemhenry:
Also I know this makes the diff messy, but I sorted the packages in each section so it is easier to catch dupes

mjw99 (approved):
LGTM!

mikemhenry:
Sweet! I expect something to fail post merge, I might have to add the PAT to this repo, but I think it's better to see what fails and fix it iteratively

mikemhenry:
Not sure what you mean, it looks like I will need to add `secrets.AWS_ACCESS_KEY_ID` `secrets.AWS_ACCESS_KEY_ID` `secrets.AWS_ACCESS_KEY_ID` and  `secrets.GH_PERSONAL_ACCESS_TOKEN` 

They are set correctly in the repo that is currently using this yaml file https://github.com/choderalab/perses/
I thought that maybe I had set them on the org level, but per-repo level is nice if we ever need/want to use different aws accounts for security reasons.

mikemhenry:
I think I know what you mean, no these are not set OK on master so it will fail until I add them, which I will do right now.

PR #585, opened 2022-05-19
Allow override t0 and g_t when computing free energy

zhang-ivy:
## Description
The `MultiStateAnalyzer` automatically determines t0 and g_t [here](https://github.com/choderalab/openmmtools/blob/0771ddc36f9a7d4cb183cafe3b95bbd1988fd048/openmmtools/multistate/multistateanalyzer.py#L1466). 

I have discovered that for my barnase:barstar repex simulations, t0 is always 1, which is too small. I'd like to be able to manually feed in t0 (and g_t) before retrieving the free energy.

This PR allows overriding t0 and g_t in the following way:
```python
# Given a reporter, create an analyzer
analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=max_n_iterations)

# Override t0 and g_t
t0 = 1000
subsample_rate = 10
decorrelated_u_ln, decorrelated_N_l = analyzer._compute_mbar_decorrelated_energies(t0=t0, subsample_rate=subsample_rate)

# Compute free energy
f_ij, df_ij = analyzer.get_free_energy()
```

## Todos
- [ ] Implement feature / fix bug
- [ ] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)
- [ ] Update [documentation](https://github.com/choderalab/openmmtools/tree/master/docs) as needed
- [ ] Update [changelog](https://github.com/choderalab/openmmtools/blob/master/docs/releasehistory.rst)Notable points that this PR has either accomplished or will accomplish.

## Status
- [ ] Ready to go

PR #586, opened 2022-05-25
Allow overriding of n_equilibration_iterations and statistical_inefficiency args for computation of free energy

zhang-ivy:
## Description
The `MultiStateSamplerAnalyzer` automatically determines n_equilibration_iterations and statistical inefficiency (g_t) [here](https://github.com/choderalab/openmmtools/blob/0771ddc36f9a7d4cb183cafe3b95bbd1988fd048/openmmtools/multistate/multistateanalyzer.py#L1466).

I have discovered that for my barnase:barstar repex simulations, number_equilibrated (aka n_equilibration_iterations) is always 1, which is too small. I'd like to be able to manually feed in n_equilibration_iterations and/or statistical inefficiency before retrieving the free energy.

This PR allows overriding n_equilibration_iterations and statistical_inefficiency in the following way:
```python
# Given a MultiStateReporter, create a MultiStateSamplerAnalyzer
analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=3000, n_equilibration_iterations=2001, statistical_inefficiency=10)

# Compute free energy
f_ij, df_ij = analyzer.get_free_energy()
```

Note that the changes in this PR only affect the determination of the decorrelated samples (which are inputed to MBAR to compute the free energy). It does not effect functions like `read_energies()` or `generate_mixing_statistics()`

Replaces https://github.com/choderalab/openmmtools/pull/585

## Todos

## Status
- [x] Ready to go


openmmtools/multistate/multistateanalyzer.py
```diff
@@ -2009,28 +2021,35 @@ def _get_equilibration_data(self, energies=None, neighborhoods=None, replica_sta
...
+                pass
+
+            # Discard equilibration samples.
+            # TODO: if we include u_n[0] (the energy right after minimization) in the equilibration detection,
+            # TODO:         then number_equilibrated is 0. Find a better way than just discarding first frame.
+            i_t, g_i, n_effective_i = multistate.utils.get_equilibration_data_per_sample(u_n[t0:])
+            n_effective_max = n_effective_i.max()
+            i_max = n_effective_i.argmax()
+            n_equilibration = self._n_equilibration_iterations if self._n_equilibration_iterations is not None else i_t[i_max] + t0 # if self._n_equilibration_iterations was not specified, account for initially discarded frames
+            g_t = self._statistical_inefficiency if self._statistical_inefficiency is not None else g_i[i_max]
```
ijpulidos, line 2052: I think we should use better names for the variables in this part, avoiding single letter names as possible. Just to improve readability when we have to come back to this code.

ijpulidos: I also see in this part that we use the specified values for `_statistical_inefficiency` and `_n_equilibration_iterations`, separately, but I wonder if specifying only one of them would need to affect the `n_effective_max` as in line 2028. I'm not really familiar with these calculations, but thought of bringing this up in case we need to find a more clever way to compute this values such that they use the specified parameters.

ijpulidos (commented):
Looks good, thanks! Just some comments that I believe should be addressed.

openmmtools/multistate/multistateanalyzer.py
zhang-ivy: All of the variable names here are the same as they were before, I didn't change them. `i` here represents index, so I could rename that to `index` if you think that's better. `g` represents the statistical inefficiency, and `t0` represents the number of equlibrations to discard due to equilibration. These variable choices are explained here: http://getyank.org/latest/algorithms.html#autocorrelate-algorithm

zhang-ivy: > I wonder if specifying only one of them would need to affect the n_effective_max as in line 2028. 

Specifying only `_statistical_inefficiency` means that we will need to use `_get_equilibration_data_per_sample()` to compute the  `n_effective_max` because its used to compute `n_equilibration`.

Meanwhile, specifying only `n_equilibration` means that we need to use `_get_equilibration_data_per_sample()` to compute the  `n_effective_max` because its used to compute `g_t` (i.e. the statistical inefficiency).

ijpulidos (approved):
Looks good, thanks!

PR #590, opened 2022-06-02
Release 0.21.4 - bugfix release

ijpulidos:
## Description
Preparing release 0.21.4.

Detailed information in the release history file.

## Status
- [ ] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,12 +8,12 @@ Bugfixes
 --------
 - Bug in statistical inefficiency computation -- where self.max_n_iterations wasn't being used -- was fixed (`#577 <https://github.com/choderalab/openmmtools/pull/577>`_).
 - Bug in estimated performance in realtime yaml file -- fixed by iterating through all MCMC moves (`#578 <https://github.com/choderalab/openmmtools/pull/578>`_)
-- Explicit update and broadcast of thermodynamic states when running equilibration. Issue `#579 <https://github.com/choderalab/openmmtools/issues/579>`_ (`#587 <https://github.com/choderalab/openmmtools/pull/587>`_).
+- Potential bug fixed by explicitly updating and broadcasting thermodynamic states in replicas, when used in an MPI (distributed) context. Issue `#579 <https://github.com/choderalab/openmmtools/issues/579>`_ (`#587 <https://github.com/choderalab/openmmtools/pull/587>`_).
 
 Enhancements
 ------------
-- Eliminate parmed dependency from ``DHFRExplicit``. Issue `#539 <https://github.com/choderalab/openmmtools/issues/539>`_ (`#588 <https://github.com/choderalab/openmmtools/pull/588>`_).
-- ``MultiStateSamplerAnalyzer`` now allows to manually specify equilibrium iterations and statistical inefficiency (`#586 <https://github.com/choderalab/openmmtools/pull/586>`_).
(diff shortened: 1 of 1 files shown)


mikemhenry:
Note for conda-forge release: `parmed` is no longer a dependency.

mikemhenry:
@jchodera once you get a chance to look at this I'll get it on conda-forge

docs/releasehistory.rst
```diff
@@ -8,6 +8,12 @@ Bugfixes
 --------
 - Bug in statistical inefficiency computation -- where self.max_n_iterations wasn't being used -- was fixed (`#577 <https://github.com/choderalab/openmmtools/pull/577>`_).
 - Bug in estimated performance in realtime yaml file -- fixed by iterating through all MCMC moves (`#578 <https://github.com/choderalab/openmmtools/pull/578>`_)
+- Explicit update and broadcast of thermodynamic states when running equilibration. Issue `#579 <https://github.com/choderalab/openmmtools/issues/579>`_ (`#587 <https://github.com/choderalab/openmmtools/pull/587>`_).
```
jchodera, line 11: This one-sentence description doesn't explain to the user what bug this would have caused, which is what all the other descriptions do. Can you be more explicit about what problematic behavior was corrected?

jchodera: For example, is the correct explanation "Bug in equilibration not doing anything when running with MPI -- fixed by explicitly broadcasting thermodynamics states after equilibration"

docs/releasehistory.rst
```diff
@@ -8,6 +8,12 @@ Bugfixes
...
+
+Enhancements
+------------
+- Eliminate parmed dependency from ``DHFRExplicit``. Issue `#539 <https://github.com/choderalab/openmmtools/issues/539>`_ (`#588 <https://github.com/choderalab/openmmtools/pull/588>`_).
```
jchodera, line 15: Can you explain why this is an enhancement or is desirable in this description? 
For example, "Eliminate `parmed` dependency---which was an optional dependency of `openmmtools`---from `DHFRExplicit`".

docs/releasehistory.rst
```diff
@@ -8,6 +8,12 @@ Bugfixes
...
+- ``MultiStateSamplerAnalyzer`` now allows to manually specify equilibrium iterations and statistical inefficiency (`#586 <https://github.com/choderalab/openmmtools/pull/586>`_).
```
jchodera, line 16: Can you mention what the new parameters are called?

jchodera (changes requested):
Apologies for requesting changes, but it's helpful to see if we can write really clear release notes! 

It seems like this process of crafting release notes only at release time makes it hard for the person writing the release notes to actually write _useful and informative_ release notes, since they didn't necessarily write the PRs. I wonder if we should change our practice and go back to having the person submitting the PR write clear and comprehensible release notes explaining what that PR changes. Presumably, the person who is contributing the PR should be able to succinctly explain what it does---we're having a hard time with this using the current approach.

docs/releasehistory.rst
ijpulidos: This doesn't really solve a bug, since the `equilibrate()` function is not really called inside an MPI Context. In my opinion this is just for "completness" to match the changes made to `_mix_replicas()` in https://github.com/choderalab/openmmtools/pull/562 (this one did fix the bug with mixing replicas in an MPI context).

ijpulidos:
@jchodera I updated the descriptions of the changes. Can you take another look? Thanks!

jchodera (approved):
Looks great, thanks!

PR #591, opened 2022-06-06
Add changelog entry spot to checklist

mikemhenry:
I think adding (and making sure we are happy with) the changelog entry as part of the template will:
1) Improve the quality of the change log entry, since that can be reviewed at commit time between the person who submitted the PR and the reviewers
1) Prevent annoying [changelog](https://github.com/choderalab/openmmtools/blob/master/docs/releasehistory.rst) that happen when PRs all try and edit the same line in the change log.

Then at release time (we should make a release PR template) the person doing the release can copy and paste everything into the changelog on github as well as the changelog we ship with the code.

Does this sound good @jchodera ?


ijpulidos (approved):
This works for me, we just have to be more careful when merging changes, making contributors and ourselves add the "changelog" entries to the PRs.

PR #593, opened 2022-06-07
Handling unsampled states in offline/realtime analysis.

### LINKED ISSUE #592: Don't use online analysis to set offline analysis matrix dim
We want to make sure the `_offline_analysis` works. `_online_analysis` is experimental and can be neglected for now. We need to make sure the matrix/array dimensions include the endstates for the offline analysis.

ijpulidos:
## Description
This aims to refactor `MultiStateSampler._offline_analysis` to handle unsampled states when doing the mbar estimate.

In this case the `_offline_analysis` method was decoupled from the `_online_analisis` one (which is still experimental). So that it initializes the first time with zeros as guess and subsequent results are updated  using the previous calculation.

For this I had to make a new private attribute `_last_mbar_f_k_offline` which gets also written to the nc file independently of the attributes used in the `_online_analysis` method.

Resolves #592 

## Todos
- [n/a] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)

## Status
- [ ] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -8,6 +8,7 @@ Bugfixes
 --------
 - Bug in statistical inefficiency computation -- where self.max_n_iterations wasn't being used -- was fixed (`#577 <https://github.com/choderalab/openmmtools/pull/577>`_).
 - Bug in estimated performance in realtime yaml file -- fixed by iterating through all MCMC moves (`#578 <https://github.com/choderalab/openmmtools/pull/578>`_)
+- Bug in handling unsampled states in realtime/offline analysis -- fixed by using ``MultiStateSampler._unsampled_states`` to build the mbar estimate array. Issue `#592 <https://github.com/choderalab/openmmtools/issues/592>`_ (`#593 <https://github.com/choderalab/openmmtools/pull/593>`_)
 
 
 0.21.3 - Bugfix release
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1514,9 +1514,14 @@ class MultiStateSampler(object):
         bump_error_counter = False
         # Set up analyzer
         # Unbias restraints is False because this will quickly accumulate large time to re-read the trajectories
-        # and unbias the restraints every online-analysis. Current model is to use the last_mbar_f_k as a
-        # hot-start to the analysis. Once we store unbias info as part of a CustomCVForce, we can revisit this choice.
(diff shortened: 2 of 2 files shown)


ijpulidos:
[This test](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/tests/test_sampling.py#L203), inside `TestHarmonicOscillatorsMultiStateSampler` should be executing the `_offline_analysis` method using unsampled states. Therefore these changes should be already covered in the tests.

mikemhenry (approved):
LGTM

zhang-ivy:
>[This test](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/tests/test_sampling.py#L203), inside TestHarmonicOscillatorsMultiStateSampler, should be executing the _offline_analysis method using unsampled states. Therefore these changes should be already covered in the tests.

@ijpulidos :  Was this test passing before? If so, we may want to discuss editing it or adding another version of it that fails without this PR

zhang-ivy:
Ah, I think I see why it was probably passing before -- its because the unsampled endstates are fed to the test _after_ the simulation is done: https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/tests/test_sampling.py#L171-L176. Therefore, the unsampled endstates are not being fed to the online/offline analysis functions..

ijpulidos:
@zhang-ivy Yeah, good catch!

I guess it is worth mentioning that failing to make the MBAR estimates in the `_offline_analysis` method has never killed the simulation, I think we want to leave like that so far, just because I have the feeling there are many ways this estimate could fail (not just with the unsampled states thing).

My guess is that since we are relying more and more in this real time analysis file, it is a good idea to write a test for it. I'll raise an issue about it, and maybe also discuss in the same issue if it's worth to stop/kill the simulation if the MBAR estimates are not correctly computed. Does that sound okay?

ijpulidos:
So this means running the `_offline_analysis` method could've been silently failing this whole time. Since the exception gets caught and it's never raised.

zhang-ivy:
>I guess it is worth mentioning that failing to make the MBAR estimates in the _offline_analysis method has never killed the simulation

Ah, I didn't realize this! 

>My guess is that since we are relying more and more in this real time analysis file, it is a good idea to write a test for it. I'll raise an issue about it, and maybe also discuss in the same issue if it's worth to stop/kill the simulation if the MBAR estimates are not correctly computed. Does that sound okay?

Yes sounds good, I think its ok for that to be a part of the next release, not openmmtools 0.21.4

PR #600, opened 2022-07-10
Fixes #599 : Change LangevinDynamicsMove to use openmm.LangevinMiddleIntegrator instead of openmm.LangevinIntegrator

### LINKED ISSUE #599: Switch LangevinDynamicsMove from openmm.LangevinIntegrator to openmm.LangevinMiddleIntegrator
We should switch `LangevinDynamicsMove` from `openmm.LangevinIntegrator` (based on OBABO) to `openmm.LangevinMiddleIntegrator` (based on BAOAB) to provide better configuration-space accuracy.

jchodera:
## Description
Fixes #599 : Changes `mcmc.LangevinDynamicsMove` to use `openmm.LangevinMiddleIntegrator` (a variant of BAOAB) instead of `openmm.LangevinIntegrator` (a variant of OBABO) to provide higher quality configuration-space samples.

## Todos
- [X] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)
- [X] Update [documentation](https://github.com/choderalab/openmmtools/tree/master/docs) as needed
- [X] Update [changelog](https://github.com/choderalab/openmmtools/blob/master/docs/releasehistory.rst)

## Status
- [ ] Ready to go


zhang-ivy:
Looks good to me! Thanks for opening this.

zhang-ivy:
@mikemhenry : 
>Is this a "breaking" change science wise? 

I'm not sure what you mean by "breaking" -- swapping out LangevinIntegator with LangevinMiddleIntegrator will not break anything, but the integrators do use different algorithms so fundamentally the we would not expect them to produce the exact same behavior.

>Does this have any implications for time series analysis tools like pymbar?

No, I don't think so, as the positions/velocities are not used for timeseries analyses, only the energies are.

mikemhenry:
What I meant when I asked about breaking was, is this change big enough to warrant a version bump? I know it doesn't break code, but changing the integrator feels like something that is a big change, like it would mean changing the text of a paper you were writing if you were using this tool (it would at least if you mentioned the integrator you used, which is pretty common in the methods section IMHO)

docs/releasehistory.rst
```diff
@@ -1,6 +1,15 @@
 Release History
 ***************
 
+0.21.5 - Bugfix release
```
mikemhenry, line 4:
```suggestion
0.22.0 - Bugfix release
```

I think we should make this bump-the-version-release since changing the integrator is kinda a big deal since it changes what someone would say the integrator they used in their method section was. I'm not sure if we should also just add a warning and keep the warning in for a release or two saying the integrator here has changed. 

I'm still going to approve the PR, but just thought we should have a discussion about how to make sure our users are prepared for this change.

mikemhenry (approved):

docs/releasehistory.rst
mikemhenry: @jchodera let me know what you think. I know people should pin versions and this code is pre 1.x so people are using it at their own peril.

ijpulidos (approved):
Looks good to me, thanks!

docs/releasehistory.rst
mikemhenry: Sounds good, do you think the release notes are enough to communicate the change? Or should we add a warning or something? Maybe it really isn't an issue at all since people should pin versions when generating data for papers.

PR #602, opened 2022-07-13
Returning velocities in sampler state when propagating replicas

### LINKED ISSUE #531: Augment checkpoint storage to save velocities
I am seeing dips in potential energy when I resume a repex simulation:
image (5)

I next decided to investigate the first dip in potential energy (at iteration 791). Here are histograms of replica energies for each iteration, colored by iteration number:
image (6)

We figured out that these dips are caused by resuming simulations from a checkpoint file. Since we are not saving velocities in the checkpoint, the velocities start at 0 when resuming a simulation. 

The short term solution for this is to set `reassign_velocities=True` in the `LangevinSplittingDynamicsMove()` when I instantiate it.
The long term solution is to save velocities in the checkpoint.

ijpulidos:
## Description
Velocities were not being passed when returning the sampler state when propagating replicas.

Resolves #531 

Note that the bug is only present when MPI (multiple GPUs) are being used. We need to test our software using MPI as well (issue #603 )

## Todos

## Status
- [x] Ready to go

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -1,6 +1,13 @@
 Release History
 ***************
 
+0.21.5 - Bugfix release
+=======================
+
+Bugfixes
+--------
+- Bug in returning velocities when propagating replicas. Fixed by using ``ignore_velocities=False`` in ``_propagate_replica``. Issue `#531 <https://github.com/choderalab/openmmtools/issues/531>`_ (PR `#602 <https://github.com/choderalab/openmmtools/pull/602>`_).
+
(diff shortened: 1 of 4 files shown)


ijpulidos:
I was also missing to change the same option when the replicas get updated inside the same loop. Should be ready to go now.

@zhang-ivy I think it makes sense if you can test this with your system before we merge this. Can you please do that? Thanks!

jfennick:
Oops, sorry about this. Looks like `ignore_velocities=False` slipped past me.

Issue #609, opened 2022-07-19
Adapt _get_equilibration_data_per_sample() to discard the first 1% of samples by default


zhang-ivy:
Change this line https://github.com/choderalab/openmmtools/blob/47f59f71aefe574b477a1a4949bd2ef653148c9e/openmmtools/multistate/utils.py#L180  to return `i_t[1:], g_i[1:], n_effective_i[1:]`

And make a note that the first 1% is being discarded due to equilibration because it throws off the t0 estimate otherwise
Note that the arrays returned will be of size `max_subset - 1`

ijpulidos:
I wonder if we want to add a new optional parameter to the function to allow specifying the percentage of the samples we would like to discard. That is, if we want to be able to discard more or less in different scenarios, but keeping 1% as the default value.

ijpulidos:
We want to make sure we use `ceiling` to round it up when we compute the number of samples.

zhang-ivy:
@jchodera : I'm realizing that this fix isn't as simple as I thought it would be. In https://github.com/choderalab/openmmtools/pull/586, I introduced options for the user to specify the `statistical_inefficiency` and `n_equilibration_iterations`, but I think are problems with how I implemented it. 

The main function that is affected by the aforementioned PR is 
[`_get_equilibration_data()`](https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistateanalyzer.py#L1998), which returns the following: 
```python
n_equilibration_iterations : int
    Number of equilibration iterations discarded
statistical_inefficiency : float
    Statistical inefficiency of production iterations
n_uncorrelated_iterations : float
    Effective number of uncorrelated iterations
``` 
Note that `n_uncorrelation_iterations` == `n_effective_max` below

There are multiple branches of logic in `_get_equilibration_data()` depending what the user has supplied to the `MultiStateSamplerAnalyzer`, let me walk through each of these branches: 

**Case 0:** Neither `n_equilibration_iterations` nor `statistical_inefficiency` are specified
- Compute the statistical inefficiency (using pymbar) for a bunch of t0s 
- `n_effective_max` -- `n_effective_samples` is computed from the pymbar-computed statistical inefficiencies and the maximum value is selected as `n_effective_max`
- `n_equilibration_iterations` is the t0 that maximizes `n_effective_samples` plus 1
-  `statistical_inefficiency` is the statistical inefficiency that corresponds to the selected `n_effective_max`

[...]

PR #610, opened 2022-07-27
Fix equilibration detection

zhang-ivy:
## Description

Changes: 
- Do not allow user to specify the `statistical_inefficiency` without `n_equilibration_iterations`
- When `n_equilibration_iterations` is specified and `statistical_inefficiency` is not, make sure that the final `n_equilibration_iterations` includes the t0 selected from the time origin analysis (aka `i_t[i_max]` [here](https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistateanalyzer.py#L2055))
- Always discard the first time origin in `_get_equilibration_data_per_sample()`
- Add API point for specifying `max_subset` when creating the a `MultiStateSamplerAnalyzer` 

For more context on why these changes are necessary, see https://github.com/choderalab/openmmtools/issues/609#issuecomment-1196051877

Fixes https://github.com/choderalab/openmmtools/issues/609

## Todos

## Status
- [x] Ready to go


zhang-ivy:
I've tested these changes locally by checking that: 
- An exception is raise when the user specifies the statistical_inefficiency without n_equilibration_iterations
- When n_equilibration_iterations is specified and statistical_inefficiency is not, the final n_equilibration_iterations includes the t0 selected from the time origin analysis (aka i_t[i_max] [here](https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistateanalyzer.py#L2055))
- The arrays returned by _get_equilibration_data_per_sample() are of length max_subset - 1
- The API point for specifying max_subset when creating the a MultiStateSamplerAnalyzer works and does change the max_subset in `_get_equilibration_data_per_sample()`

docs/releasehistory.rst
```diff
@@ -12,7 +12,8 @@ Changed behaviors
 Bugfixes
 --------
 - Bug in returning velocities when propagating replicas. Fixed by using ``ignore_velocities=False`` in ``_propagate_replica``. Issue `#531 <https://github.com/choderalab/openmmtools/issues/531>`_ (PR `#602 <https://github.com/choderalab/openmmtools/pull/602>`_).
-
+- Bug in equilibration detection when user specifies either ``statistical_inefficiency`` or ``n_equilibration_iterations`` (but not both). Fixed by preventing user from specifying only the former, and updating the behavior for the latter to include ``t0`` in the returned ``n_equilibration_iterations``. Issue `#609 <https://github.com/choderalab/openmmtools/issues/609>`_ (PR `#610 <https://github.com/choderalab/openmmtools/pull/610>`_). 
+- Bug in equilibration detection where ``get_equilibration_data_per_sample`` returns 0 for ``n_equilibration_iterations``. Fixed by always discarding the first time origin returned by ``get_equilibration_data_per_sample``. Issue `#609 <https://github.com/choderalab/openmmtools/issues/609>`_ (PR `#610 <https://github.com/choderalab/openmmtools/pull/610>`_).
```
mikemhenry, line 16: Could you add:
- [ ]  Note about new `max_subset` argument/how to use it?

mikemhenry (changes requested):

docs/releasehistory.rst
```diff
@@ -12,7 +12,8 @@ Changed behaviors
 Bugfixes
 --------
 - Bug in returning velocities when propagating replicas. Fixed by using ``ignore_velocities=False`` in ``_propagate_replica``. Issue `#531 <https://github.com/choderalab/openmmtools/issues/531>`_ (PR `#602 <https://github.com/choderalab/openmmtools/pull/602>`_).
-
+- Bug in equilibration detection when user specifies either ``statistical_inefficiency`` or ``n_equilibration_iterations`` (but not both). Fixed by preventing user from specifying only the former, and updating the behavior for the latter to include ``t0`` in the returned ``n_equilibration_iterations``. Issue `#609 <https://github.com/choderalab/openmmtools/issues/609>`_ (PR `#610 <https://github.com/choderalab/openmmtools/pull/610>`_).
```
jchodera, line 15: Instead of just "Bug in X", can we say what happened in that case, or just state that the behavior was undefined? For example, "If the user specified `statistical_inefficiency` but not `n_equilibration_detection`, this would cause problem X. This is now fixed by ..." These kinds of statements get us in the habit of clearly communicating with users (and ourselves!) what went wrong to help evaluate if folks need to re-run calculations.

openmmtools/multistate/utils.py
```diff
@@ -177,7 +181,7 @@ def get_equilibration_data_per_sample(timeseries_to_analyze, fast=True, max_subs
         except:
             g_i[i] = (time_size - t + 1)
         n_effective_i[i] = (time_size - t + 1) / g_i[i]
-    return i_t, g_i, n_effective_i
+    return i_t[1:], g_i[1:], n_effective_i[1:] # We should never choose data from the first time origin as the equilibrated data
```
jchodera, line 184: Could we add "because it contains snapshots warming up from minimization, which causes problems with correlation time detection"?

jchodera (approved):
This is great! Thanks!

docs/releasehistory.rst
zhang-ivy: Done

docs/releasehistory.rst
zhang-ivy: Done

openmmtools/multistate/utils.py
zhang-ivy: Done

zhang-ivy:
@mikemhenry @ijpulidos : I just pushed changes to address Mike and John's feedback. Can you re-review and merge?

ijpulidos (approved):
This looks good to me. We should try to come up with some sort of test that can help us validate the logic. I know some of these tests can be hard (it probably requires a trajectory if we want to test the `MultiStateAnalyzer`, for example) but maybe some others don't need a full trajectory. We had a conversation with @zhang-ivy on the possibility of sharing some of her notebooks and then, hopefully I can come up with some sort of test based on them.

jchodera:
> We should try to come up with some sort of test that can help us validate the logic.

I think we have some tests that _run_ the multistate sampler that run on the GPU. Perhaps we can just add an analysis step that checks to make sure that `t0` is not 0 or 1?

zhang-ivy:
Note that the gpu tests have been disabled: 

Screen Shot 2022-08-01 at 9 43 57 AM

zhang-ivy:
>I think we have some tests that run the multistate sampler that run on the GPU. Perhaps we can just add an analysis step that checks to make sure that t0 is not 0 or 1?

I just pushed a commit that adds an analysis step checking to make sure n_equilibration_iterations is > 1 in `TestHarmonicOscillatorsMultiStateSampler()`!

zhang-ivy:
I just pushed changes that also check:
- specifying only `statistical_inefficiency` (without `n_equilibration_iterations`) raises an exception
- specifying only `n_equilibration_iterations=10` (without `statistical_inefficiency`) returns `n_equilibration_iterations` > 10 (because the equilibration detection analysis will be run and added to the user specified value)
- specifying both `statistical_inefficiency` and `n_equilibration_iterations` returns the user specified values

zhang-ivy:
@mikemhenry @ijpulidos : Can you review again and merge this?

mikemhenry (approved):

mikemhenry:
Looking at it now!
RE GPU Tests: I've disable them until we figure out which ones we want to run, since right now we run all of them which is a waste of AWS money, but not if we think running all the tests are wroth it.

openmmtools/tests/test_sampling.py
```diff
@@ -161,9 +161,35 @@ def run(self, include_unsampled_states=False):
             logger.setLevel(logging.CRITICAL)
             simulation.run()
 
-            # Create Analyzer.
+            # Create Analyzer specfiying statistical_inefficiency without n_equilibration_iterations and 
+            # check that it throws an exception
+            assert_raises(Exception, self.ANALYZER, reporter, statistical_inefficiency=10)
```
ijpulidos, line 166: Not blocking. Just to have a better readability of what the test is doing here I suggest the following change 
```suggestion
            with assert_raises(Exception):
                self.ANALYZER(reporter, statistical_inefficiency=10)
```

ijpulidos (approved):
This looks good to me, thanks for going through this. Just a minor non-blocking suggestion.

openmmtools/tests/test_sampling.py
ijpulidos: Marked as solved, since it was merged during the re-review.

PR #611, opened 2022-07-27
`LangevinDynamicsMove` now accepts `constraint_tolerance` parameter

### LINKED ISSUE #608: Allow integrator tolerance to be specified for mcmc.LangevinDynamicsMove
Currently, [`mcmc.LangevinDynamicsMove`](https://github.com/choderalab/openmmtools/blob/main/openmmtools/mcmc.py#L1018-L1156) does not allow the integrator tolerance to be specified. We should add this as an optional argument to the initializer and enable it to be a public attribute that the user can change later.

ijpulidos:
## Description
`LangevinDynamicsMove` now accepts a `constraint_tolerance` parameter (defaults to the same value as the previously used `LangevinIntegrator`).

Resolves #608 

## Todos

## Status
- [ ] Ready to go

## Changelog
```
``LangevinDynamicsMove`` now allows ``constraint_tolerance`` parameter and public attribute, for specifying the fraction of the constrained distance within which constraints are maintained for the integrator (Refer to `Openmm's documentation <http://docs.openmm.org/latest/api-python/generated/openmm.openmm.LangevinMiddleIntegrator.html#openmm.openmm.LangevinMiddleIntegrator.setConstraintTolerance>_ for more information).
```


openmmtools/mcmc.py
```diff
@@ -1143,17 +1150,21 @@ def __getstate__(self):
         serialization = super(LangevinDynamicsMove, self).__getstate__()
         serialization['timestep'] = self.timestep
         serialization['collision_rate'] = self.collision_rate
+        serialization['constraint_tolerance'] = self.constraint_tolerance
```
mikemhenry, line 1153: Do we need to add a default key here so that way someone can deserialize an older file without getting a key error here?

mikemhenry (changes requested):

mikemhenry:
Also maybe we should add to the changelog entry a link to the openmm doc for that setting.

zhang-ivy:
Looks great, thanks!! And I agree with Mike's suggestions.

zhang-ivy (approved):

openmmtools/mcmc.py
ijpulidos: Yes, good catch! Thanks.

mikemhenry:
Serialization will be important if you want to load an `nc` simulated with an older version of openmmtools but want to look at it after this version comes out, see https://github.com/choderalab/openmmtools/issues/612

zhang-ivy:
Yes I definitely agree it would be useful, but I just meant I’ll leave it up to you guys to figure out what you want to do

mikemhenry (changes requested):
Accidently dismised my review.

The only blocker on this is that we need to handle the case where `constraint_tolerance` isn't in the nc file.

ijpulidos:
> The only blocker on this is that we need to handle the case where `constraint_tolerance` isn't in the nc file.

@mikemhenry I just checked and it isn't a problem, to my surprise. I was kind of expecting for it to be a problem, so it took me a while to realize why I think it isn't. As far as I can see, the flow of the object when deserializing is something like the following.

1. The sampler gets created from the `nc` files with the [`from_storage` method](https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistatesampler.py#L264).
2. This method then creates the instance of the `sampler` with the sampler's [`_instantiate_sampler_from_reporter` method](https://github.com/choderalab/openmmtools/blob/7034e41eea1030185adffdf5da81f5899b2ad8e3/openmmtools/multistate/multistatesampler.py#L287)
3. The previous step generates/deserialize instances of the `mcmc_moves` using the reporter's [`read_mcmc_moves` method](https://github.com/choderalab/openmmtools/blob/7034e41eea1030185adffdf5da81f5899b2ad8e3/openmmtools/multistate/multistatesampler.py#L949)
4. Lastly, after calling the utility deserializing function, `deserialize`, this function just creates new instances of the `mcmc_moves` objects by calling their respective `__init__` via the `__new__` method [here](https://github.com/choderalab/openmmtools/blob/7034e41eea1030185adffdf5da81f5899b2ad8e3/openmmtools/utils.py#L676), since we are not overloading `__new__` in the MCMCmove objects. Resulting in using the attributes that are serialized, falling back to defaults for those who were not serialized.

I hope that's understandable. Let me know if I am making myself clear here.

The good news is that in #613 we are also indirectly testing this, since we are using legacy/older versions of serialized systems in the test. But I don't know if we want a specific test for this checking the match between serialized attributes vs. the ones that are falling back to default when deserializing an older simulation. Generally speaking, we don't really have much tests for resuming simulations, as far as I can tell.

mikemhenry:
> The good news is that in https://github.com/choderalab/openmmtools/pull/613 we are also indirectly testing this, since we are using legacy/older versions of serialized systems in the test.

This is great since it will check for regressions, I want to then get 613 merged in first so we can have that check for this PR.

zhang-ivy:
@mikemhenry @ijpulidos : I merged #613 into `main` and then merged `main` into this branch and then double checked that `TestSerializedMultiStateSampler` (introduced in #613) is passing successfully in the latest GHA workflows. Are you ok with me merging this now?

mikemhenry:
Sounds good, we can have John review the release PR

zhang-ivy:
@mikemhenry Can you "approve" this PR? merging is blocked because you requested changes

mikemhenry (approved):

PR #613, opened 2022-07-28
Restore velocities only if found in serialized nc file

### LINKED ISSUE #612: Load serialized files generated with openmmtools<=0.21.2
When trying to load simulation results from netcdf files generated with `openmmtools<=0.21.2` we encounter that the velocities are not present, since they were added in [0.21.3](https://github.com/choderalab/openmmtools/releases/tag/0.21.3). This results in an error when deserializing as follows
```
openmmtools.multistate.multistatesampler.MultiStateSampler._throw_restoration_error.<locals>.RestorationError: Attempting to restore from any checkpoint failed. Either your data is fully corrupted or something has gone very wrong to see this message. Please open an issue on the GitHub issue tracker if you see this!
```
We should be able to read older serialized `.nc` files with up to date versions of `openmmtools`. We could just try catching the `KeyError` exception when attempting to restore the velocities [here](https://github.com/choderalab/openmmtools/blob/cec8c3d6627d50e638daf79c759ec670a4fddff3/openmmtools/multistate/multistatereporter.py#L1727).

ijpulidos:
## Description
Restores velocities only if variable is found in the serialized files. This allows back compatibility with serialized files using `openmmtools<0.21.3`, since velocities were added in that release.

Resolves #612 

## Todos

## Status
- [x] Ready to go

## Changelog
```
Bug when restoring serialized files generated by ``openmmtools<0.21.3``. Fixed by catching the ``KeyError`` exception when deserializing.
```


ijpulidos:
I had to add `.nc` data files using an older version of `openmmtools` (0.21.2) to test that we can resume simulations from them correctly.

I thought about emulating this behavior by removing the `velocities` variables from the files themselves, but `NetCDF` doesn't have an API for removing variables.

mikemhenry (changes requested):
Could you add a note that this fix enables reading of older nc files? Other than that, LGTM

ijpulidos:
Please note that I had to restructure the tests a bit, you can see the same number of tests are getting run (just in case you are wondering). But I had to make a new base class for avoiding running redundant tests in child classes. 

The core reason for this is that the serialized objects in the data `.nc` files that the new test is using are meant to be used only with the `MultiStateSampler` object and not all the others that inherit from it in the tests, due to missing variables/fields in the files.

@mikemhenry @jchodera This should be good to go now, can you please review it? Thanks.

mikemhenry (approved):
LGTM

PR #616, opened 2022-08-02
Building path using `os` library instead of f-strings for real time analysis

### LINKED ISSUE #612: Load serialized files generated with openmmtools<=0.21.2

### LINKED ISSUE #615: Realtime analysis output failing when output has no subdirectory
In #561 the realtime analysis file was introduced but the path for it did not account for simulations where the results are to be generated in the same current owrking directory, resulting in the wrong path and a `PermissionError` like the following, when trying to write the yaml file.
```python
PermissionError: [Errno 13] Permission denied: '/alanine_dipeptide_legacy_real_time_analysis.yaml'
```
This can be solved by using `os.path.join` instead of f-strings when dealing with paths.

ijpulidos:
## Description

Resolves #615 by using the `os.path.join` method instead of relying on manually introduced f-strings.

## Todos

## Status
- [ ] Ready to go

## Changelog 
```
Bug in generating real time analysis output when reporter is in the same working directory. Fixed by using ``os.path.join`` for creating the output path.
```

### DIFF
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -1330,7 +1330,7 @@ class MultiStateReporter(object):
         reporter_dir, reporter_filename = os.path.split(self._storage_analysis_file_path)
         # remove extension from filename
         yaml_prefix = os.path.splitext(reporter_filename)[0]
-        output_filepath = f"{reporter_dir}/{yaml_prefix}_real_time_analysis.yaml"
+        output_filepath = os.path.join(reporter_dir, f"{yaml_prefix}_real_time_analysis.yaml")
         # Remove if it is a fresh reporter session
         if self._overwrite_statistics:
             try:
--- a/openmmtools/tests/test_sampling.py
+++ b/openmmtools/tests/test_sampling.py
@@ -1577,8 +1577,8 @@ class TestMultiStateSampler(object):
             storage_dir, reporter_filename = os.path.split(sampler._reporter._storage_analysis_file_path)
             # remove extension from filename
             yaml_prefix = os.path.splitext(reporter_filename)[0]
-            output_filepath = f"{storage_dir}/{yaml_prefix}_real_time_analysis.yaml"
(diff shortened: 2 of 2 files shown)


ijpulidos:
It might seem trivial, but does it make sense to create a test for this? It can be something that checks running some short simulation and outputs the real time analysis file in the current working directory.

zhang-ivy:
Previously, I was having trouble when feeding a file with no directory to the reporter, e.g. `analysis.yaml`. I had to change this to be `./analysis.yaml`, but I think we should make sure that specifying only a file works as well. Can we make sure this PR allows this?

ijpulidos:
@zhang-ivy Yes, I guess you mean specifying `output.nc` or similar when instantiating the reporter. And yes, this is exactly the problem that this solves, when you don't specify a directory for the reporter its directory attribute becomes an empty string and that is fine with `os.path.join`. That is, `os.path.join('', 'output.nc')` returns `output.nc` as expected.

zhang-ivy:
Great!

mikemhenry (approved):
LGTM!
For the refactor, we should try and use `pathlib` for everything

PR #619, opened 2022-08-03
add python 3.10 to CI

mikemhenry:
## Description

## Todos

## Status
- [ ] Ready to go

## Changelog message
```

```


ijpulidos:
According to https://github.com/choderalab/openmmtools/runs/7661564216?check_suite_focus=true#step:7:3003 we shouldn't import `Iterable` directly from collections but we should be using `collections.abc.Iterable` instead (https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable). This should fix the tests erroring with python 3.10.

ijpulidos (approved):
Looks good to me, thanks!

PR #634, opened 2022-11-25
remove del operator on PhaseAnalyzer

### LINKED ISSUE #633: PhaseAnalyzer del causing error.
The delete operator on PhaseAnalyzer seems to be tripping up.  PR to follow

typical traceback below
```
OSError: /tmp/tmpyggicnli/c8c410b8-040b-480a-b0c8-8ff0ca0207f6/complex_rbfe.nc does not exist
DEBUG:mpiplus.mpiplus:Single node: executing <bound method MultiStateReporter.close of <openmmtools.multistate.multistatereporter.MultiStateReporter object at 0x7fbd785fcac0>>
Exception ignored in: <function PhaseAnalyzer.__del__ at 0x7fbdc6454f70>
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 581, in __del__
    self.clear()
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 1227, in clear
    super().clear()
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 597, in clear
    self._invalidate_cache_values('reporter')
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 780, in _invalidate_cache_values
    self._invalidate_cache_values(k)
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 780, in _invalidate_cache_values
    self._invalidate_cache_values(k)
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 780, in _invalidate_cache_values
    self._invalidate_cache_values(k)
  [Previous line repeated 1 more time]
  File "/home/richard/miniconda3/envs/openfe/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 777, in _invalidate_cache_values
    self._computed_observables[observable] = None
AttributeError: 'MultiStateSamplerAnalyzer' object has no attribute '_computed_observables'
```

richardjgowers:
fixes #633

the `__del__` doesn't do anything except delete attributes, so isn't necessary

### DIFF
--- a/openmmtools/multistate/multistateanalyzer.py
+++ b/openmmtools/multistate/multistateanalyzer.py
@@ -576,12 +576,6 @@ class PhaseAnalyzer(ABC):
         self._use_online_data = use_online_data
         self._read_online_data_if_present()
 
-    def __del__(self):
-        # Explicitly close storage
-        self.clear()
-        if self._reporter is not None:
-            del self._reporter
-
     def clear(self):
         """Reset all cached objects.
 


mikemhenry:
I haven't looked at this bit of code lately, maybe @ijpulidos could check, but `__del__` does call `self.clear()`, which does some cache clearing stuff

richardjgowers:
I did a quick scan, there's no explicit `.close()` or similar calls, it's just removing references from the `_cache` dict, which is what Python garbage collection should also be doing.  I didn't have time to go through the git history to see if there used to be a file handle held....

ijpulidos:
@richardjgowers thanks for reporting the issue and contributing the possible solution.

I tend to agree with your remarks, the Python garbage collector should already be doing this. Nevertheless, I would like to have a reproducible example/snippet where the issue appears, it doesn't have to be a simple one, just something I can use to reproduce the issue and go from there. If you have one I can use that would be very helpful. Thanks!

mikemhenry:
We should probably catch the `AttributeError` in `_invalidate_cache_values` since (from https://stackoverflow.com/a/1481512) mentions

>However, there are valid use cases for __del__: e.g. if an object X references Y and also keeps a copy of Y reference in a global cache (cache['X -> Y'] = Y) then it would be polite for X.__del__ to also delete the cache entry.

Which looks like what we are doing here.

Also from the [docs](https://docs.python.org/3/reference/datamodel.html?highlight=__del__#object.__del__) it sounds like `__del__` " [is] called when the instance is about to be destroyed" so we really just be using context mangers if cleanup is important with an `__exit__` (or `finally` block) since it might not get called until the intrepter exits.

richardjgowers:
Aha ok.  For context, I think I was getting this error when I was trying to create a `PhazeAnalyzer` and it was failing in `__init__` which is why you end up trying to delete an attribute that should probably always exist.

PR #644, opened 2023-01-09
Update docstring default for alchemical_pme_treatment

IAlibay:
## Description
Tiny one, just realised the docstring didn't reflect the init arguments.

Didn't think it was changelog worthy so I omitted adding anything there, can do so if wanted.

## Status
- [ ] Ready to go


IAlibay:
Probably worth asking whilst I'm here, I'm currently digging into charge handling for AbsoluteAlchemicalFactory, is the following still correct in the context of https://github.com/choderalab/openmmtools/pull/380?

https://github.com/choderalab/openmmtools/blob/2e64980d86cf087f69e282927eebf2b41b73a3ea/openmmtools/alchemy.py#L524-L529

mikemhenry (approved):
Looks good! @jchodera will have to answer your question

lhmzhou:
13 checks seem to stay in a lock state. Anyway to re-trigger from your end?

PR #658, opened 2023-03-10
Report caught exceptions as warnings in base integrator move

### LINKED ISSUE #643: BaseIntegratorMove does not distinguish between reasons for Exceptions
Currently, [`BaseIntegratorMove`](https://github.com/choderalab/openmmtools/blob/2e64980d86cf087f69e282927eebf2b41b73a3ea/openmmtools/mcmc.py#L601), which is the base class of classes like [`LangevinDynamicsMove`](https://github.com/choderalab/openmmtools/blob/2e64980d86cf087f69e282927eebf2b41b73a3ea/openmmtools/mcmc.py#L1018), [does not distinguish between causes for `Exceptions` during dynamics](https://github.com/choderalab/openmmtools/blob/2e64980d86cf087f69e282927eebf2b41b73a3ea/openmmtools/mcmc.py#L718). Meant to guard against NaNs, this behavior means that even important failures that should be relayed to the user, like
```
openmm.OpenMMException: The periodic box size has decreased to less than twice the nonbonded cutoff.
```
instead trigger restarts rather than useful warnings. We should instead catch behavior like this that is important to communicate to users because it can frustrate debugging.

ijpulidos:
## Description
This is a simple change that reports the exception error message caught when trying to restart a simulation. It reports it as a warning.

Resolves #643 

In general I think we want to rework the logic of how the simulation is getting restarted when it is encountering NaNs. But this probably needs to be in sync with catching special Exceptions generated by OpenMM and could be a critical change that would probably have to be taken care of in more detail, outside of the scope of this PR.

Finding a system that fails specifically the way we want it and avoids being caught in any of the other code paths is not trivial. hence testing specifically for this does not seem trivial.

## Todos

## Status
- [ ] Ready to go

## Changelog message
```
Report error messages as warnings when NaNs are encountered while running the simulation, instead of silently restarting simulation.
```


openmmtools/mcmc.py
```diff
@@ -715,8 +717,10 @@ def apply(self, thermodynamic_state, sampler_state, context_cache=None):
                 # Run dynamics.
                 timer.start("{}: step({})".format(move_name, self.n_steps))
                 integrator.step(self.n_steps)
-            except Exception:
+            except Exception as e:
                 # Catches particle positions becoming nan during integration.
+                # Return the exception message as a warning
+                warnings.warn(str(e))
```
mikemhenry, line 723: This should do it, but we probably want to also add a message about how we will continue running

mikemhenry (approved):
LGTM

openmmtools/mcmc.py
ijpulidos: Not sure what you mean by this exactly, the user does get an error message saying that it will be restarted, as per https://github.com/choderalab/openmmtools/pull/658/files#diff-85baf1df1041cffd52aca6ac6b2d4a7f00244716d1a0d14ac54af193c6835072R740-R742

Would this do it? I think we do want to refactor this restart logic to make it more readable and maintainable in the future.

ijpulidos: I'll be merging this for now, but feel free to raise an issue about this if needed.

PR #630, opened 2022-11-21
Make compatible with pymbar 3 and 4

mattwthompson:
## Description
This PR attempts to introduce compatibility with pymbar 4 in a way that does not break compatibility with pymbar 3. This is heavily inspired by https://github.com/shirtsgroup/physical_validation/pull/220

Resolves #629 

## Todos
- [ ] Implement feature / fix bug
- [ ] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)
- [ ] Update [documentation](https://github.com/choderalab/openmmtools/tree/master/docs) as needed
- [ ] Update [changelog](https://github.com/choderalab/openmmtools/blob/master/docs/releasehistory.rst) to summarize changes in behavior, enhancements, and bugfixes implemented in this PR

## Status
- [ ] Ready to go

## Changelog message
```

```

PR #659, opened 2023-03-10
Pymbar 3 4

mikemhenry:
## Description

## Todos

## Status
- [ ] Ready to go

## Changelog message
```

```

### DIFF
--- a/openmmtools/multistate/multistatereporter.py
+++ b/openmmtools/multistate/multistatereporter.py
@@ -1810,7 +1810,7 @@ class MultiStateReporter(object):
                 dimension_name = 'scalar'
             # Create variable.
             nc_variable = storage_nc.createVariable(path, variable_type,
-                                                    dimension_name, zlib=True)
+                                                    dimension_name, zlib=False)
 
         # Assign the value to the variable.
         if fixed_dimension:
--- a/openmmtools/multistate/utils.py
+++ b/openmmtools/multistate/utils.py
@@ -24,11 +24,12 @@ LICENSE
 This code is licensed under the latest available version of the MIT License.
 
 """
+from typing import Dict
 import logging
 import warnings
 import numpy as np
 
-from pymbar import timeseries  # for statistical inefficiency analysis
+from openmmtools.multistate.pymbar import subsample_correlated_data
 
 logger = logging.getLogger(__name__)
 
@@ -287,6 +288,6 @@ def subsample_data_along_axis(data, subsample_rate, axis):
     cast_data = np.asarray(data)
     data_shape = cast_data.shape
     # Since we already have g, we can just pass any appropriate shape to the subsample function
-    indices = timeseries.subsampleCorrelatedData(np.zeros(data_shape[axis]), g=subsample_rate)
+    indices = subsample_correlated_data(np.zeros(data_shape[axis]), g=subsample_rate)
     subsampled_data = np.take(cast_data, indices, axis=axis)
     return subsampled_data
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,3 +9,8 @@ Provide a brief description of the PR's purpose here.
 
 ## Status
 - [ ] Ready to go
+
+## Changelog message
+```
+
+```
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,21 +21,20 @@ jobs:
       fail-fast: false
       matrix:
         cfg:
-          - { os: macOS-latest,   python-version: '3.8',    openmm: latest,  pymbar: '3' }
-          - { os: windows-latest, python-version: '3.8.10', openmm: latest,  pymbar: '3' }
-          - { os: ubuntu-latest,  python-version: '3.8',    openmm: latest,  pymbar: '3' }
-          - { os: ubuntu-latest,  python-version: '3.9',    openmm: latest,  pymbar: '3' }
-          - { os: ubuntu-latest,  python-version: '3.8',    openmm: nightly, pymbar: '3' }
-          - { os: ubuntu-latest,  python-version: '3.10',   openmm: latest,  pymbar: '3' }
-          - { os: ubuntu-latest,  python-version: '3.8',    openmm: latest,  pymbar: '4' }
-          - { os: ubuntu-latest,  python-version: '3.10',   openmm: rc,      pymbar: '4'  }
+          - { os: macOS-latest,   python-version: '3.8',    openmm: latest,  pymbar: 3 }
+          - { os: windows-latest, python-version: '3.8.10', openmm: latest,  pymbar: 3 }
+          - { os: ubuntu-latest,  python-version: '3.8',    openmm: latest,  pymbar: 3 }
+          - { os: ubuntu-latest,  python-version: '3.9',    openmm: latest,  pymbar: 3 }
+          - { os: ubuntu-latest,  python-version: '3.8',    openmm: nightly, pymbar: 3 }
+          - { os: ubuntu-latest,  python-version: '3.10',   openmm: latest,  pymbar: 3 }
+          - { os: ubuntu-latest,  python-version: '3.8',    openmm: latest,  pymbar: 4 }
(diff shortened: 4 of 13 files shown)


mikemhenry:
Supersedes https://github.com/choderalab/openmmtools/pull/630

mikemhenry:
hmmmm, seeing the same errors for pymbar 3&4

```

ERROR: Test multistate sampler on a harmonic oscillator without unsampled endstates
  File "/home/runner/micromamba-root/envs/openmmtools-test/lib/python3.8/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/runner/work/openmmtools/openmmtools/openmmtools/tests/test_sampling.py", line 234, in test_without_unsampled_states
    self.run(include_unsampled_states=False)
  File "/home/runner/work/openmmtools/openmmtools/openmmtools/tests/test_sampling.py", line 223, in run
    raise Exception("Dimensionless free energy difference exceeds MAX_SIGMA of %.1f" % MAX_SIGMA)
Exception: Dimensionless free energy difference exceeds MAX_SIGMA of 6.0
```

mikemhenry:
That test passes fine on `main` but on this branch it fails ~10% of the time...

mikemhenry:
I yelled when I saw this: https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/utils.py#L178-L183

I could not figure out how an import fixed things, well it "solved" the problem that this bare except hid. 

BUT what is crazier is that we don't get a failure every time, but will essentially calculate a subsample rate larger than than our input array, which causes our equilibrated indices to be `indices=[0]` which sometimes passes our tests if we are lucky... 

Anyway, bare excepts were a bad idea. I can't belive the amount of time I spent on this.

mikemhenry:
I will be cleaning up those exceptions in this PR as well....

ijpulidos:
Oh yeah, of course bare excepts are biting us, good catch! Thanks.

openmmtools/tests/test_cache.py
```diff
@@ -218,10 +218,12 @@ def test_copy_integrator_state(self):
...
             try:
                 return getattr(integrator, attribute_name)
-            except:
+            except Exception as err:
+                raise err
                 try:
                     return getattr(integrator, 'get' + attribute_name)()
-                except:
+                except Exception as err:
+                    raise err
```
ijpulidos, line 226: I think we just want to check for `AttributeError` here, in both cases.

mikemhenry: Yep! I wanted to see what sorts of error(s) they could be and I agree!

openmmtools/multistate/multistatereporter.py
```diff
@@ -368,7 +368,7 @@ def _open_dataset_robustly(self, *args, n_attempts=5, sleep_time=2,
         for attempt in range(n_attempts-1):
             try:
                 return netcdf.Dataset(*args, **kwargs)
-            except:
+            except: # This should be safe since we will raise the error below on return
```
ijpulidos, line 371: I think we still want to output the error/exception message here in the debug log, since it could be potentially different than the one being generated on return. Does that make sense?

mikemhenry: Sure! We can log it

mikemhenry:
On [windows-latest, py-3.8.10, OpenMM-8.0, pymbar-4](https://github.com/choderalab/openmmtools/actions/runs/4424185291/jobs/7757714130#logs)
```
ERROR: Test expected number of entries in real time analysis output yaml file.
  File "C:\Users\runneradmin\micromamba-root\envs\openmmtools-test\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "D:\a\openmmtools\openmmtools\openmmtools\tests\test_sampling.py", line 1577, in test_real_time_analysis_yaml
    sampler.run()
  File "D:\a\openmmtools\openmmtools\openmmtools\multistate\multistatesampler.py", line 791, in run
    self._update_timing(iteration_time, partial_total_time, run_initial_iteration, iteration_limit)
  File "D:\a\openmmtools\openmmtools\openmmtools\multistate\multistatesampler.py", line 1773, in _update_timing
    self._timing_data["ns_per_day"] = iteration_simulated_nanoseconds / (
ZeroDivisionError: float division by zero
```

and on [windows-latest, py-3.10, OpenMM-8.0, pymbar-3](https://github.com/choderalab/openmmtools/actions/runs/4424185291/jobs/7757713525#logs)

```
ERROR: Test that creating a MultiState by storage string and reporter is the same
  File "C:\Users\runneradmin\micromamba-root\envs\openmmtools-test\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "D:\a\openmmtools\openmmtools\openmmtools\tests\test_sampling.py", line 1436, in test_storage_reporter_and_string
    sampler.run()
  File "D:\a\openmmtools\openmmtools\openmmtools\multistate\multistatesampler.py", line 791, in run
    self._update_timing(iteration_time, partial_total_time, run_initial_iteration, iteration_limit)
  File "D:\a\openmmtools\openmmtools\openmmtools\multistate\multistatesampler.py", line 1773, in _update_timing
    self._timing_data["ns_per_day"] = iteration_simulated_nanoseconds / (
ZeroDivisionError: float division by zero
```

All the other windows builds are fine, but we probably do want to catch that error so it doesn't crash a simulation if it happens. That can be done in another PR.

mikemhenry:
Okay! Since everything is passing except those two errors, I am now going to switch up the CI matrix so that we are testing mostly with pymbar 4, but will keep a few jobs at pymbar 3 to make sure we maintain support for both.

mikemhenry:
Once we finalize the matrix, I will mark them as required

openmmtools/multistate/multistatesampler.py
```diff
@@ -1014,7 +1013,8 @@ def _read_options(check_iteration):
                                                     "Either your data is fully corrupted or something has gone very "
                                                     "wrong to see this message. "
                                                     "Please open an issue on the GitHub issue tracker if you see this!")
-            except:
+            except Exception as err:
+                raise err
```
ijpulidos, line 1017: We probably want to change this to either not raise the exception and catch whatever exception it can raise, or catch a specific exception. This doesn't seem to affect the tests but maybe we do affect applications/users by raising the exception here.

openmmtools/multistate/utils.py
```diff
@@ -177,12 +178,14 @@ def get_equilibration_data_per_sample(timeseries_to_analyze, fast=True, max_subs
     i_t = np.floor(counter * time_size / max_subset).astype(int)
     for i, t in enumerate(i_t):
         try:
-            g_i[i] = timeseries.statisticalInefficiency(series[t:], fast=fast)
-        except:
+            g_i[i] = statistical_inefficiency(series[t:], fast=fast)
+        except Exception as e:
+            print("If you see this, open an issue https://github.com/choderalab/openmmtools/issues")
+            raise e
```
ijpulidos, line 184: Do we want to raise the exception here? I feel like this might be disruptive for current users/applications.

ijpulidos (changes requested):
This is looking great. Just a few comments I think we need to adress.

openmmtools/multistate/multistatesampler.py
mikemhenry: Since I didn't understand what kind of error this could throw, I was thinking we would need to catch this one in the wild and I rather do that than not know. We could run perses on this PR and see if it causes a probme.

openmmtools/multistate/utils.py
mikemhenry: See https://github.com/choderalab/openmmtools/pull/659#discussion_r1138117826 I think we can see if perses has any problems as well as other downstream packages as a double check. I worry it does more damage leaving it in the codebase. I could be convinced to just log an error but I think no one will see it.

openmmtools/multistate/multistatesampler.py
ijpulidos: I see, okay. I guess that should work.

ijpulidos (approved):
Looks great, thanks! I tested with perses locally in a tyk2 benchmark system and it seems to be working (using pymbar3). LGTM!

PR #662, opened 2023-03-13
Avoid expecting a fixed index/order in forces -- Fix windows tests

ijpulidos:
## Description
These changes avoid shadowing the variable names from the outer scope and hopefully will be fixing the failing CI workflow for windows with Openmm8, for example in https://github.com/choderalab/openmmtools/actions/runs/4409361691/jobs/7725555141#step:7:3200

## Todos

## Status
- [ ] Ready to go

### DIFF
--- a/openmmtools/tests/test_integrators.py
+++ b/openmmtools/tests/test_integrators.py
@@ -520,7 +520,13 @@ class TestExternalPerturbationLangevinIntegrator(TestCase):
             omm_force.setParticleParameters(2, charge=0.417, sigma=0, epsilon=1)
             omm_force.updateParametersInContext(omm_context)
 
-        force = wbox.system.getForce(2)  # Non-bonded force.
+        # Get Nonbonded force from system -- Note: order may vary by platform. Don't expect a specific index.
+        for force in wbox.system.getForces():
+            if isinstance(force, openmm.NonbondedForce):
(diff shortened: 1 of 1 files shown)


openmmtools/tests/test_integrators.py
```diff
@@ -508,19 +508,25 @@ def test_ncmc_update_parameters_in_context(self):
...
+
+        def switchon(omm_force, omm_context):
+            omm_force.setParticleParameters(0, charge=-0.834, sigma=0.3150752406575124, epsilon=0.635968)
+            omm_force.setParticleParameters(1, charge=0.417, sigma=0, epsilon=1)
+            omm_force.setParticleParameters(2, charge=0.417, sigma=0, epsilon=1)
+            omm_force.updateParametersInContext(omm_context)
+
+        # Get Nonbonded force from system -- Note: order may vary by platform. Don't expect a specific index.
+        for force in wbox.system.getForces():
+            if isinstance(force, openmm.NonbondedForce):
```
mikemhenry, line 525: lmao, we have fixed this in a very different way everywhere else, but I ain't mad. Only pitfall is if there was more than one `openmm.NonbondedForce` and we grab the wrong one, but we can fix that if it ever happens (this is testing code anyway)

mikemhenry (approved):

PR #672, opened 2023-03-28
Replacing Fire minimization with LocalEnergyMinimizer

### LINKED ISSUE #668: Use `LocalEnergyMinimizer` instead of `FireMinimizer`
We want to use the `LocalEnergyMinimizer` instead of the `FireMinimizer` in the multistate sampler. Supersedes #557

ijpulidos:
## Description
Replaces `FireMinimizer` with `LocalEnergyMinimizer` to do the replica minimization.

This should enhance the stability of minimization. Supersedes #557 

Resolves #668 

## Todos

## Status
- [ ] Ready to go

## Changelog message
```
L-BFGS based `openmm.LocalEnergyMinimizer` is now used instead of `FireMinimizer` for energy minimization in replicas.
```

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -49,8 +49,6 @@ import mpiplus
 from openmmtools.multistate.utils import SimulationNaNError
 from openmmtools.multistate.pymbar import ParameterError
 
-from openmmtools.integrators import FIREMinimizationIntegrator
-
 logger = logging.getLogger(__name__)
 
 
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1376,9 +1376,6 @@ class MultiStateSampler(object):
         # Compute the final energy of the system for logging.
         final_energy = thermodynamic_state.reduced_potential(sampler_state)
         logger.debug(f'Replica {replica_id + 1}/{self.n_replicas}: final energy {final_energy:8.3f}kT')
-        # TODO if energy > 0, use slower openmm minimizer
-        # Clean up the integrator
-        del context
 
         # Return minimized positions.
         return sampler_state.positions
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1356,11 +1356,8 @@ class MultiStateSampler(object):
         thermodynamic_state = self._thermodynamic_states[thermodynamic_state_id]
         sampler_state = self._sampler_states[replica_id]
 
-        # Use the FIRE minimizer
-        integrator = FIREMinimizationIntegrator(tolerance=tolerance)
-
         # Get context and bound integrator from energy_context_cache
-        context, integrator = self.energy_context_cache.get_context(thermodynamic_state, integrator)
+        context, integrator = self.energy_context_cache.get_context(thermodynamic_state)
         # inform of platform used in current context
         logger.debug(f"{type(integrator).__name__}: Minimize using {context.getPlatform().getName()} platform.")
 
@@ -1373,31 +1370,15 @@ class MultiStateSampler(object):
             replica_id + 1, self.n_replicas, initial_energy))
 
         # Minimize energy.
-        try:
-            if max_iterations == 0:
-                logger.debug('Using FIRE: tolerance {} minimizing to convergence'.format(tolerance))
-                while integrator.getGlobalVariableByName('converged') < 1:
-                    integrator.step(50)
-            else:
-                logger.debug('Using FIRE: tolerance {} max_iterations {}'.format(tolerance, max_iterations))
-                integrator.step(max_iterations)
-        except Exception as e:
-            if str(e) == 'Particle coordinate is nan':
-                logger.debug('NaN encountered in FIRE minimizer; falling back to L-BFGS after resetting positions')
-                sampler_state.apply_to_context(context)
-                openmm.LocalEnergyMinimizer.minimize(context, tolerance, max_iterations)
-            else:
-                raise e
+        openmm.LocalEnergyMinimizer.minimize(context, tolerance, max_iterations)
 
         # Get the minimized positions.
         sampler_state.update_from_context(context)
 
         # Compute the final energy of the system for logging.
         final_energy = thermodynamic_state.reduced_potential(sampler_state)
-        logger.debug('Replica {}/{}: final energy {:8.3f}kT'.format(
-            replica_id + 1, self.n_replicas, final_energy))
+        logger.debug(f'Replica {replica_id + 1}/{self.n_replicas}: final energy {final_energy:8.3f}kT')
         # TODO if energy > 0, use slower openmm minimizer
-
         # Clean up the integrator
         del context
 


ijpulidos:
I'll be also running a pass of the perses tyk2 benchmark just to check this doesn't change the results from it.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1373,31 +1370,15 @@ def _minimize_replica(self, replica_id, tolerance, max_iterations):
...
 
         # Compute the final energy of the system for logging.
         final_energy = thermodynamic_state.reduced_potential(sampler_state)
-        logger.debug('Replica {}/{}: final energy {:8.3f}kT'.format(
-            replica_id + 1, self.n_replicas, final_energy))
+        logger.debug(f'Replica {replica_id + 1}/{self.n_replicas}: final energy {final_energy:8.3f}kT')
         # TODO if energy > 0, use slower openmm minimizer
-
         # Clean up the integrator
         del context
```
jchodera, line 1383: @ijpulidos : I think we might need to remove the `del context` if we're using `self.energy_context_cache.get_context()` since this would delete a `Context` object under the management of the context cache. I can't recall what the correct behavior is here---I think it was to leave the `Context` under management.

This was perhaps different for the FIRE minimizer since a different integrator would lead to a different `Context` we would not re-use.

jchodera (changes requested):
Thanks! This looks good except for the question of whether we should be `del context` at the end of this---see comment.

openmmtools/multistate/multistatesampler.py
ijpulidos: That is a good point, I think we should be okay by removing the `del context` here.

ijpulidos:
I found out that with these changes the minimization seems to be significantly slower and is not using the 100% (nor close) of the GPU. For the `tyk2` benchmark, the differences in times when minimizing complex phase replicas are as follows:

* Using `FIREMinimizationIntegrator`: 32.290s
* Using `LocalEnergyMinimizer`: 500.754s

I could reproduce this both locally with my workstation (that's where I can track the GPU usage) and on lilac (HPC). Is there something special in using our own custom integrators (such as the `FireMinimizationIntegrator` or the `GradientDescentMinimizationIntegrator` used in #557 ) that may explain these differences compared to just using `openmm.LocalEnergyMinimizer`?

@jchodera Maybe you have some suggestions on why this is the case. Thanks!

ijpulidos:
In case it helps, the `LocalEnergyMinimizer` ends up using the `GeodesicBAOABIntegrator` from OpenMM in the end.

ijpulidos:
FYI, I'm having a bit of trouble understanding the `tolerance` units here, we are passing `Quantity(value=1.0, unit=kilojoule/(nanometer*mole))` as tolerance, but as far as I understand, tolerance for energy should just be `kJ/mol`, why does it have `nanometers` in the denominator? How does this translate between the `FireMinimizationIntegrator` and `LocalEnergyMinimizer`? Since we are passing the same value for both.

jchodera:
The tolerance for LocalEnergyMinimizer is a force tolerance, while it is an energy tolerance for FIRE. We probably need to relax the force tolerance considerably so that we don't max out the number of iterations. Can you try timing with something much less stringent, like 10 or 100 kJ/mol/nm?

mikemhenry:
Another thing to consider is, how often do we run this minimizer? It is just at the start of the simulation, right? So adding 9 minutes to a 12-hour-long simulation isn't too bad if this will be more correct and robust--but we should first explore if any of the slowdown is something we can improve (ex John's comment https://github.com/choderalab/openmmtools/pull/672#issuecomment-1489453404)

jchodera:
It's just at the beginning of the calculation, but (1) it makes for a terrible user experience if startup times are really long, (2) it makes testing really slow, and (3) it _shouldn't_ take 9 minutes to get to the point where we can simulate something stably.

ijpulidos:
Timings are as follows (initial and final energies are just for the first replica):

| Minimizer | tolerance | time minimizing all reps. | initial E kT | Final E. kT |
|----------------|--------------|-------|----------------|---------------|
| FIRE | 1 kJ/mol/nm | 32.290s | -306594.761 | -309780.056 |
| LocalEnergyMinimizer | 1 kj/mol | 489.617s | -306554.708 | -326469.138 |
| LocalEnergyMinimizer | 10 kj/mol | 126.885s | -313782.497 | -330901.743 |
| LocalEnergyMinimizer | 100 kj/mol | 3.570s | -313902.645 | -314011.075 |

jchodera:
Hm, this is more difficult since we're not starting from the same initial energy. Any idea what is causing the non-determinisim?

I think we want to compare the quick (<60 second) methods: FIRE and LocalEnergyMinimizer with a loose tolerance (e.g. 50 kJ/mol or so).

I'll look at whether I can fix the FIRE minimizer since it _is_ really fast.

jchodera:
@ijpulidos : Were you able to sort out whether we should or should not delete the `Context` at the end of minimization?

ijpulidos:
> Hm, this is more difficult since we're not starting from the same initial energy. Any idea what is causing the non-determinisim?
> 

Yeah, i also noticed that, the only thing I can tell so far is that the first two (which start from similar energies) were run on the same day and around the same time of the day. Whereas the other two were run at a different day but around the same time. Maybe some correlation with the seed? (This shouldn't happen though, right?)

jchodera:
> Deleting or not deleting does not seem to affect the outcome in terms of energies.

The deleting of `Context`s is a different issue:
* if we do not delete `Context` objects we will never re-use, we consume GPU resources and can end up with new simulations accidentally being run on the CPU
* if we _do_ delete a `Context` under the control of the context cache, we could potentially corrupt the `ContextCache`

We may want to make sure to create a new `Context` and clean it up, rather than use the `ContextCache`.

jfennick:
This PR only 'supercedes' my PR #557 coincidentally, because L-BFGS does not (currently) modify the unit cell vectors, even if pressure is enabled. I still maintain that the proper fix is to simply temporarily disable the pressure during initial minimization, as I have done here https://github.com/choderalab/openmmtools/pull/557/commits/f8dd5c7062cd31a384ca3b927ebe5f53be2ea758 That said, it has been >1 year without any movement on this issue and I have moved on to other projects. If you want to merge any of my PRs, now is the time; I will be deleting my fork soon.

IAlibay:
@ijpulidos I think this switch would still be really useful for us. Is there anything we can help out with to get this to a merged state?

IAlibay:
@ijpulidos could we possible revisit this PR? The current work around here is to run a minimization prior to running the multistatesampler, which is adding as much, if not a bigger, overhead than the performance loss we face going from the FIRE minimizer to LocalEnergyMinimizer.

Ideally just allowing users to set their own tolerance when calling `minimize` would be a ton better.

ijpulidos:
@IAlibay Apologies for such a long discussion and delay on this. Happy to revisit. Do we have a case where the minimization is unstable? It has been reported but I think we don't have systems to actually test this. I have used tyk2 for the performance benchmarks but the actual issue is not performance but stability. Do you happen to have such a system and can you share it in this thread? Thanks!

IAlibay:
@ijpulidos now that L-BFGS is on the GPU (OpenMM 8.5beta), may I suggest we revisit doing this? The performance hit should be negligible.

mikemhenry (approved):
love a good negative pr!

ijpulidos:
While revisiting the test I realized we were **never** actually running any replica minimization in it. With these changes we are actually running replica minimization in the test. This opened a whole thing that can be summarized in the following items:

1. `MonteCarloAnisotropicBarostat(...)` needs a `Vec3` pressure, not a scalar.
2. Our `tracking_minimize` helper indexes `_thermodynamic_states` by `replica_id` directly instead of via `_replica_thermodynamic_states[replica_id]`.
3. ReplicaExchangeSampler requires `n_replicas` <= `n_states`. We are now doing `n_states` == `n_replicas` for that class.
4. Translated-position assertion is now done with a small translation of 0.1nm to avoid artifacts of periodic boundary conditions.
5. barostat injection is a no-op due to the copy-returning system property. Needs the barostat added to the raw System before constructing `ThermodynamicState`.
6. Fixed setting the pressure for anisotropic barostat in `states.py`. Using its `.PressureX/Y/Z()` methods instead of the nonexistent plain `.Pressure()`

mikemhenry:
Way better run times!

openmmtools/states.py
```diff
@@ -1787,7 +1787,9 @@ def _set_barostat_pressure_in_context(barostat, pressure, context):
         """Set barostat pressure."""
         if isinstance(barostat, openmm.MonteCarloAnisotropicBarostat):
             p = pressure.value_in_unit(unit.bar)
-            context.setParameter(barostat.Pressure(), openmm.Vec3(p, p, p)*unit.bar)
+            context.setParameter(barostat.PressureX(), p)
+            context.setParameter(barostat.PressureY(), p)
+            context.setParameter(barostat.PressureZ(), p)
```
mikemhenry, line 1792: Is this from a change in openmm API or just a more clear way to do this?

mikemhenry (approved):
LGTY! Just one question but no blockers!

openmmtools/states.py
ijpulidos: Good question. I don't know if this ever changed in OpenMM at some point. As far as I can tell it has been that way for a while for the anisotropic barostat (which makes sense because the very nature of its anisotropy makes it such that you need to set the pressure independently for each axis). I think the problem was that we never tested it and I suspect it's a feature that people are not actually using (?).

ijpulidos: That said, in our case, for thermodynamic states, it's a little more intricate since we need well-defined thermodynamic states (and associated variables), so the pressure has to be constant, but as far as I understand we can still make the box changes to be different between the axes, that's how we interpret the "anisotropy" here (while maintaining the pressure constant). Sorry for all the details but I guess this is good to have around here at least, I learned about this while dealing with this PR.

PR #679, opened 2023-04-06
INFO logging level for iteration

### LINKED ISSUE #584: MultistateSampler - make iteration logging info level instead of debug?
Discussing things with @mikemhenry today, it would be quite useful if we could get a general sense of run progression for MultistateSampler derived classes without getting all the debug information that usually comes along with it.

I am specifically referring to the following two logging events:

https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistatesampler.py#L683

https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistatesampler.py#L770

Alternatively something like an optional progress bar (tqdm? extra verbose argument for `run`?) would be really useful (and possibly less noisy than dealing with logger outputs).

ijpulidos:
## Description
I forgot to add the actual iteration information at info logging level in #673 . This should take care of that.

Resolves #584

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -765,9 +765,9 @@ class MultiStateSampler(object):
             # Increment iteration counter.
             self._iteration += 1
 
-            logger.debug('*' * 80)
-            logger.debug('Iteration {}/{}'.format(self._iteration, iteration_limit))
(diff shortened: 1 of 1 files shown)


mikemhenry (approved):

mikemhenry (approved):
Looks good!

PR #690, opened 2023-04-20
[WIP] Add explicit deletion to contexts in ContextCache on empty

### LINKED ISSUE #688: Ensure ContextCache deletes all Context objects when it cleans up
We need to make sure that the [`ContextCache`](https://github.com/choderalab/openmmtools/blob/main/openmmtools/cache.py#L214) cleans up all its `Context` objects when it is deleted.

### LINKED ISSUE #689: Should there be an extra context cache clearing in `multistatesampler.__del__`?
Context: From today's OpenFE call, @mikemhenry recommended opening this just to review later.

Quite simply, should there be explit calls to clearing the context cache in `__del__` for multistatesampler?

dotsdl:
## Description
Note: this may not be sufficient; will need to investigate that it
actually succeeds in clearing contexts.

## Todos

## Status
- [ ] Ready to go

## Changelog message
```

```


lhmzhou (approved):

dotsdl:
This may not be finished yet; please do not merge.

jchodera:
This should address https://github.com/choderalab/openmmtools/issues/688

ijpulidos:
We want to write tests that creates many context caches and make sure that the resources of the GPU don't get depleted (if we are emptying the context cache correctly).

We need GPU CI testing for this.

ijpulidos:
After the basic test we can also create a replica exchange sampler test where we create and destroy many repex samplers and check if the GPU resources are correctly handled. 

Minimizers might be creating their own context. We can give it a pre-minimized structure or tell it to do only one minimization step (using the multistatesampler minimize method)

jchodera:
We also need a `__del__` that calls `empty()`.

PR #692, opened 2023-04-24
Appending real time stats

### LINKED ISSUE #691: Real-time analysis YAML files should append, not overwrite
Currently, the real-time analysis YAML files overwrite the old ones if the job is resumed.
Instead, we should open the files to append.

ijpulidos:
## Description
Changing the default behavior of the multistate reporter to append in the YAML real time statistics file instead of overwriting.

We already had the machinery for this, we probably want to discuss if we want to expose this option to the user.

I tested this resuming a run, but we probably want to write a proper test for this behavior. WIP.

Resolves #691 

## Todos

## Status
- [x] Ready to go

## Changelog message
```
Real time statistics in YAML file are now appended when resuming simulations, instead of being overwritten.
```

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -1,6 +1,22 @@
 Release History
 ***************
 
+0.22.0 - pymbar 4 support and gentle equilibration
+==================================================
+
+Enhancements
+------------
+- Openmmtools now supports both Pymbar 3 and 4 versions. (PR `#659 <https://github.com/choderalab/openmmtools/pull/659>`_)
+- Gentle equilibration protocol utility function available in ``openmmtools.utils.gentle_equilibration`` (PR `#669 <https://github.com/choderalab/openmmtools/pull/669>`_).
+- Timing information for multiple state sampler is now reported by default (PRs `#679 <https://github.com/choderalab/openmmtools/pull/679>`_ and `#671 <https://github.com/choderalab/openmmtools/issues/671>`_).
+
+Bugfixes
+--------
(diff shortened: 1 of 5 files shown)


ijpulidos:
A limitation that we have with this approach is that we would need the interval for writing the real time yaml files to be a multiple of the checkpoint interval, or else we would have duplicated iterations information when we resume simulations (think when we stop a simulation in the somewhere in middle of a checkpoint interval and we resume it, the real time yaml file will have duplicated iterations).

I guess we could add a check for this in the constructor of the `MultiStateSampler` object to avoid this issue.

ijpulidos:
Looking at the code probably the better API point where we should be creating this is in `MultiStateSampler.create`, rather than in the constructor. Since this is the place where the storage reporter is created as well, which is the one that handles the checkpointing.

ijpulidos:
This breaks quite a few tests that rely on not making any checkpoints but doing the online analysis. I wonder what's the best approach here.

mikemhenry:
I think we only need the check if we are writing checkpoints, right?

ijpulidos:
> I think we only need the check if we are writing checkpoints, right?

Not sure exactly what you mean here, but we don't have a way to have no checkpointing at all, as far as I can tell. Tests just make the `checkpoint_interval` very large to avoid creating checkpoints.

ijpulidos:
> Not sure exactly what you mean here, but we don't have a way to have no checkpointing at all, as far as I can tell. Tests just make the `checkpoint_interval` very large to avoid creating checkpoints.

We maybe want to have this capability in the future, maybe passing `checkpoint_interval=None` for the reporter to not generate any checkpoints at all.

PR #698, opened 2023-05-25
Importing everything in utils module

ijpulidos:
## Description

Importing all members of the `utils.py` submodule.

In previous changes we created a `openmm.utils` package/module with a `utils.py`sub module for the sake of organizing and modularizing many of the utility functions we have in openmmtools. By doing so I mistakenly only imported some of the members of the modules instead of importing all of them.

## Todos

## Status
- [ ] Ready to go

## Changelog message
```
We are now importing all the members of the new `utils.py` submodule by default. Recovers previous behavior of importing utils with `from openmm.utils import ...`, avoiding API breaking changes.
```


ijpulidos:
This should fix the problems importing the openmmtools utils in downstream packages, such as in https://github.com/choderalab/yank/pull/1286 and https://github.com/choderalab/yank/issues/1299

openmmtools/utils/__init__.py
```diff
@@ -7,27 +7,6 @@
...
-    serialize,
-    SubhookedABCMeta,
-    temporary_directory,
-    time_it,
-    Timer,
-    TrackedQuantity,
-    typename,
-    with_timer,
-)
+from .utils import *
```
mikemhenry, line 10: Do we really want to import everything into the namespace?

ijpulidos: Good question, there's no easy way to know what packages downstream are using, my first attempt was trying NOT to import everything, and only the things that openmmtools itself was needing, but it's hard to tell what other packages are using from `utils.py` so the easiest way is to import everything. I think in the future we probably want to remove the `utils.py` file and just have submodules that make more sense (such as what we did with the gentle equilibration utility function)

ijpulidos: > I haven't skimmed through to make sure that we don't have collisions, if this is exactly how it was working, then that is fine.

Yeah, it should be analogous to what was happening before the changes, such that users can still do `import openmmtools.utils` (or similar)

mikemhenry: Okay then I think it is okay! If we ever do a big breaking change, we can cleanup the API but for now lets make downstream life easy

mikemhenry:
@ijpulidos looks like the tests need fixing (or change the import to be backwards compatible)

ijpulidos:
> @ijpulidos looks like the tests need fixing (or change the import to be backwards compatible)

I think you were right on doubting using the `*` from the start :). I think we just need to be explicit on the imports. I don't know about the details, but apparently _private_ attributes/variables/members (those that start with an underscore for example) are not imported when we do `from module import *` .

mikemhenry (changes requested):
Can you run a sort on these? Or can I?

mikemhenry (approved):
LGTM, I ran a sort on the imports

PR #699, opened 2023-05-25
report mode GPU is in and write out gpu UUID

### LINKED ISSUE #693: Don't output nvidia-smi failure in automated platform search
We've had folks a bit confused about this message showing up when they don't have a GPU device:

```
/bin/sh: nvidia-smi: command not found
```

Is there a way to avoid outputting this to users?

### LINKED ISSUE #697: [ENH] Utility for catching GPUs in exclusive mode
After discussions with @mikemhenry we think this might need to live here.

We have had issues with folks trying to use a multistatesampler derived class (repex, sams, etc...) and things failing in a non-clean manner beause their GPU setup was set in exclusive mode (tagging @ijpulidos who found this issue at the OMSF workshop).

It would be good to have a utility here that does a call to `nvidia-smi --query-gpu="compute_mode" --format=csv` or similar to get the compute mode and check that we aren't in exclusive mode.

mikemhenry:
## Description
Improve GPU logging information, inform user of potential issue with GPU compute mode, provide a better error message when `openmm.OpenMMException: No compatible CUDA device is available` is thrown.

Resolves #693 
Resolves #697 

## Todos

## Status
- [ ] Ready to go

## Changelog message

```
Fixes #693 #697 Improve GPU logging information, inform user of potential issue with GPU compute mode, provide a better error message when `openmm.OpenMMException: No compatible CUDA device is available` is thrown. Note, formatting of "CUDA devices available:" message has changed. 
```

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -35,6 +35,7 @@ import typing
 import inspect
 import logging
 import datetime
+import subprocess
 
 import numpy as np
 
@@ -1774,10 +1775,16 @@ class MultiStateSampler(object):
     def _display_cuda_devices():
         """Query system nvidia-smi to get available GPUs indices and names in debug log."""
         # Read nvidia-smi query, should return empty strip if no GPU is found.
-        cuda_query_output = os.popen("nvidia-smi --query-gpu=index,gpu_name --format=csv,noheader").read().strip()
-        # Split by line jump and comma
-        cuda_devices_list = [entry.split(',') for entry in cuda_query_output.split('\n')]
-        logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+        cuda_query_output = subprocess.run("nvidia-smi --query-gpu=gpu_uuid,gpu_name,compute_mode  --format=csv", shell=True, capture_output=True, text=True)
+        print(cuda_query_output)
+        if cuda_query_output.stdout:
+            # Split by line jump and comma
+            cuda_devices_list = [entry for entry in cuda_query_output.stdout.splitlines()]
+            logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+            if "Default" in cuda_query_output.stdout:
+                logger.warn(f"GPU in 'Exclusive_Process' mode, one context is allowed per device. This may prevent some openmmtools features from working.")
+        else:
+            logger.debug(f"nvidia-smi command failed: {cuda_query_output.stderr}, this is expected if there is no GPU available")
 
     def _flatten_moves_iterator(self):
         """Recursively flatten MCMC moves. Handles the cases where each move can be a set of moves, for example with


openmmtools/multistate/multistatesampler.py
```diff
@@ -1774,10 +1775,16 @@ def _update_timing(self, iteration_time, partial_total_time, run_initial_iterati
...
+        cuda_query_output = subprocess.run("nvidia-smi --query-gpu=gpu_uuid,gpu_name,compute_mode  --format=csv", shell=True, capture_output=True, text=True)
+        print(cuda_query_output)
+        if cuda_query_output.stdout:
+            # Split by line jump and comma
+            cuda_devices_list = [entry for entry in cuda_query_output.stdout.splitlines()]
+            logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+            if "Default" in cuda_query_output.stdout:
+                logger.warn(f"GPU in 'Exclusive_Process' mode, one context is allowed per device. This may prevent some openmmtools features from working.")
+        else:
+            logger.debug(f"nvidia-smi command failed: {cuda_query_output.stderr}, this is expected if there is no GPU available")
```
ijpulidos, line 1787: This is looking good! I think here we probably want to check the return code, we could do that just by checking the `cuda_query_output.returncode` value (if it's 0 it should be okay), or we could also probably add the `check=True` argument to the subprocess call, which would return a `CalledProcessError if it fails, we can try catching that error to report it in the log.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1774,10 +1775,16 @@ def _update_timing(self, iteration_time, partial_total_time, run_initial_iterati
...
-        # Split by line jump and comma
-        cuda_devices_list = [entry.split(',') for entry in cuda_query_output.split('\n')]
-        logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+        cuda_query_output = subprocess.run("nvidia-smi --query-gpu=gpu_uuid,gpu_name,compute_mode  --format=csv", shell=True, capture_output=True, text=True)
+        print(cuda_query_output)
+        if cuda_query_output.stdout:
+            # Split by line jump and comma
+            cuda_devices_list = [entry for entry in cuda_query_output.stdout.splitlines()]
+            logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+            if "Default" in cuda_query_output.stdout:
```
ijpulidos, line 1784: I guess you mean to be checking for `E. Process` or not `Default` here, instead.

ijpulidos (changes requested):
Looking good, just a few suggestions.

openmmtools/multistate/multistatesampler.py
mikemhenry: Oooops, that was because I couldn't get my card to simulate `E. process` since I only have one card in this laptop, but I wanted to make sure the logic, will update that when I test this on a box I put the gpu in `E. process` mode.

openmmtools/multistate/multistatesampler.py
mikemhenry: You know I think you are right, I was assuming that if `nvidia-smi` was not there, nothing would be written to stdout, so the logic check would work fine, but if there is a case where something is written to stdout and stderr, then this wouldn't work. I thought this was cleaner than catching a `CalledProcessError`, but I think there is an edge case where it doesn't work. I will refactor it to catch the error (which is easy, just need to replace the if/esle with try/except, and run with check=True inside the try block. Oh and I need to remove the `print` there that was for debugging.

openmmtools/multistate/multistatesampler.py
```diff
@@ -1773,11 +1774,15 @@ def _update_timing(self, iteration_time, partial_total_time, run_initial_iterati
...
-        cuda_query_output = os.popen("nvidia-smi --query-gpu=index,gpu_name --format=csv,noheader").read().strip()
-        # Split by line jump and comma
-        cuda_devices_list = [entry.split(',') for entry in cuda_query_output.split('\n')]
-        logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+        try:
+            cuda_query_output = subprocess.run("nvidia-smi --query-gpu=gpu_uuid,gpu_name,compute_mode  --format=csv", shell=True, capture_output=True, text=True)
+            # Split by line jump and comma
+            cuda_devices_list = [entry for entry in cuda_query_output.stdout.splitlines()]
+            logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+            if "Exclusive_Process" in cuda_query_output.stdout:
```
ijpulidos, line 1782: I think the string here isn't `Exclusive_Process` but `E. Process`, that's why I suggested maybe just catching `not "Default" in cuda_query_output.stdout` ? But whatever you think works best. As of now, we only support the `Default` compute mode (the other mode left is simply disabled).

ijpulidos (changes requested):
Almost there! :)

openmmtools/multistate/multistatesampler.py
mikemhenry:
```
$ nvidia-smi --help-query-gpu
...
"compute_mode"
The compute mode flag indicates whether individual or multiple compute applications may run on the GPU.
"Default" means multiple contexts are allowed per device.
"Exclusive_Process" means only one context is allowed per device, usable from multiple threads at a time.
"Prohibited" means no contexts are allowed per device (no compute apps).
...
```

I was going off what the help string seemed to indicated, I think that `not "Default" in cuda_query_output.stdout` is less clear what we are doing, I will explain in the doc string

openmmtools/multistate/multistatesampler.py
```diff
@@ -1773,11 +1774,15 @@ def _update_timing(self, iteration_time, partial_total_time, run_initial_iterati
     @staticmethod
     def _display_cuda_devices():
         """Query system nvidia-smi to get available GPUs indices and names in debug log."""
-        # Read nvidia-smi query, should return empty strip if no GPU is found.
-        cuda_query_output = os.popen("nvidia-smi --query-gpu=index,gpu_name --format=csv,noheader").read().strip()
-        # Split by line jump and comma
-        cuda_devices_list = [entry.split(',') for entry in cuda_query_output.split('\n')]
-        logger.debug(f"CUDA devices available: {*cuda_devices_list,}")
+        try:
+            cuda_query_output = subprocess.run("nvidia-smi --query-gpu=gpu_uuid,gpu_name,compute_mode  --format=csv", shell=True, capture_output=True, text=True)
```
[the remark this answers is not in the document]
mikemhenry: Yes of course, I really don't know how I missed that when I switched to try/except

mikemhenry:
Do you calling the command twice is okay? https://github.com/choderalab/openmmtools/pull/699/commits/cf286f557e2d7b1c62a95871ace841bdcb322cb0#diff-2329790a58bd1510dc3f9b49f4ba83c331d71627f62e70d01f57677bf6f6a1b2R1786-R1787

Or should I check the error code instead of the try/except? If I check the error code then I don't have to worry about calling it twice. Thoughts?

ijpulidos:
> Or should I check the error code instead of the try/except? If I check the error code then I don't have to worry about calling it twice. Thoughts?

Couldn't you just get the error message from the previous `cuda_query_output`? As with `cuda_query_output.stdout.read()`?

mikemhenry:
No because it throws a `subprocess.CalledProcessError`, it doesn't return a completed process object if it throws an error.

ijpulidos (approved):
Tested this on my local machine with two different GPUs, it works! LGTM!

mikemhenry:
I'm going to change it a bit, I'm worried that we could end up with an error on the first call, but no error on the second call to nvidia-smi or sommething

mikemhenry:
@ijpulidos sorry for changing my mind and the last second here, but I think this is better now

ijpulidos (approved):
Looks good to me! LGTM.

mikemhenry:
Few flaky tests it seems, re-running them

Issue #700, opened 2023-05-31
Errors when resuming simulations from storage with recent numba 0.57


ijpulidos:
Many changes were made in the latest numba release 0.57. Some of them are API breaking changes, specifically we cannot use `MakedArray`s with numba, they appear to never been supported but now trying to use them is raising an exception with the changes upstream in https://github.com/numba/numba/pull/8597

We are getting these exceptions in our CI in https://github.com/choderalab/perses/actions/runs/5116171967/jobs/9198059420#step:9:858

Note that `openmmtools` CI is not affected by this, so my guess is that this is something that perses is doing when passing the arguments to `openmmtools.multistate.MultiStateSampler` especially in the replica mixing as pointed in the previous error.

ijpulidos:
Closed via #701

PR #701, opened 2023-05-31
manually casting objects to np arrays when reading checkpoint

ijpulidos:
## Description
Numba 0.57 explicitly raises an exception when trying to use `MaskedArrays` in `@jit` decorated functions. When we are reading `MultiStateSampler` attributes from serialized objects (storage) these attributes become `MaskedArray`. We are casting them to arrays here to make sure we can still use numba when mixing replicas after resuming the simulation.

Solves #700 

## Todos

## Status
- [ ] Ready to go

## Changelog message
```
Support for latest numba 0.57 by deserializing attributes as numpy arrays instead of MaskedArrays.
```

### DIFF
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1025,12 +1025,12 @@ class MultiStateSampler(object):
         self._thermodynamic_states = thermodynamic_states
         self._unsampled_states = unsampled_states
         self._sampler_states = sampler_states
-        self._replica_thermodynamic_states = state_indices
+        self._replica_thermodynamic_states = np.array(state_indices)
         self._energy_thermodynamic_states = energy_thermodynamic_states
         self._neighborhoods = neighborhoods
         self._energy_unsampled_states = energy_unsampled_states
-        self._n_accepted_matrix = n_accepted_matrix
-        self._n_proposed_matrix = n_proposed_matrix
+        self._n_accepted_matrix = np.array(n_accepted_matrix)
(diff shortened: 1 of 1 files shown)


mikemhenry:
Any ideas on the consequence of this?

mikemhenry:
We might need to check the mask to make sure this is safe
https://numpy.org/doc/stable/reference/maskedarray.generic.html#what-is-a-masked-array

mikemhenry:
Ah so we are not using the actual masking functionality here? I'm just worried that if there is a mask on these arrays, that means we don't want to include some of that data in the analysis.

mikemhenry (changes requested):
Just want to make sure that we are not using the mask

ijpulidos:
> Just want to make sure that we are not using the mask
I think we don't use it or needing, since they weren't masked array from the start and resuming simulations seems to be working. That said, I cannot be completely sure.

I don't know of a systematic way to be sure we are not using the mask, it just seems that every time we read from a .nc file we will always get a `MaskedArray` , as in https://github.com/Unidata/netcdf4-python/pull/787

ijpulidos:
We could probably come up with a less hacky or a more general solution that uses `set_auto_mask(False)` whenever we read from the .nc files

mikemhenry:
If we don't want the masking behavior or use it, then when we save the nc file we should use `set_auto_mask(False)` but it sounds like this is okay to merge then

mikemhenry (approved):

PR #705, opened 2023-06-06
Fix issue with test that would run forever

mikemhenry:
## Description

## Todos

## Status
- [ ] Ready to go

## Changelog message
```

```

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
         python-version: ["3.8", "3.9", "3.10"]
         openmm: ["7.7", "8.0"]
         os: [macOS-latest, ubuntu-latest]
-        pymbar-version: ["3"]
+        pymbar-version: ["4"]
         include:
           # Test openmm dev build on newest python + linux
           - openmm: "dev"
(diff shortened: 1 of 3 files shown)


ijpulidos:
I think I understand now where the problem is happening. The problem is in https://github.com/choderalab/openmmtools/blob/72709cef2d776a373bc22235d02ba97f8dd8544f/openmmtools/tests/test_sampling.py#L1475 (MacOS fails the assertion). That `._last_err_free_energy` gets asigned/modified by two functions [`MultiStateSampler._online_analisis`](https://github.com/choderalab/openmmtools/blob/ccd6b941adb5fe56072e0e9b9848c4f3b199e645/openmmtools/multistate/multistatesampler.py#L1600) and [`MultiStateSampler._offline_analysis`](https://github.com/choderalab/openmmtools/blob/ccd6b941adb5fe56072e0e9b9848c4f3b199e645/openmmtools/multistate/multistatesampler.py#L1501).

* `_online_analysis` just sets it to `np.Inf`
* `_offline_analysis` uses MBAR to estimate the error, and in the case of the test it results in an error of `0.0`, failing the assertion (only for MacOS, Linux results in errors different than 0).
* For historical reasons, `_offline_analysis` is called every `online_analysis_interval` (this is what generates the real time yaml file), whereas `_online_analysis` is called at every iteration. This is confusing, but at the time we did not want to break/change the API, since this has always been the behavior.

ijpulidos (approved):
Seems good! LGTM

PR #712, opened 2023-07-13
MultiStateReporter variable pos/vel save frequency

richardjgowers:
## Description
allows MultiStateReporter to write positions and velocities and different frequency to energies data.

## Todos

## Status
- [ ] Ready to go

## Changelog message
```
MultiStateReporter now takes optional `position_interval` and `velocity_interval` keyword args to control the frequency these are saved at
```


mikemhenry:
Will we run into key errors if you try and use this with an `.nc` file from an older version?

mikemhenry (approved):

ijpulidos (commented):
Thanks for this contribution! This looks good to me.

I do have questions about the motivation for this, since it adds a complexity that I don't seem to get why it is needed, can you expand on this?

Also, we probably want to write tests for this, since it's changing the behavior of the reporter and that can be critical for our users.

richardjgowers:
@ijpulidos the filesizes for MultiStateReporter can get pretty large, since you're saving ~11 trajectories.  You typically want to save the energies at a high frequency (since you're very interesting in seeing correlation times) but positions you can have these at a lower frequency, velocities you might not even care about.

Can you point me towards the existing tests so I can make sure the tests I write fit in?

ijpulidos:
I see, yeah I had the feeling it was a storage issue. We probably need to clean things up in general but I think this will help. Thanks for clarifying the motivation.

As per the tests, I think having them as part of the [`TestReporter`](https://github.com/choderalab/openmmtools/blob/6cef59b67327d1a16685304368e87e99a37a441c/openmmtools/tests/test_sampling.py#L268) class should be the place, we might consider moving this class to its own test module, but we don't need that right now. Thanks!

mikemhenry:
Yah I think we need to also make sure we figure out what our analysis tools expect. My hunch is that we need to save a lot of stuff to resume a simulation, but that should be a checkpoint, the time series data we save for analysis should be user configurable.

ijpulidos:
> the time series data we save for analysis should be user configurable.

Yes, decoupling the _checkpointing_ with the user-configurable time series data would probably work better here. I'd vote for that route if it makes sense.

As far as I can tell these are coupled. We would need to think a little bit further about this, since I'm not sure how big of a refactor this could be. There's a lot of moving back and forth between the `MultiStateReporter` and the `MultiStateSampler`.

openmmtools/tests/test_sampling.py
```diff
@@ -414,6 +419,124 @@ def test_write_sampler_states(self):
...
+                assert np.allclose(state.positions, restored_state.positions)
+                # By default stored velocities are zeros if not present in origin sampler_state
+                assert np.allclose(np.zeros(state.positions.shape), restored_state.velocities)
+                assert np.allclose(state.box_vectors / unit.nanometer, restored_state.box_vectors / unit.nanometer)
+            # Second frame should not have positions or velocities
+            restored_sampler_states = reporter.read_sampler_states(iteration=1, analysis_particles_only=True)
+            for state, restored_state in zip(sampler_states, restored_sampler_states):
+                # missing values are returned as numpy masked array
+                # so we check that these arrays are all masked
+                assert restored_state.positions._value.mask.all()
```
richardjgowers, line 451: this is something that I'm not 100% sure of currently.  netCDF will return a [numpy masked array ](https://numpy.org/doc/stable/reference/maskedarray.html) when you access data that isn't present (here we're saving vels every other frame, so accessing velocities here has no data).  I hadn't ever encountered these, so maybe it's not the best thing to return? (or maybe it is if it's the netCDF normal return).

openmmtools/tests/test_sampling.py
```diff
@@ -414,6 +419,124 @@ def test_write_sampler_states(self):
...
+                assert restored_state.velocities._value.mask.all()
+                assert restored_state.box_vectors is None  # not periodic
```
richardjgowers, line 453: this is something I need to double check, I'm a little confused why the checkpoint has a box and future frames don't, so this is still WIP till I figure that out

richardjgowers:
@ijpulidos @mikemhenry this is much closer to ready, I've added some tests so might be a good time for you both to read through

mikemhenry:
@richardjgowers getting some errors in CI 

```
  File "/home/runner/micromamba-root/envs/openmmtools-test/lib/python3.9/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/runner/work/openmmtools/openmmtools/openmmtools/tests/test_sampling.py", line 523, in test_write_sampler_states_no_pos
    assert restored_state.positions._value.mask.all()
AttributeError: 'numpy.ndarray' object has no attribute 'mask'
```

See the CI log here: https://github.com/choderalab/openmmtools/actions/runs/8052365133

Honestly if you make a commit to bump this PR, you should get the same results, I was having an issue re-starting CI

ijpulidos:
I wonder if this is what's biting us here, though I don't know why are we expecting these arrays to be masked? https://github.com/choderalab/openmmtools/pull/701

ijpulidos:
Now, to be fair, I don't think the changes in that linked PR have anything to do with the serialized/deserialized positions or velocities. So I doubt it.

dotsdl:
@ianmkenney will work with @mikemhenry to complete this!

ijpulidos:
#767 is taking over this one. Closing this.

PR #721, opened 2023-11-06
Create alchemy subpackage

ijpulidos:
## Description
The idea of this PR is to create the alchemy subpackage in a way that doesn't break the current API.

This is to give room to extend the `alchemy` capabilities of openmmtools with the tools that would be coming from `perses` and `openfe` (for example check roadmap in https://github.com/orgs/choderalab/projects/5)

In the future, I expect that these objects will be moved around to different modules or that the current `alchemy.py` module changes its name for something more descriptive of what it has. So far, this is just thought as a starting point.

## Todos

This shouldn't be changing any of the behavior, this should be reflected in the tests in `test_sampling.py` and `test_alchemy.py` passing, which are the ones that call the objects in the `alchemy` module.

## Status
- [x] Ready to go


mikemhenry:
Thanks for kicking this off!

ijpulidos:
There are some imports in the tests to be fixed, they were relying on the `alchemy.py` module to import things like `openmm` and `copy`, and I think that shouldn't happen. Will fix this.

openmmtools/tests/test_alchemy.py
```diff
@@ -25,11 +27,15 @@
 
 import nose
 import scipy
+import numpy as np
 from nose.plugins.attrib import attr
 
-from openmmtools import testsystems, forces
-from openmmtools.constants import kB
-from openmmtools.alchemy import *
```
mikemhenry, line 32: ooooof yes lets try and not use `*` import except for in __init__.py to pull everything from a module into a different namesapce.

mikemhenry (approved):

PR #741, opened 2024-08-15
Fix CI

mikemhenry:
## Description
Provide a brief description of the PR's purpose here.

## Todos
- [ ] Implement feature / fix bug
- [ ] Add [tests](https://github.com/choderalab/openmmtools/tree/master/openmmtools/tests)
- [ ] Update [documentation](https://github.com/choderalab/openmmtools/tree/master/docs) as needed
- [ ] Update [changelog](https://github.com/choderalab/openmmtools/blob/master/docs/releasehistory.rst) to summarize changes in behavior, enhancements, and bugfixes implemented in this PR

## Status
- [ ] Ready to go

## Changelog message
```

```

PR #744, opened 2024-08-26
Use OMSF action for AWS self-hosted runner

mikemhenry:
This migrates us to using the new OMSF action for setting up a self-hosted runner. Pretty much the same setup as last time but maintained by OMSF infra team and isn't tied to one cloud provider.


mikemhenry:
This will be tested in #741

ijpulidos (approved):
Looks good to me!

PR #749, opened 2024-09-25
"Fix" flaky tests

mikemhenry:
We had a test that seg faults on osx sometimes and a test that always fails on windows, this PR keeps CI green


ijpulidos (approved):
Yes! We want this :) LGTM!

PR #767, opened 2025-01-16
Multistatereporter variable pos frequency

ianmkenney:
Continuation of #712. ~There are multiple avenues to address the current failing tests. I would like to address those concerns here before continuing in that PR.~


ianmkenney:
This PR will take over #712.

ianmkenney:
@ijpulidos @IAlibay @mikemhenry I can't request reviewers, but I think this is ready for review.

ijpulidos (approved):
This looks good to me. Thanks a lot! Great work. Just a couple of non-blocking comments.

openmmtools/multistate/multistatereporter.py
```diff
@@ -1727,21 +1763,35 @@ def _read_sampler_states_from_given_file(self, iteration, storage_file='checkpoi
...
-                x = storage.variables['positions'][read_iteration, replica_index, :, :].astype(np.float64)
-                positions = unit.Quantity(x, unit.nanometers)
+                try:
+                    x = storage.variables['positions'][read_iteration, replica_index, :, :].astype(np.float64)
+                    # if any data is masked, treat as if iteration index didn't exist
+                    if x.mask.any():
+                        raise IndexError
+                    # pull the valid data out of masked array
+                    positions = unit.Quantity(x.data, unit.nanometers)
+                except (IndexError, KeyError):
```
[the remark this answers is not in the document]
ianmkenney: On the topic of how `MaskedArray`s work here. Assuming that a particular coordinate exists in a dimension (e.g. at least some data exists for a given iteration) we always get back a `MaskedArray`. The `MaskedArray` will have a `data` attribute with a numpy array that will be masked by the `MaskedArray`s `mask` attribute. If the data was all valid, then we just pull that information out and expose it to the user. If no data exists at that dimension coordinate, then we'll just get an `IndexError` so we give back a zero array in place of the nonsense data inside the `MaskedArray`.  By doing it this way, a user never has to deal with a `MaskedArray`.

I think that if the data was masked it's basically the same behavior as if the data didn't exist and we would have expected a zero array anyways. I don't really see much of a reason to give a warning about this since having a giant array of zeros is kind of a warning already. I'd even be careful saying the data is missing since that would imply it should have been there in the first place.

openmmtools/multistate/multistatereporter.py
```diff
@@ -100,6 +100,12 @@ class MultiStateReporter(object):
     analysis_particle_indices : tuple of ints, Optional. Default: () (empty tuple)
         If specified, it will serialize positions and velocities for the specified particles, at every iteration, in the
         reporter storage (.nc) file. If empty, no positions or velocities will be stored in this file for any atoms.
+    position_interval : int, default 1
+        the frequency at which to write positions relative to analysis
+        information, 0 would prevent information being written
+    velocity_interval : int, default 1
+        the frequency at which to write positions relative to analysis
+        information, 0 would prevent information being written
```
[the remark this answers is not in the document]
ianmkenney: I think this would depend on the intent of the simulation. I could see that someone interested in doing VAC calculations wouldn't care as much about the positions.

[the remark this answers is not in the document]
IAlibay: So the one case I can think of is that you might not want to store the velocities but you do want to store the positions.

I might be wrong, but as far as I remember the "state" is stored in the checkpoint? So you could recover a simulation from the checkpoint, and just store positions in the simulation file.

IAlibay:
I'll be testing this in a Protocol before I review :)

openmmtools/multistate/multistatereporter.py
```diff
@@ -100,6 +100,12 @@ class MultiStateReporter(object):
     analysis_particle_indices : tuple of ints, Optional. Default: () (empty tuple)
         If specified, it will serialize positions and velocities for the specified particles, at every iteration, in the
         reporter storage (.nc) file. If empty, no positions or velocities will be stored in this file for any atoms.
+    position_interval : int, default 1
```
IAlibay, line 103: Could we store these values inside of the NetCDF file as additional variables? This would make it a lot easier to know what the spacing is between frames, rather than having to iterate through the positions.

openmmtools/multistate/multistatereporter.py
```diff
@@ -100,6 +100,12 @@ class MultiStateReporter(object):
...
+        the frequency at which to write positions relative to analysis
```
IAlibay, line 104: Could you also make it clear here that this controls whether or not box_vectors are written down? (it's self-evident, but it might be good to make it clear to users)

openmmtools/multistate/multistatereporter.py
```diff
@@ -202,6 +214,16 @@ def checkpoint_interval(self):
         """Returns the checkpoint interval"""
         return self._checkpoint_interval
 
+    @property
+    def position_interval(self):
+        """Interval relative to energies that positions are written at"""
```
IAlibay, line 219: And box_vectors

openmmtools/multistate/multistatereporter.py
```diff
@@ -1727,21 +1763,35 @@ def _read_sampler_states_from_given_file(self, iteration, storage_file='checkpoi
...
+                    # pull the valid data out of masked array
+                    velocities = unit.Quantity(x.data, unit.nanometer / unit.picoseconds)
+                except (IndexError, KeyError):  # Velocities key/variable not found in serialization (openmmtools<=0.21.2)
                     # pass zeros as velocities when key is not found (<0.21.3 behavior)
                     velocities = np.zeros_like(positions)
 
                 if 'box_vectors' in storage.variables:
                     # Restore box vectors.
                     x = storage.variables['box_vectors'][read_iteration, replica_index, :, :].astype(np.float64)
+                    # TODO: Are box vectors also variably saved?
```
IAlibay, line 1794: From wwhat I can tell, they seem to be saved at the same rate as positions - worth double checking though!

ianmkenney: It is saved at the same rate of positions assuming that the sampler_state's box_vectors is not None (line 1669). I find this line a little odd though since there is a`is_periodic` property of the reporter (line 204) that follows slightly different logic. I'm not sure which is best to use. @ijpulidos do you have any thoughts on this?

openmmtools/multistate/multistatereporter.py
```diff
@@ -1638,7 +1666,7 @@ def _write_sampler_states_to_given_file(self, sampler_states: list, iteration: i
 
         storage = self._storage_dict[storage_file]
         # Check if the schema must be initialized, do this regardless of the checkpoint_interval for consistency
-        is_periodic = True if (sampler_states[0].box_vectors is not None) else False
+        is_periodic = sampler_states[0].box_vectors is not None
```
ianmkenney, line 1669: @ijpulidos is there a reason we aren't using the `MultiStateReporter.is_periodic` property here?

ijpulidos: I cannot say for sure. I guess this means that one can have a `box_vectors` variable in the reporter but not necessarily have `box_vectors` attribute in the sampler states, and vice versa. I don't know if we ever want those NOT to match. Is there a use case for such scenario?

openmmtools/multistate/multistatereporter.py
ijpulidos: To me it makes sense that these are stored at the same rate of positions, but again, there might be a use case where we want different rates for it? The logic here being that to maintain the thermodynamic states (especially pressure) you need to vary the box size and that's dependent on the positions. I'd argue that having positions without box sizes, or vice versa, is probably something we don't want to deal with.

openmmtools/multistate/multistatereporter.py
ianmkenney: I suppose I'd be inclined to keep it the same in this PR, it just struck me as strange. Both lines were introduced in the same commit so I'm going to go on the assumption it was deliberate rather than an oversight.

ianmkenney:
@IAlibay @ijpulidos unless there is more to do here, I think this can be merged!

IAlibay (commented):
I can't approve on this repo, but lgtm!

PR #773, opened 2025-03-27
Disable the barostat during FIRE minimization

hannahbaumann:
## Description
As has been described here (https://github.com/choderalab/openmmtools/pull/557), box vectors can change during energy minimization using the FIRE minimizer, leading to instabilities. This PR temporarily disables the barostat for the minimization.

## Changelog message
```

```

### DIFF
--- a/openmmtools/integrators.py
+++ b/openmmtools/integrators.py
@@ -2371,7 +2371,7 @@ class FIREMinimizationIntegrator(mm.CustomIntegrator):
         self.addGlobalVariable("delta_t", timestep.value_in_unit_system(unit.md_unit_system))
 
         # Update context state.
-        self.addUpdateContextState()
+        #self.addUpdateContextState()
 
         # Assess convergence
         # TODO: Can we more closely match the OpenMM criterion here?
--- a/openmmtools/multistate/multistatesampler.py
+++ b/openmmtools/multistate/multistatesampler.py
@@ -1360,6 +1360,13 @@ class MultiStateSampler(object):
         # Retrieve thermodynamic and sampler states.
         thermodynamic_state_id = self._replica_thermodynamic_states[replica_id]
         thermodynamic_state = self._thermodynamic_states[thermodynamic_state_id]
+
+        # Temporarily disable the barostat during minimization.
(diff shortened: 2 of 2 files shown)


hannahbaumann:
@ijpulidos @mikemhenry : This seems to be solving the issues with random crashes when starting NPT equilibration after FIRE minimization that I mentioned on Tuesday. On some short test calculations I'm not seeing any crashes.

openmmtools/integrators.py
```diff
@@ -2371,7 +2371,7 @@ def __init__(self, timestep=1.0 * unit.femtoseconds, tolerance=None, alpha=0.1,
         self.addGlobalVariable("delta_t", timestep.value_in_unit_system(unit.md_unit_system))
 
         # Update context state.
-        self.addUpdateContextState()
+        #self.addUpdateContextState()
```
hannahbaumann, line 2374: I removed this as I had seen this (https://github.com/openmm/openmm/issues/1237#issuecomment-1938047795), but in test calculations I don't see crashes if we update the context state so this may not be necessary?

hannahbaumann:
Actually, as @jthorton pointed out, removing the `self.addUpdateContextState()` and setting the pressure to None should both have the same effect, so we could do either!

mikemhenry:
I like setting the pressure to `None` as it is a bit more clear IMHO

mikemhenry:
looks like we have a test that checks that actually `FAILED openmmtools/tests/test_integrators.py::test_update_context_state_calls - AssertionError: Integrator 'FIREMinimizationIntegrator' has 0 calls to addUpdateContextState(), while there should be only one.`

openmmtools/integrators.py
mikemhenry, line 2374:
```suggestion
        self.addUpdateContextState()
```

mikemhenry:
I can't edit this PR, @hannahbaumann can you check the box described here? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #775, opened 2025-03-31
chore: migrate to new OMSF start/stop runners

ethanholz:
## Description
Implement the new OMSF `gha-runner` actions. 

## Todos

## Status
- [x] Ready to go

## Changelog message
```
Migrate to OMSF's `start-aws-gha-runner` and `stop-aws-gha-runner`. 
```

Is this the kind of change you would like to see reflected in the changelog?

### DIFF
--- a/.github/workflows/gpu-runner.yaml
+++ b/.github/workflows/gpu-runner.yaml
@@ -18,13 +18,10 @@ jobs:
           aws-region: us-east-1
       - name: Create cloud runner
         id: aws-start
-        uses: omsf-eco-infra/gha-runner@v0.4.0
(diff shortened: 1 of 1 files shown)


mikemhenry:
> Is this the kind of change you would like to see reflected in the changelog?

Sure, always better to have something. It will likely get cut when we make the final changelog since we try and write those more for end users (who don't really care about CI stuff) BUT it is always easier to edit than to try and come up with something when we are cutting the release.

mikemhenry (approved):

ethanholz:
Sounds good to me! Is there anything else I can do here or is this good to go?

PR #809, opened 2026-06-02
update to forcefactories.py

blakemertz:
## Description
Had to introduce a small change in syntax when upgrading from openmm7 to openmm8.

## Todos

## Status
- [ ] Ready to go

## Changelog message
```

```


openmmtools/forcefactories.py
```diff
@@ -101,6 +101,8 @@ def restrain_atoms_by_dsl(thermodynamic_state, sampler_state, topology, atoms_ds
     restrained_atoms = mdtraj_topology.select(atoms_dsl).tolist()
     restrain_atoms(thermodynamic_state, sampler_state, restrained_atoms, **kwargs)
 
+    reutrn system
```
mikemhenry, line 104: I think you mean
```suggestion
    return system
```

mikemhenry:
What error were you getting that this solved?

openmmtools/forcefactories.py
blakemertz: Yes, I forgot to correct that after I ran into that misspelling on my own workstation. Sorry about that.

blakemertz:
@mikemhenry When restraints are created the code did not return the new system with restraints, so I put in python code to return the system as part of the restrain function.

My original pull request is pretty stale, and I think that the issue has been resolved in the meantime. I am closing this for now. When I get a chance I will go back and see if I can reproduce the issue -- if the issue persists, I'll open a new PR with better detail on it.

PR #812, opened 2026-08-06
Document GHMCMove performance considerations

yihengwuKP:
## Description

This adds a short documentation note that `GHMCMove` uses a Metropolized propagator and can be substantially more expensive per integration step than non-Metropolized dynamics moves such as `LangevinDynamicsMove`.

The note also points users to the existing GHMC acceptance statistics (`n_proposed`, `n_accepted`, and `fraction_accepted`) and recommends benchmarking throughput for performance-sensitive workflows.

## Motivation

`GHMCMove` has a similar user-facing constructor shape to `LangevinDynamicsMove`, but it is algorithmically different and may carry meaningful overhead, especially for small systems, short propagation segments, or many-state/multistate workflows.

## Testing

- `python -m py_compile openmmtools/mcmc.py`


mikemhenry:
Thanks for this!

mikemhenry:
ci errors:
```
FAILED openmmtools/tests/test_sampling.py::TestReplicaExchange::test_from_storage - TypeError: only 0-dimensional arrays can be converted to Python scalars
FAILED openmmtools/tests/test_sampling.py::TestParallelTempering::test_from_storage - TypeError: only 0-dimensional arrays can be converted to Python scalars
```
unrelated to the PR but just an FYI to future people

ijpulidos (changes requested):
Thanks! This looks like a great clarification and addition to the docs. I would only suggest if we could be more specific about where the bigger cost come from. If I recall correctly, this mostly comes from having to do more energy evaluations by the Metropolization, rather than the somewhat small cost of computing the acceptance/rejection. Could we add this detail to the comments? Otherwise, looking great!

yihengwuKP:
Sounds good. I just pushed a quick clarification

ijpulidos:
I'm marking this ready for a final review/approval. Thanks!

ijpulidos (approved):
LGTM!

